PR #9, opened 2022-02-10
Adds a capture operator to capture output

davidselassie:
Adds a `capture` operator that runs a `captor` function on every piece
of data flowing through a point in the dataflow. It does an internal
exchange so it will see _all_ data, but that also means it has more
overhead. It passes a single `(epoch, item)` to the function
provided. You can use this with things like `set.add` or `list.append`
to store what was seen for later assertion in tests or processing in
the rest of your program.

Realize that Timely does not ensure any ordering for calls to
`captor`; so asserts in tests should be sorted by epoch and item.

Writes some Python unit tests for all the operators using this
feature. Not sure how to run them in CI, but we'll do that later.

Because we can't register multiple ctrl-c handlers, comments that
out. I believe Dan is figuring out a longer term solution to this as
part of his benchmarking / entry point changes.


whoahbot (approved):
This looks good! Do you want to fix the ctrlc handler now that it can be disabled in `build_and_run`?

davidselassie:
Yup! Done.

PR #11, opened 2022-02-10
Renaming pyexamples folder to examples and updating related paths

miccioest:
This PR renames `pyexamples` folder to `examples` and also updates related paths in python files.


awmatheson (approved):
🚀

PR #16, opened 2022-02-14
Exceptions and interrupts in multiple worker threads

davidselassie:
Ctrlc works a lot better
------------------------

Stops registering a custom ctrlc handler. Instead do what CPython
does during `Thread.join()` which is sleep for a bit then check
interrupts then check if workers are done.

If we got an interrupt, set the "graceful shutdown" flag. All the
worker threads will see this and stop.

Something I didn't realize is that Timely _keeps spinning_ after the
worker main finishes until the dataflows are dropped. So does that.

To communicate when workers are done, increment an atomic counter that
the main thread checks. There's a nightly method
`JoinHandle::is_running` we could maybe use one day.

Shutdown all worker threads when one has error
----------------------------------------------

Use a panic hook to see if a worker thread had a panic. If so, tell
all other workers to gracefully shutdown. Panic hooks are a weird
per-process global thing, so use a lib's `defer!` macro to somewhat
gracefully unregister them. This isn't great but unlike the global
ctrlc handler I think will work ok because hopefully there's no other
Rust code running in some other Python lib that needs this!

`guards.join()` will return if any of the worker threads panic'd so
raise a generic exception in the main Python thread.

Print nice Python tracebacks
----------------------------

My "holy grail" would be to re-raise Python exceptions from a worker
thread in the main thread. Towards this end, modify `with_traceback!`
to instead of sending a string with the exception, try straight
attaching the entire `PyErr` exception. Usually, we can access
whatever object this is from `handle.join()` and then we could
re-raise it, but Timely unceremonially turns that into a debug string,
so we can't recreate it.

Instead, nicely print the traceback in the panic hook. Good enough for
now.

Future work
-----------

I'd eventually like to replicate this behavior for the multi-process
case, not just the multi-threaded case as here. We'll have to see what
our options for a "control plane" IPC is and if we can piggy-back on
what Timely uses.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,6 +1,8 @@
 extern crate rand;
+#[macro_use(defer)]
+extern crate scopeguard;
 
-use pyo3::exceptions::PyKeyboardInterrupt;
+use pyo3::exceptions::PyRuntimeError;
 use pyo3::exceptions::PyTypeError;
 use pyo3::prelude::*;
 use pyo3::types::*;
@@ -12,6 +14,7 @@ use std::hash::Hash;
 use std::hash::Hasher;
 use std::ops::Deref;
 use std::sync::atomic::AtomicBool;
+use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering;
 use std::sync::Arc;
 use std::thread;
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -947,115 +950,129 @@ impl Executor {
...
-                        all_probes.push(probe);
+            .try_build()
+            .map_err(PyRuntimeError::new_err)?;
+
+            let should_shutdown = Arc::new(AtomicBool::new(false));
+            let should_shutdown_w = should_shutdown.clone();
+            let should_shutdown_p = should_shutdown.clone();
+            // We can drop these if we want to use nightly
+            // Thread::is_running
+            // https://github.com/rust-lang/rust/issues/90470
```
blakestier, line 991: When you were deciding on this approach over the `is_running`, what were the trade-offs?

davidselassie: We can't use `is_running` without flipping our whole Rust tool chain over to using nightly builds. I'm concerned about the stability guarantees of that. So this wasn't really an option, but I wanted to note it in case it does become stable eventually.

whoahbot (approved):

PR #21, opened 2022-02-18
Adding a notebook example

awmatheson:
Converted wikistream into a notebook example.


whoahbot (commented):
This example requires a few external dependencies, and doesn't stop running until terminated. Is there a more straight-forward example that doesn't operate on an endless stream that would be easier to see for a notebook user?

PR #22, opened 2022-02-24
Restructure Python imports

whoahbot:
- Uses `__all__` to provide an explicit index of the packagewhen doing `from bytewax import *`.


pysrc/bytewax/__init__.py
```diff
@@ -1,177 +1,13 @@
...
-    spawn_cluster(
-        flow, input_builder, output_builder, proc_count, worker_count_per_proc
-    )
-
-    return out
+from .bytewax import Dataflow, cluster_main, sleep_keep_gil, sleep_release_gil
+
+from .execution import run, spawn_cluster, run_cluster
+
+__all__ = [
```
davidselassie, line 5: I'm not super familiar with this. Is there a specific behavior you're hoping to cause or prevent with `__all__`? Otherwise isn't importing a symbol into `__init__.py` enough to cause it to be included in `from bytewax import *`?

davidselassie (approved):
Looks good. One question.

pysrc/bytewax/__init__.py
whoahbot: You are correct in that importing a symbol into `__init__.py` is enough to include it. I was attempting to carve out a few of the parts of our public API that were pulled in, like `_run`, but I just tested it and it didn't end up working anyway.

I think on closer inspection, I should skip using `__all__`. I'll update it.

blakestier (approved):
👏  Yay

PR #23, opened 2022-02-24
Tests ability to interrupt execution

davidselassie:
Adds some bespoke tests which send SIGINT and ensure that the
functions raise a `KeyboardInterrupt`.


davidselassie:
Heh. Of course Windows wouldn't support signals. Let me see if there's an easy way to test this in a portable way. Thanks CI!

whoahbot (approved):
Seems like Windows is still not working. Looks good otherwise!

pytests/test_run.py
```diff
@@ -65,3 +71,95 @@ def boom(item):
...
+        run(flow, inp.fully_ordered(range(1000)))
+
+    killer.join()
+
+
+def test_run_cluster_can_be_sigint():
+    manager = multiprocessing.Manager()
+    is_running = manager.Event()
+    test_pid = os.getpid()
+    if os.name == "nt":
```
whoahbot, line 108: Maybe this check isn't firing. We could try https://docs.python.org/3.7/library/platform.html#platform.system

davidselassie:
Looks like the origin of (some) of this craziness on Windows is: https://docs.microsoft.com/en-us/windows/console/ctrl-c-and-ctrl-break-signals

> By default, these signals are passed to all console processes that are attached to the console.

So trying to send `CTRL_C_EVENT` to the subprocess under test would also send it to the pytest invocation itself and cancel the test run.

One might be able to get around this by dropping down to using `subprocess.Popen` instead of `multiprocessing.Process` and https://docs.python.org/3/library/subprocess.html#subprocess.CREATE_NEW_PROCESS_GROUP but I don't want to deal with that now so just skipping these tests on Windows.

blakestier (approved):

PR #26, opened 2022-02-25
Adds an "order" input helper and allows tumbling "event time"

davidselassie:
Order
-----

Timely, and thus Bytewax, requires that the input timestamp / epoch
never "goes backwards". This is because Timely is able to know that it
can kick off final computation once a timestamp has closed.

Our current "input is an iterator of `(epoch, item)`" API doesn't have
the flexibility to allow multiple "open" epochs, although Timely
supports this.

To help bridge this gap, a common use case is: allow late data in a
"buffering window" and drop anything after that. Here's an
`ordering()` function which implements that and you can wrap your
source input iterator in it.

Event Time
----------

Allows `tumbling_epoch` to take a `time_getter` argument which will
produce the time for each item from the data within, instead of always
just looking at the wall clock time.

This will let you bucket into epoch of a delayed data stream. Also
plays well with `order`.

Other
-----

Also renames `inp` to `inputs` since we're renaming everything now.


whoahbot (approved):
Looks good to me. I'm not sure why CI is failing. It's not able to import `bytewax.inputs`.

Do we need to add: `from .inputs import *` in `__init__.py` ?

davidselassie:
I'm trying to recreate this failure locally. I rmd the virtualenv and made it again and removed `__pycache__` and can't get these failures locally. The tests pass for me. Any thoughts?

I thought we only need to import it if we want a local name in the main `bytewax` space. Otherwise to import any submodule, you'd have to import all parent modules, which I don't think is true.

pysrc/bytewax/inputs.py
```diff
@@ -0,0 +1,204 @@
...
+
+        wrap_iter: Existing input iterable of just items.
+
+    Yields: Tuples of (epoch, item).
+    """
+    for item in wrap_iter:
+        yield (0, item)
+        
+
+def order(
```
awmatheson, line 29: This is great, I think this is a good way to handle late-arriving and out-of-order data on the input side of things. TIL about heapq built-in

awmatheson:
I like the rename of `inp` to `inputs`!

pysrc/bytewax/inputs.py
```diff
@@ -0,0 +1,204 @@
...
+        else:
+            epoch = int((time - epoch_0_start_time) / epoch_length)
+
+        yield (epoch, item)
+
+                      
+def fully_ordered(wrap_iter: Iterable) -> Iterable[Tuple[int, any]]:
+    """Each input item increments the epoch.
+
+    Be carful using this in high-volume streams with many workers, as
```
blakestier, line 189: while you're here, mind fixing this typo?

pysrc/bytewax/inputs.py
```diff
@@ -0,0 +1,204 @@
...
+        time_getter: Callable[[Any], Any] = itemgetter(0),
+        on_drop: Callable[[Any], None] = None,
+) -> Iterable[Tuple[int, Any]]:
+    """Order out-of-order values within a sliding window.
+
+    Values are emitted in order after seeing a new time that would
+    push them out of the sliding window. Values that arrive after they
+    would have been emitted are too late and are dropped.
+
+    Use if you have a possibly out-of-order data source as Bytewax
```
blakestier, line 41: lowercase bytewax? i don't remember where we landed on this

pysrc/bytewax/inputs.py
```diff
@@ -0,0 +1,204 @@
...
+    ...         datetime.timedelta(seconds=2),
+    ...         itemgetter("timestamp"),
+    ...     )
+    ... ]
+    [(0, 'a'), (0, 'b'), (2, 'c')]
+
+    By default, uses "ingestion time" and you don't need to specify a
+    way to access the timestamp in each item.
+
+    >>> import pytest; pytest.skip("Figure out sleep in test.")
```
blakestier, line 150: Should this be here?

pysrc/bytewax/inputs.py
```diff
@@ -0,0 +1,204 @@
...
+    Yields: Tuples of (epoch, item).
+    """
+    for item in wrap_iter:
+        yield (0, item)
+        
+
+def order(
+        wrap_iter: Iterable,
+        window_length: Any,
+        time_getter: Callable[[Any], Any] = itemgetter(0),
```
blakestier, line 32: Dropping the default in favor of users needing to specify shape of data (can't pass in a hash without defining `itemgetter`

pysrc/bytewax/inputs.py
```diff
@@ -0,0 +1,204 @@
...
+    ...         itemgetter("timestamp"),
+    ...     )
+    ... ]
+    ['b', 'c']
+
+    Args:
+    
+        wrap_iter: Existing input iterable.
+   
+        window_length: Maximum out-of-order-ness before dropping
```
blakestier, line 73: Trying to brainstorm the affirmative version of this to communicate "interval of opportunity".

pysrc/bytewax/inputs.py
```diff
@@ -0,0 +1,204 @@
...
+            values.
+
+        time_getter: Function to call to produce a timestamp for each
+            value. Defaults to using the first item in each value;
+            e.g. `(time, x)`.
+
+        on_drop: Function to call with each dropped item. Defaults to
```
blakestier, line 80: Worth suggesting logging?

pysrc/bytewax/inputs.py
```diff
@@ -0,0 +1,204 @@
...
+    """
+    buffer_heap = []
+    buffer_close_time = None
+    for item in wrap_iter:
+        time = time_getter(item)
+        if buffer_close_time is None or time >= buffer_close_time:
+            heapq.heappush(buffer_heap, (time, item))
+
+            buffer_close_time_item = time - window_length
+            if buffer_close_time is None or buffer_close_time_item > buffer_close_time:
```
blakestier, line 93: Potentially check times instead

pysrc/bytewax/inputs.py
```diff
@@ -0,0 +1,204 @@
...
+                buffer_close_time = buffer_close_time_item
+        
+                while len(buffer_heap) > 0 and buffer_heap[0][0] < buffer_close_time:
+                    time, item = heapq.heappop(buffer_heap)
+                    yield item
+            # Else: This item didn't push forward the window, so no
```
blakestier, line 99: Move above calculation?

pysrc/bytewax/inputs.py
blakestier, line 29: I think we should consider renaming this or `fully_ordered`. Sort? `sorted_window`?

pysrc/bytewax/inputs.py
```diff
@@ -0,0 +1,204 @@
...
+    ...         items,
+    ...         datetime.timedelta(seconds=2),
+    ...         itemgetter("timestamp"),
+    ...     )
+    ... ]
+    ['b', 'c']
+
+    Args:
+    
+        wrap_iter: Existing input iterable.
```
blakestier, line 71: could this just be `items`? I know it's not from this pr though 🤷

pytests/test_inputs.py
```diff
@@ -1,44 +1,96 @@
...
+            (0, "b"),
+            (0, "c"),
         ]
     )
 
 
-def test_fully_ordered():
-    def add_one(item):
-        return item + 1
+def test_order():
```
blakestier, line 24: thanks for tests 😭

blakestier (commented):
Thank you!  A handful of small suggestions, the most pressing I think being a rename of `order` to distinguish it from `fully_ordered`.

pysrc/bytewax/inputs.py
davidselassie: I've been trying to keep all the doctests executable without errors, but this one doesn't actually work because of the wall time stuff. So this marks it as "skip" so we know to fix it later.

blakestier (approved):

PR #28, opened 2022-02-25
Maturin develop before running tests.

whoahbot:
Trying to fix inconsistent issues when running tests.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -2,7 +2,9 @@ name: CI
 
 on:
   push:
-  pull_request:
+    branches:
+    - main
+  pull_request: {}
 
 jobs:
   linux:
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -16,16 +16,20 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
         architecture: x64
+    - name: Run tests
+      shell: bash
+      run: |
+        python -m venv venv
+        source venv/bin/activate
+        pip install -U pip
+        pip install -r build.requirements.txt
+        maturin develop
+        python -m pytest
     - uses: messense/maturin-action@v1
       with:
         manylinux: auto
         command: build
         args: --release -o dist --interpreter python${{ matrix.python-version }}
-    - name: Run tests
-      run: |
-        pip install bytewax --find-links dist -v --pre
-        pip install --upgrade pytest==7.0.0
-        python -m pytest pytests/
     - name: Upload wheels
       uses: actions/upload-artifact@v2
       with:
@@ -37,7 +41,7 @@ jobs:
     strategy:
       matrix:
         python-version: ['3.7', '3.8', '3.9', '3.10']
-    steps: 
+    steps:
     - uses: actions/checkout@v2
     - uses: actions/setup-python@v2
       with:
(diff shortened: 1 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -16,16 +18,20 @@ jobs:
...
       with:
         manylinux: auto
         command: build
         args: --release -o dist --interpreter python${{ matrix.python-version }}
-    - name: Run tests
-      run: |
-        pip install bytewax --find-links dist -v --pre
-        pip install --upgrade pytest==7.0.0
-        python -m pytest pytests/
     - name: Upload wheels
```
davidselassie, line 35: Does this upload wheels at all on PR runs? Should that only happen on tagged releases?

.github/workflows/CI.yml
```diff
@@ -16,16 +18,20 @@ jobs:
...
+    - name: Run tests
+      shell: bash
+      run: |
+        python -m venv venv
+        source venv/bin/activate
+        pip install -U pip
+        pip install -r build.requirements.txt
+        maturin develop
+        python -m pytest
     - uses: messense/maturin-action@v1
```
davidselassie, line 30: To prevent the "double build" thing, could you put this step before running tests then tell pip to install from the already packaged wheel on the filesystem? I assume it's still hanging out on the worker somewhere there?

.github/workflows/CI.yml
```diff
@@ -2,7 +2,9 @@ name: CI
 
 on:
   push:
-  pull_request:
+    branches:
+    - main
+  pull_request: {}
```
davidselassie, line 7: What does this mean? Run all these actions on only `main` pushes and on any PR?

davidselassie (approved):
Thanks! This seems to work, so great! I have a few questions because I haven't really looked at this CI file before in depth. We can merge this now to fix up other tests and refine later, though.

.github/workflows/CI.yml
whoahbot: It only happens during tagged releases because of the filter here:

```
    if: "startsWith(github.ref, 'refs/tags/')"
```

whoahbot: It only runs on GitHub events that are ragged because of this filter:

```
    if: "startsWith(github.ref, 'refs/tags/')"
```

.github/workflows/CI.yml
whoahbot: That's what we were doing before. There are a couple problems with that approach that we ran into (not that they can't be solved)

The primary issue is that pip doesn't enforce an order of preference when looking for versions to install:

https://pip.pypa.io/en/latest/cli/pip_install/#finding-packages

> pip looks for packages in a number of places: on PyPI (if not disabled via --no-index), in the local filesystem, and in any additional repositories specified via --find-links or --index-url. There is no ordering in the locations that are searched. Rather they are all checked, and the “best” match for the requirements (in terms of version number - see [PEP 440](https://www.python.org/dev/peps/pep-0440) for details) is selected.

So we kept running into the issue where we were installing from PyPI instead of the local wheel we had built.

You can disable installing packages from PyPI with `--no-index`, but we need to allow using pypi due to the dependency we have on `multiprocess`.

.github/workflows/CI.yml
whoahbot: Yes, only run these actions on pushes to `main`, and on any `pull_request` event.

blakestier (approved):

PR #29, opened 2022-02-28
Debug operators

blakestier:
First pass adding some debugging to our operators.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -12,3 +12,15 @@ macro_rules! with_traceback {
         }
     };
 }
+
+#[macro_export]
+macro_rules! log_func {
+    () => {{
+        fn f() {}
+        fn type_name_of<T>(_: T) -> &'static str {
+            std::any::type_name::<T>()
+        }
+        let name = type_name_of(f);
+        &name[..name.len() - 3]
+    }};
+}
--- a/src/operators/mod.rs
+++ b/src/operators/mod.rs
@@ -4,29 +4,37 @@ use pyo3::prelude::*;
 
 use crate::pyo3_extensions::{TdPyAny, TdPyCallable, TdPyIterator};
 use crate::with_traceback;
+use crate::log_func;
+
+use log::debug;
 
 // These are all shims which map the Timely Rust API into equivalent
 // calls to Python functions through PyO3.
 
 pub(crate) fn map(mapper: &TdPyCallable, item: TdPyAny) -> TdPyAny {
+    debug!("{}, mapper:{:?}, item:{:?}", log_func!(), mapper, item);
     Python::with_gil(|py| with_traceback!(py, mapper.call1(py, (item,))).into())
 }
 
 pub(crate) fn flat_map(mapper: &TdPyCallable, item: TdPyAny) -> TdPyIterator {
+    debug!("{}, mapper:{:?}, item:{:?}", log_func!(), mapper, item);
     Python::with_gil(|py| with_traceback!(py, mapper.call1(py, (item,))?.extract(py)))
 }
 
 pub(crate) fn filter(predicate: &TdPyCallable, item: &TdPyAny) -> bool {
+    debug!("{}, predicate:{:?}, item:{:?}", log_func!(), predicate, item);
     Python::with_gil(|py| with_traceback!(py, predicate.call1(py, (item,))?.extract(py)))
 }
 
 pub(crate) fn inspect(inspector: &TdPyCallable, item: &TdPyAny) {
+    debug!("{}, inspector:{:?}, item:{:?}", log_func!(), inspector, item);
     Python::with_gil(|py| {
         with_traceback!(py, inspector.call1(py, (item,)));
     });
 }
 
 pub(crate) fn inspect_epoch(inspector: &TdPyCallable, epoch: &u64, item: &TdPyAny) {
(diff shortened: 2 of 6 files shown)


whoahbot (approved):
This looks great! Thanks for putting this together.

src/operators/mod.rs
```diff
@@ -70,6 +80,7 @@ pub(crate) fn reduce_epoch(
     _key: &TdPyAny,
     value: TdPyAny,
 ) {
+    debug!("{}, reducer:{:?}, key:{:?}, value:{:?}, agg:{:?}", log_func!(), reducer, _key, value, aggregator);
```
davidselassie, line 83: Might be nice to also log out the updated aggregator after the reducer call, similar to stateful map.

src/macros.rs
```diff
@@ -12,3 +12,15 @@ macro_rules! with_traceback {
...
 }
+
+#[macro_export]
+macro_rules! log_func {
+    () => {{
+        fn f() {}
+        fn type_name_of<T>(_: T) -> &'static str {
+            std::any::type_name::<T>()
+        }
+        let name = type_name_of(f);
```
davidselassie, line 23: It looks like the goal of this macro is to get the name of the containing function? How does this work?

src/operators/mod.rs
```diff
@@ -40,6 +48,7 @@ pub(crate) fn reduce(
     value: TdPyAny,
 ) -> (bool, impl IntoIterator<Item = TdPyAny>) {
     Python::with_gil(|py| {
+        debug!("{}, reducer:{:?}, key:{:?}, value:{:?}", log_func!(), reducer, key, value);
```
davidselassie, line 51: Aggregator before and after?

src/operators/mod.rs
```diff
@@ -4,29 +4,37 @@ use pyo3::prelude::*;
...
 use crate::with_traceback;
+use crate::log_func;
+
+use log::debug;
 
 // These are all shims which map the Timely Rust API into equivalent
 // calls to Python functions through PyO3.
 
 pub(crate) fn map(mapper: &TdPyCallable, item: TdPyAny) -> TdPyAny {
+    debug!("{}, mapper:{:?}, item:{:?}", log_func!(), mapper, item);
```
davidselassie, line 15: It would be nice to log out epoch as well in all of these so you can flip a single logging switch and see a complete trace of execution. But that would be a bigger change because it would require changing the Timely operators we use and re-jiggering some of the shim functions, so I don't think we need to do it now. Just a note for posterity!

src/operators/mod.rs
```diff
@@ -57,6 +66,7 @@ pub(crate) fn reduce(
 
         if should_emit_and_discard_aggregator {
             let emit = (key.clone_ref(py), updated_aggregator).to_object(py).into();
+            debug!("{}, reducer:{:?}, key:{:?}, emit:{:?}, complete:{:?}", log_func!(), reducer, key, emit, should_emit_and_discard_aggregator);
```
davidselassie, line 69: This might better reflect the logic here if it logged the call to `is_complete` and its arguments? Right now I can't tell what function is being called nor what its arguments are, which is the meat of the second step of the reduce operator.

davidselassie (commented):
Thanks for writing these up! I have one question on the macro and some thoughts about what might be relevant to log. But if you want to tackle these later, that's fine too.

src/macros.rs
blakestier: That is the goal although [the rust function](https://doc.rust-lang.org/std/any/fn.type_name.html) isn't as precise as I'd like.

src/operators/mod.rs
blakestier: You mean what `is_complete` function is being called?

src/macros.rs
blakestier: Candidly, the internet helped me with this, but I believe you can wrap the function in your own code and thereby access the type of the function which resolves to the fully qualified name.

src/operators/mod.rs
```diff
@@ -106,11 +181,25 @@ pub(crate) fn stateful_map(
     key: &TdPyAny,
     value: TdPyAny,
 ) -> (bool, impl IntoIterator<Item = TdPyAny>) {
+    debug!(
+        "{}, mapper:{:?}, key:{:?}, value:{:?}",
```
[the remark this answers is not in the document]
blakestier: Hmm, yes. We should

src/operators/mod.rs
```diff
@@ -40,6 +63,15 @@ pub(crate) fn reduce(
     value: TdPyAny,
 ) -> (bool, impl IntoIterator<Item = TdPyAny>) {
     Python::with_gil(|py| {
+        debug!(
+            "{}, reducer:{:?}, key:{:?}, value:{:?}, is_complete:{:?}(agg={:?})",
```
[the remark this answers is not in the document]
blakestier: wow that doesn't make sense at all now that I re-read it... thanks for the reiteration!

davidselassie (approved):
Great! Thanks

src/macros.rs
davidselassie: I think that makes sense. Hence the slice to `-3` to remove `f()` as the last step.

PR #31, opened 2022-03-04
Async execution

davidselassie:
_AKA in which I rewrite the execution model again._

Since this was exploratory and coordinates changes to lots of pieces,
I couldn't really break it up into smaller PRs or commits. But it
barely changes the user-facing API! So that's something.

Background
==========

Previously, `run()` and `run_cluster()` would fully turn their inputs
into lists before running the dataflow upon them. This was easy to
code, but also means we didn't support infinite streams or things that
wouldn't all fit in memory. It would be super cool if you could run a
dataflow on an un-ending metrics stream in a notebook, e.g.

This change table-flips the execution of workflows to allow that to
happen. It seems to work, but ups the complexity of our execution code
quite a bit: you now have to know nitty gritty details about
generators, coroutines, thread pools, process pools, why nobody uses
`dill`, IPC queues, unwind logic safety.

I put forth the question: do you think it's worth it?

Implementation
==============

Coroutine Main
--------------

The heart of the whole operation is `WorkerCoro` which is a PyO3 class
which implements Python's `Coroutine` interface. `WorkerCoro.send()`
is the new "main loop" for each worker doing all the same things as
before: pump pumps, step Timely, check probes for output. But it's not
a loop, just one iteration. It will be "looped" by whatever is running
the coroutine allowing other code to run in-between calls.

Because this requires persisting the Timely worker in a PyO3 class,
there are some gotchas surrounding panics and `allow_threads` that are
commented upon in the code.

`WorkerCoro::compile` what builds a `WorkerCoro` from the `Dataflow`
and IO builders, and, crucially, a Timely Worker.

There are also a few other shim methods to match the `Coroutine`
interface with minimal implementations.

Worker Builder
--------------

Timely is very clever. It takes advantage of closures, threads,
ownership, the send trait to ensure that `execute` and
`execute_directly` both magically start up the right number of worker
and background communication threads without race conditions and the
ability to accidentally perform operations in the wrong order.

Because we're wrestling control of the execution of the worker back
from Timely, we have to learn and re-implement some of that internal
magic. But also, because we're playing in Python-land, we don't have
the nice guarantees that Rust gives us to ensure that it's done right.

A `WorkerBuilder` contains the coordination information to start up a
single worker thread. We piggy back on
`timely::CommunicationConfig::try_build()` to generate all the
builders we need to setup a cluster (or just one for single-threaded
mode). Once we have those builders, we can make a worker thread for
each of them, send them to those threads, then actually construct the
worker coroutine with the dataflow we care about.

`run` Generator
---------------

The `WorkerCoro` is all we need to implement running a dataflow in the
current thread. We use the generator as the "run loop" for that
coroutine, stepping it manually and yielding output.

For cluster mode, we need a few more concepts:

Communication Threads
---------------------

A sort of secret thing that Timely did for us when it managed
execution, was start up background threads to communicate with other
processes over the network in cluster mode. We still delegate the
creation of those threads to Timely (they are made at the same time as
the builders above), but we need to wrap them in our own Python object
so we can join them when work is complete.

The `CommsThreads` class encapsulates that.

Graceful Shutdown
-----------------

If you remember in our previous "main" functions, we had some
`AtomicBool` flags to say "shut 'er down" when something went wrong in
another worker. Since the execution is now being managed from Python,
we have to move it there. There's now some `threading.Event` or
`multiprocessing.Event` flags that take on that role.

Then as a way to provide a generic interface to that in the
`WorkerCoro`, we thread around a new callback `should_stop` (or
`proc_should_stop`) which will return `True` whenever the level above
in the cluster found a problem and all workers should shutdown.

Exceptions
----------

Within a cluster process, exceptions are bubbled up from individual
worker threads through using `Future.exception()` created from a
`ThreadPoolExeuctor` in `cluster_main()`. When we see that one thread
excepted, use the event flags described above to shutdown all other
workers, then bubble the exception up to the main thread in that
process.

Then when the main thread throws an exception, the same sort of action
happens on the cluster level via the `AsyncResult` of
`multiprocess.Pool` in `spawn_cluster()`: set a flag that all cluster
processes watch and shutdown.

These two work together to allow a whole cluster to stop on error or
ctrl-c.

A Pile of Queues
----------------

To get the magic of `run_cluster()` continuously yielding output to
work there's three queues which allow us to send out input from the
input iterator and combine output from the workers back to this
process. Then calling all of those, using the output generator as the
"run loop" for the coroutine.

Ugly Bits
=========

I couldn't find much prior art on using a generator as the run loop
for a coroutine, hence all the magic surrounding `send()` and
`StopIteration`. I don't think there's a general way to do this since
most coroutines are waiting on _something_ not just a generic
"yielding of control" like we're doing.

Unfortunately, `asyncio` doesn't EAFP and instead checks if coroutine
objects are subclasses of `abc.Coroutine` and PyO3 doesn't support
subclassing Python types, so there's this `WorkerCoroWrapper` thing.

For some reason `multiprocess.Manager.Queue` doesn't support
`.close()` so I have to enqueue "finished" sentinel values.

It also feels strange to use `multiprocess.Pool` which has a
non-`asyncio` compatible API and so you don't have access to the nice
waiting functions it gives you. But we have to use it to support
`dill` pickling of lambdas in dataflows.

On that note, if we want to drop lambda support, there are some async
libraries that might streamline a lot of this code. One that looks
promising to me is
[`aiomultiprocess`](https://aiomultiprocess.omnilib.dev/en/stable/index.html);
if we discover that having multiple threads per process isn't really
worth it because of the GIL, we could drop a lot of the complexity of
coordinating threads and instead have only one pool to deal with.


src/lib.rs
```diff
@@ -27,8 +24,7 @@ fn sleep_release_gil(py: Python, secs: u64) {
 }
 
 #[pymodule]
-#[pyo3(name = "bytewax")]
-fn mod_tiny_dancer(_py: Python, m: &PyModule) -> PyResult<()> {
+fn bytewax(_py: Python, m: &PyModule) -> PyResult<()> {
```
blakestier, line 27: 😭

pysrc/bytewax/execution.py
```diff
@@ -1,20 +1,113 @@
...
+from multiprocess import Manager, Pool, TimeoutError
 
+from .bytewax import Dataflow, WorkerBuilder, WorkerCoro
 
-def run(flow: Dataflow, inp: Iterable[Tuple[int, Any]]) -> List[Tuple[int, Any]]:
-    """Pass data through a dataflow running in the current thread.
 
-    Blocks until execution is complete.
+class WorkerCoroWrapper(abc.Coroutine, abc.Iterable):
+    """We can't subclass Python-native classes in PyO3 yet.
```
blakestier, line 16: May as well make this a TODO for 🛠️  later

pysrc/bytewax/execution.py
```diff
@@ -1,20 +1,113 @@
...
+    output_builder: Callable[[int, int], Callable[[Tuple[int, Any]], None]],
+    should_stop: Callable[[], bool] = lambda: False,
+) -> None:
+    """Execute a dataflow in the current thread.
+
+    You'd commonly use this for prototyping custom input and output
+    builders with a single worker before using them in a cluster
+    setting.
+
+    Returns after execution is complete. Use with `asyncio.run()` if
```
blakestier, line 59: What does this mean to "use with asyncio.run()" ? Perhaps an example or direction to docs?

davidselassie: Yes, can put an example of that. It'd look like `asyncio.run(run_main(...))`.

pysrc/bytewax/execution.py
```diff
@@ -189,26 +411,85 @@ def run_cluster(
...
+                msg = input_buffer.pop()
+                in_q.put(msg, False)
+            except Full:
+                input_buffer.append(msg)
+                break
+
+        # 4. Get as much output as we can from workers.
+        while not out_q.empty():
+            try:
+                yield out_q.get(False)
```
davidselassie, line 493: There is a race condition here because `out_q.empty` is not guaranteed to be correct. So the cluster might push final values into the `out_q`, shutdown, but then the output won't be yielded. We'll need to properly signal shutdown through the queue.

src/execution/mod.rs
```diff
@@ -41,277 +38,485 @@ impl Pump {
...
+            .output_probes
+            .drain(..)
+            .filter(|probe| !probe.done())
+            .collect();
+
+        let should_stop = self.should_stop.call0(py)?.extract(py)?;
+        let work_complete = self.input_pumps.is_empty() && self.output_probes.is_empty();
+
+        if should_stop || work_complete {
+            self.close()?;
```
davidselassie, line 292: We always have to shutdown the worker threads before shutting down comms threads.

src/execution/mod.rs
```diff
@@ -41,277 +38,485 @@ impl Pump {
...
+            Err(PyStopIteration::new_err(()))
+        } else {
+            Ok(())
+        }
+    }
+
+    /// Run code on shutdown of worker.
+    fn close(&mut self) -> PyResult<()> {
```
davidselassie, line 300: `close()` is described as "raising `GeneratorExit` in the code, and the result of that is shutting down. Hence, no exceptions here.

pysrc/bytewax/execution.py
```diff
@@ -39,68 +141,174 @@ def run(flow: Dataflow, inp: Iterable[Tuple[int, Any]]) -> List[Tuple[int, Any]]
...
 
-    return out
+    worker_raised_ex = threading.Event()
+
+    def worker_should_stop():
+        return proc_should_stop() or worker_raised_ex.is_set()
+
+    def worker_main(builder):
+        coro = builder.build(flow, input_builder, output_builder, worker_should_stop)
+        asyncio.run(WorkerCoroWrapper(coro))
```
davidselassie, line 251: It is possibly there is some issue with `WorkerCoro` being unsendable and the Python garbage collector running in the main thread. Sometimes we got spurious errors.

davidselassie (commented):
I'm not totally convinced that this is a good idea. I did some followup work last week to make this not use a hot loop and that ended up reducing my confidence that this doesn't have latent bugs in it. And it looks like we're not getting a ton of benefit from the external API changes. I'm going to add some comments here about what I was thinking, then close this PR. We can resurrect it if we need to, and take our learnings about coroutines and asyncio into the future.

PR #34, opened 2022-03-09
Testable examples

davidselassie:
Adds some pytest hook code to "run" our Markdown example
documentation.
    
This will take each Markdown file in the `docs/` directory, parse it,
and run all code blocks in each file in order. The stdout of each code
block is then compared to an expected value in an optional following
`{testoutput}` block. If the output is missing, the output is assumed
to be nothing.

Eventually, I'd like to start porting our blog posts over to this format.
This post also will need some fixups to interface with Konrad's work
so links are translated properly and we can link to other
documentation. The big thing I wanted to do here was get these tests
working.

### DIFF
--- a/docs/examples/search_session.md
+++ b/docs/examples/search_session.md
@@ -1,0 +1,309 @@
+Building Sessions from Search Logs
+==================================
+
+Here is a basic example of using Bytewax to turn an incoming stream of
+event logs from a hypothetical search engine into metrics over search
+sessions. In this example, we're going to focus on the dataflow itself
+and aggregating state, and gloss over details of building this
+processing into a larger system.
+
+Schema
+------
+
+Let's start by defining a data model / schema for our incoming
+events. We'll make a little model class for all the relevant events
+we'd want to monitor.
+
+```python
+from dataclasses import dataclass
+from typing import List
+
+
+@dataclass
+class AppOpen:
+    user: int
+
+
+@dataclass
+class Search:
+    user: int
+    query: str
+
+
+@dataclass
+class Results:
+    user: int
+    items: List[str]
+
+
+@dataclass
+class ClickResult:
+    user: int
+    item: str
+
+
+@dataclass
+class AppClose:
+    user: int
+
+
+@dataclass
+class Timeout:
+    user: int
+```
+
+In a more mature system, these might come from external schema or be
+auto generated.
+
+Epochs
+------
+
+Now that we've got those, here's a small dump of some example data you
+could imagine comming from your app's events infrastructure.
+
+```python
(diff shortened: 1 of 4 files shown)


davidselassie:
Powered by https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html

Also using https://myst-parser.readthedocs.io/en/latest/index.html for Markdown

docs/examples/search_session.md
```diff
@@ -0,0 +1,309 @@
...
+def remove_key(user_event):
+    user, event = user_event
+    return event
+
+
+flow.map(remove_key)
+```
+
+Our next task is to split user sessions into search sessions. To do
+that, we'll use the [flat map operator](operators/flat-map), which
```
whoahbot, line 202: Should we normalize flat map to look like the actual operator?

```suggestion
that, we'll use the [flat_map operator](operators/flat-map), whichthat, we'll use the [flat map operator](operators/flat-map), which
```

docs/examples/search_session.md
```diff
@@ -0,0 +1,309 @@
...
+
+Now we're done with defining the dataflow. Let's run it!
+
+Execution
+---------
+
+[Bytewax provides a few different entry points for executing your
+dataflow](execution), but because we're focusing on the dataflow in
+this example, we're going to use `bytewax.run` which is the most basic
+one that pushes items in an iterator through the dataflow in-thread.
```
whoahbot, line 277:
```suggestion
execution model that pushes input items in an iterator through the dataflow
```

docs/examples/search_session.md
```diff
@@ -0,0 +1,309 @@
...
+Since the [capture](operators/capture) step is immediately after calculating CTR,
+we should see one output item for each search session. That checks
+out! There were three searches in the input: "dogs", "cats", and
+"fruit". Only the first two resulted in a click, so they contributed
+`1.0` to the CTR, while the no-click search contributed `0.0`.
+
+Now let's talk about the output epochs briefly. In general, operators
+do not modify the epoch unless they explicitly say so. The only
+operator that deals with the epoch here is [reduce](operators/reduce)
+which emits output at the timestamp of the past-processed input. Since
```
whoahbot, line 306: I stumbled on this sentence when I first read it, but I'm having difficulty thinking of a better formulation. We may want to remove the note about modifying the epoch, it seems like a more advanced topic to cover elsewhere.

We do re-introduce the word `timestamp` here when we started the paragraph with `epoch`. I think this may be adding to the confusion.

davidselassie: I gave it a reworking. Hopefully it's a little clearer!

PR #35, opened 2022-03-11
modify inputs

awmatheson:
Allow the tumbling epoch to start from an arbitrary number. 

This will allow people to set the epoch if they are tracking progress.

### DIFF
--- a/pysrc/bytewax/inputs.py
+++ b/pysrc/bytewax/inputs.py
@@ -29,8 +29,8 @@ def tumbling_epoch(
     wrap_iter: Iterable,
     epoch_length: Any,
     time_getter: Callable[[Any], Any] = lambda _: datetime.datetime.now(),
-    epoch_0_start_time: Any = None,
-    epoch_start: Any = 0,
+    epoch_start_time: Any = None,
+    epoch_start: int = 0,
 ) -> Iterable[Tuple[int, Any]]:
     """All inputs within a tumbling window are part of the same epoch.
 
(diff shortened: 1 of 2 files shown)


pysrc/bytewax/inputs.py
```diff
@@ -30,6 +30,7 @@ def tumbling_epoch(
     epoch_length: Any,
     time_getter: Callable[[Any], Any] = lambda _: datetime.datetime.now(),
     epoch_0_start_time: Any = None,
+    epoch_start: Any = 0,
```
davidselassie, line 33: We don't support any other type of epoch than `int` so you could make the type hint `: int`.

pysrc/bytewax/inputs.py
```diff
@@ -30,6 +30,7 @@ def tumbling_epoch(
     epoch_length: Any,
     time_getter: Callable[[Any], Any] = lambda _: datetime.datetime.now(),
     epoch_0_start_time: Any = None,
```
davidselassie, line 32: Rename this `first_epoch_start_time` and the other one `first_epoch`? Or something to clarify its relationship to the new argument?

davidselassie, line 32: Also not sure if it's helpful here, but another trick you could use is to make the start time constant between runs of the dataflow. You could pick some fixed value that you know is earlier than any data timestamp. Then the epoch will be re-calculated to be the same each time.

davidselassie (approved):
Looks good. Some thoughts that come to mind if you want.

davidselassie (approved):

PR #38, opened 2022-03-15
Runs doctests in CI

davidselassie:
Fixes up all the doctests. Gets them running under `pytest`. Adds that to CI.

We have to jump through two hoops:

1. pytest finds doctests by walking the source tree. But when it actually tries to execute the doctests, because we've built and installed the wheel locally to avoid "double building", that wheel will be what gets imported when `import bytewax` is called. pytest's import machinery has a check to see that the expected code under test is imported, so it fails. There's an escape hatch via the `PY_IGNORE_IMPORTMISMATCH` env var. See https://github.com/pytest-dev/pytest/issues/2042 . The only way around this I could see would be to go back to 2-step builds: run `maturin develop` so that the local code is both what is imported and walked for doctests, then separately run `maturin build` and send off that wheel. You do not need this env var when running locally because you'll be doing local development via `maturin develop` (unless you do install a wheel).

2. For some reason, anything involving `run_cluster()` is not happy on Windows in CI. I've been able to run all the tests successfully locally on a Windows machine, so I don't know what's going on. Something about process spawning in containers? I don't want to deal with it now, so skip all the tests that use `spawn_cluster()` under the hood only on Windows GitHub Actions.

Does a little bit of cleanup to the input helper doctests to show how the functions would be used in a dataflow, not just in isolation.

### DIFF
--- a/pysrc/bytewax/execution.py
+++ b/pysrc/bytewax/execution.py
@@ -5,6 +5,12 @@ from multiprocess import Manager, Pool
 from .bytewax import *  # This will import PyO3 module contents.
 
 
+def __fix_pickling_in_doctest():
+    import dill.settings
+
+    dill.settings["recurse"] = True
+
+
 def run(flow: Dataflow, inp: Iterable[Tuple[int, Any]]) -> List[Tuple[int, Any]]:
     """Pass data through a dataflow running in the current thread.
 
@@ -71,6 +77,7 @@ def spawn_cluster(
     See `cluster_main()` for starting one process in a cluster in a
     distributed situation.
 
+    >>> __fix_pickling_in_doctest()
     >>> flow = Dataflow()
     >>> flow.capture()
     >>> def input_builder(worker_index, worker_count):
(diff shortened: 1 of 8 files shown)


pysrc/bytewax/execution.py
```diff
@@ -71,7 +84,10 @@ def spawn_cluster(
     See `cluster_main()` for starting one process in a cluster in a
     distributed situation.
 
+    >>> __skip_doctest_on_win_gha()
```
whoahbot, line 87: Does the "dunder" prefix here prevent them from showing up in documentation?

whoahbot (approved):
:+1: Left you one question out of curiosity.

pysrc/bytewax/execution.py
davidselassie: Unfortunately, no. I just wanted some way to say "don't use this function." I'll think if there's a better way to fix this up, but I can't think of anything for now.

PR #39, opened 2022-03-16
API Docs

davidselassie:
Going to use [`pdoc`](https://pdoc3.github.io/pdoc/doc/pdoc/) for API doc generation. It walks through modules, classes, and docstrings and uses templates to build HTML files.

Takes a pass over all the docstrings to clean up formatting and links between names. Also adds inspect-able function signatures to the PyO3 rust functions so they will be used by pdoc.

Adds an `apidocs` dir which will hold any custom `templates/` ([See the pdoc docs for what we can add here.](https://pdoc3.github.io/pdoc/doc/pdoc/#custom-templates)) and store final `html/` output.

Konrad's updated docs page will use the GitHub API to pull these built docs for display from the correct version tags / branches.

I eventually did not decide on the route of "CI builds these docs and commits them" because:

* Do we run tests after the docs commit or before? We either can't run the tests locally that CI is running (because we don't have access to the commit yet) or we are releasing a commit that hasn't been tested. I don't expect the doc builder to touch other stuff but I don't like the idea of not having the true release commit locally.

* If we build docs on every commit to a branch or PR, we'd need some hacky way of detecting "this was a docs update commit, so don't run again in a loop". I'm sure we could figure something out, but didn't want to fight with CI longer.

For now, my recommended process is to run the docs build manually, either as part of API-modifying commits, or perhaps as part of a "release ceremony" where we do things like bump the version number and changelog and add a Git tag, etc. It feels like we already have some manual ceremony here so I'm not too worried. If this ends up being too annoying, we can do something more structured later.


whoahbot (approved):
> It feels like we already have some manual ceremony here so I'm not too worried. If this ends up being too annoying, we can do something more structured later.

Sounds good to me.

PR #40, opened 2022-03-21
Apidocs templates

konradsienkowski:
As we want to render API Docs in our new Vue-based documentation website, we need to adjust their source code.


apidocs/templates/config.mako
```diff
@@ -0,0 +1,26 @@
...
+    # Show collapsed source code block next to each item.
+    # Disabling this can improve rendering speed of large modules.
+    show_source_code = True
+
+    # A prefix to use for every HTML hyperlink in the generated documentation.
+    # No prefix results in all links being relative.
+    link_prefix = ''
+
+    # Enable syntax highlighting for code/source blocks by including Highlight.js
+    syntax_highlighting = False
```
davidselassie, line 25: It looks like [pdoc3 does its own conversion of the doctest code blocks to markdown code blocks](https://github.com/pdoc3/pdoc/blob/2cce30a9b55eeeddc1ed826c8a2ada53777c3eea/pdoc/html_helpers.py#L367) and is the thing adding `python-repl` as the language.

If you want, I could investigate changing all the docstrings to have these code blocks be explicit. Like:

````
```python
>>> x = 5
>>> x
5
```
````

But I'd need to investigate if this breaks our doctest test running equipment.

I think we should move forward with getting some docs out there and fix this up later, though.

davidselassie (approved):
Looks good. I'm happy to review any other template / built API doc changes.

apidocs/templates/config.mako
konradsienkowski: Sure, I'd merge that for now. We can come back to that issue later. Thanks for the review! 🙌

PR #42, opened 2022-03-24
Feast example

blakestier:
As a first pass, this example manufactures a sort of Kafka stream (more in a batch) of driver data, partitions in in 24 hour epoch increments, averages driver `conv_rate` and `acc_rate` over each day and saves it to a feast _online_ store.

The data is batched into the stream and then timestamped by the event timestamps during intake.

To run locally:

Bring up docker container with `docker-compose up`
From `feast` directory, run `feast apply` to establish feature store/views
Run `python stream_data.py` to populate kafka stream

if you run 
```
docker exec --interactive --tty broker kafka-console-consumer --bootstrap-server broker:9092 --topic drivers --from-beginning
```

you should see the "streaming" data in Kafka

Run `python feast_example.py` to see data store in the online store.


examples/feast/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+    environment:
+      KAFKA_BROKER_ID: 1
+      KAFKA_ZOOKEEPER_CONNECT: 'zookeeper:2181'
+      KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: PLAINTEXT:PLAINTEXT,PLAINTEXT_INTERNAL:PLAINTEXT
+      KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://localhost:9092,PLAINTEXT_INTERNAL://broker:29092
+      KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1
+      KAFKA_TRANSACTION_STATE_LOG_MIN_ISR: 1
+      KAFKA_TRANSACTION_STATE_LOG_REPLICATION_FACTOR: 1
+
+  redis:
```
blakestier, line 29: We didn't end up needing this, yet. I can remove it.

examples/feast/feast_dataflow.py
```diff
@@ -0,0 +1,106 @@
...
+
+from feast import FeatureStore
+
+# Configure the feature store for each worker to access
+store = FeatureStore(repo_path=".")
+
+
+def input_builder(worker_index, worker_count):
+    def consume_from_kafka():
+        consumer = KafkaConsumer(bootstrap_servers="localhost:9092", auto_offset_reset=OffsetType.EARLIEST)
```
blakestier, line 17: Consumes data from the beginning of the stream. We'll change this if we want to make things streamier

whoahbot: I saw in your demo that you've updated this call here to not use the `OffsetType` class. When I run this locally, the version I installed from `requirements.txt` doesn't have this class.

blakestier: Oh yes, I fixed that locally just before the demo. Well, changed it back to a string.

whoahbot: Can you add that change to this PR? I wasn't able to run the example without that modification.

examples/feast/feast_dataflow.py
```diff
@@ -0,0 +1,110 @@
...
+        print("New values for driver:", feature_views)
+
+    return output_fn
+
+def collect_events(all_events, event):
+    all_events.extend(event)
+    return all_events
+
+def calculate_avg(driver_events):
+    driver_events = driver_events[1]
```
davidselassie, line 71: I'm having a little trouble keeping track of what is supposed to be in which variable. I think replacing this with `driver_id, events = driver_events` and using those names in the rest of the function would help a little.

examples/feast/feast_dataflow.py
```diff
@@ -0,0 +1,110 @@
...
+    total_events = len(driver_events)
+    e = driver_events[0]
```
davidselassie, line 73: `first_event`?

examples/feast/feast_dataflow.py
```diff
@@ -0,0 +1,110 @@
...
+            "driver_id": [1001],
+            "event_timestamp": [datetime.now()],
+            "created": [datetime.now()],
+            "conv_rate": [1.0],
+            "acc_rate": [1.0],
+        }
+    """
+
+    def output_fn(epoch_dataframe):
+        panda_df = epoch_dataframe[1]
```
davidselassie, line 50: I love tuple unpacking vs using indices so names are apparent. `epoch, avgs_df = epoch_df`?

whoahbot (commented):
Looks good. I pulled down the latest changes and I was able to run it locally.

A few suggested remaining items:

- [ ] Remove redis from the docker-compose.yml file
- [ ] Add a README for this example

examples/feast/feast_dataflow.py
```diff
@@ -0,0 +1,110 @@
...
+                "driver_daily_stats:conv_rate",
+                "driver_daily_stats:acc_rate",
+            ],    
+            entity_rows=[{"driver_id": driver_id}]
+        ).to_dict()
+        print("New values for driver:", feature_views)
+
+    return output_fn
+
+def collect_events(all_events, event):
```
davidselassie, line 66: `event` is actually a list at this point, so `new_events`?

examples/feast/feast_dataflow.py
```diff
@@ -0,0 +1,110 @@
...
+
+    return event_df
+
+def add_driver_id(record):
+    # Record needs to be an array for `collect_events`
+    return record["driver_id"], [record]
+
+
+if __name__ == "__main__":
+    flow = Dataflow()
```
davidselassie, line 98: One thing that might help with understand-ability is using the same names for the same data between logic functions. Our dataflow model makes things tough since one has to bounce back and forth between logic functions and the dataflow itself to see what gets threaded where.

Perhaps reconciling some of those names would make this even more clear:

- `record` vs `event` vs the anonymous result of `json.loads(msg.value)`
- `event_df` vs `panda_df`

davidselassie (approved):
Looks good to me, although I assume you and Dan are trying to work out Feast details still. I have some comments on naming things that might help with understanding the demo, but if you don't want to deal with them as this is a prototype, that's fine too.

examples/feast/feast_dataflow.py
blakestier: yes, great suggestion

examples/feast/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+The data is batched into the stream and then timestamped by the original event timestamps during intake.
+
+To run locally:
+
+From `feast` directory,
+
+Bring up docker container with `docker-compose up`.
+Run `feast apply` to establish feature store/views.
+Run `python stream_data.py` to populate kafka stream.
```
whoahbot, line 11: The `stream_data.py` file is in the `utils` directory, might want to change the path here as well.

examples/feast/README.md
```diff
@@ -0,0 +1,20 @@
+This example manufactures a Kafka pseudo-stream (bulk loaded in a batch) of driver data, partitions it into 24 hour epoch increments, averages driver `conv_rate` and `acc_rate` over each day and saves it to a Feast _online_ store using their alpha `write_to_online_store` operation.
+
+The data is batched into the stream and then timestamped by the original event timestamps during intake.
+
+To run locally:
+
+From `feast` directory,
+
+Bring up docker container with `docker-compose up`.
+Run `feast apply` to establish feature store/views.
```
whoahbot, line 10: You'll need to `pip install -r requirements.txt` to make this `feast apply` work.

whoahbot (approved):

PR #44, opened 2022-03-28
Prototype async work

davidselassie:
Here's a PR for posterity with some followup work on https://github.com/bytewax/bytewax/pull/31 . I'm going to close this immediately but wanted to save the code within for perhaps being used in the future.

The big things within here were converting `run_cluster` to use asyncio and attempt to run the input processing, cluster managment, and output processing in coroutines and not our home-spun busy loop. Also remove busy looping in `spawn_cluster`.

There were some problems that I don't think are worth further effort today to surmount, so this is being thrown out. Comments about them below.

One thing we might re-use is the `async for` output builder since it allows code to run on worker output completion.


pysrc/bytewax/execution.py
```diff
@@ -39,68 +146,208 @@ def run(flow: Dataflow, inp: Iterable[Tuple[int, Any]]) -> List[Tuple[int, Any]]
...
+    )
+
+    worker_raised_ex = threading.Event()
+
+    def worker_should_stop():
+        return proc_should_stop() or worker_raised_ex.is_set()
+
+    def worker_main(builder):
+        coro = builder.build(flow, input_builder, output_builder, worker_should_stop)
+        asyncio.run(WorkerCoroWrapper(coro))
```
davidselassie, line 258: We were having trouble with a Rust `SendWrapper` type being accessed from a different thread. I __think__ this is the Python GC running on the main thread cleaning up the `WorkerCoro` but I'm not sure.

pysrc/bytewax/execution.py
```diff
@@ -118,58 +365,71 @@ def spawn_cluster(
...
-                    input_builder,
-                    output_builder,
+
+    manager = Manager()
+    proc_raised_ex = manager.Event()
+
+    def proc_should_stop():
+        return cluster_should_stop() or proc_raised_ex.is_set()
+
+    ctx = multiprocessing.get_context("spawn")
```
davidselassie, line 380: From everything I read, fork is tricky and can cause lots of mysterious errors. https://bugs.python.org/issue40379 I am not sure but using spawn might be necessary in complex situations like this.

pysrc/bytewax/execution.py
```diff
@@ -189,26 +459,199 @@ def run_cluster(
...
+    main_raised_ex = manager.Event()
+    
+    in_q = manager.Queue(ipc_buffer_size)
+
+    def input_task():
+        logging.debug("Input task started")
+        for epoch_item in inp:
+            msg = ("yield", epoch_item)
+            msg_bytes = dill.dumps(msg)
+            in_q.put(msg_bytes)
```
davidselassie, line 485: What would make this code much simpler, is an actual async IPC queue. Then we wouldn't need to spawn task threads and jump through hoops to unblock queues to cancel them.

pytests/test_execution.py
```diff
@@ -1,87 +1,245 @@
...
+
+    async def output_builder(worker_index, worker_count, epoch_items):
+        async for epoch_item in epoch_items:
+            out.append(epoch_item)
+
+    with raises(BoomError):
+        asyncio.run(run_main(flow, input_builder, output_builder))
 
 
+def test_run_main_output_reraises_exception():
```
davidselassie, line 106: I do want to steal these more advanced exception tests.

PR #47, opened 2022-03-31
`stateful_map` should pass a tuple of (key, value) to the builder

whoahbot:
When working on an example, I found it useful to have access to the full value that will be passed to the mapper function to build the initial value for `stateful_map`.

In the example I was working on, I was joining two streams of data. The first time I got a new value for a key, I wanted to use it as the initial value.

This PR changes the API of `stateful_map` to pass the `value` instead of the `key`. One could argue that you should pass the builder function both, but `value` in this case is always a tuple of (key, value). I considered destructuring it in rust to make the API for the `builder` function take two params of `key` and `value`, but left it this way for the moment.


blakestier (commented):
Makes sense to me. Will this break examples?

PR #51, opened 2022-04-06
Integrate docs with main repository

konradsienkowski:
Move docs content to the main repository & add metadata .json file for consuming them in our Vue app.


davidselassie (approved):
Great! Thanks!

davidselassie:
Oh, I totally forgot that all the tests will break until we "fix" the docs to pass the doc tester. If you remove `"docs",` from `testpaths` at the end of `pyproject.toml`, that'll do it. I can fix them up later.

davidselassie (approved):

davidselassie:
I cherry-picked this into the `v0.8.0` tag so we could test new docs via the GitHub API releases mechanism. See dc4c846a0a964ffe31a98ae848164c6a5dc2658e.

Usually we won't need to do this as the docs will already have the base format. But if we do need to do a retroactive sweeping change to all the docs (since each version is managed independently) this could be one way to do that.

PR #52, opened 2022-04-06
Dockerfile enhancements

miccioest:
This PR includes these changes:

- Setting a proper Entrypoint in Dockerfiles to allow running Bytewax docker image locally easily.
- `Dockerfile.release` has been adapted to receive a python version as an argument.
- GitHub Actions release workflow now creates a docker image for each python version.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -105,6 +105,10 @@ jobs:
     runs-on: ubuntu-latest
     if: "startsWith(github.ref, 'refs/tags/')"
     needs: [ macos, windows, linux ]
+    outputs:
+      bytewax-version: ${{ steps.get-info.outputs.bytewax-version }}
+      version-image-tag: ${{ steps.get-info.outputs.version-image-tag }}
+      latest-image-tag: ${{ steps.get-info.outputs.latest-image-tag }}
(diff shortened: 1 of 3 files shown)


.github/workflows/CI.yml
```diff
@@ -119,40 +123,34 @@ jobs:
         run: |
           pip install --upgrade twine
           twine upload --skip-existing *
+      - name: Get Version and Tags
+        id: get-info
+        run: |
+          BYTEWAX_VERSION=${GITHUB_REF:11}
```
whoahbot, line 129: I'm unfamiliar with how this syntax works. If we release beta versions (0.8.1-beta.0), will this account for that structure?

whoahbot (approved):
Looks good! Left you one question.

.github/workflows/CI.yml
miccioest: Yes, it takes the whole string after the `v` in our tags. So for the example that you say is going to take `0.8.1-beta-0` as we expect.

We also have a successful workflow using that expression when we released `0.8.0-beta-2` some weeks ago: https://github.com/bytewax/bytewax/runs/5333608042?check_suite_focus=true

PR #53, opened 2022-04-07
State Recovery

davidselassie:
Adds in first pass of state recovery machinery. See commit messages.


pysrc/bytewax/execution.py
```diff
@@ -129,10 +154,13 @@ def spawn_cluster(
                     flow,
                     input_builder,
                     output_builder,
-                    addresses,
-                    proc_id,
-                    worker_count_per_proc,
                 ),
+                {
```
blakestier, line 158: Having these bundled in a config feels right to me

src/recovery/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+
+    // TODO: We can implement state compaction / GC once we know
+    // epochs are output by every capture() operator.
+    //fn compact(&self, done_epoch: u64);
+}
+
+/// Per-operator interface for state recovery.
+///
+/// Each [`RecoveryStore`] will need to implement this with the
+/// specific queryies or inserts needed.
```
whoahbot, line 116:
```suggestion
/// specific queries or inserts needed.
```

docs/examples/cart-join.md
```diff
@@ -0,0 +1,256 @@
...
+
+We need a **recovery store**. For this example we'll use a local
+[Sqlite]() database because we're running on a single machine.
+
+
+```python
+from tempfile import TemporaryDirectory
+from bytewax.recovery import SqliteRecoveryConfig
+
+recovery_dir = TemporaryDirectory()  # We'll store this somewhere temporary for this test.
```
whoahbot, line 194: Just a note that if you run this as a script, the temporary directory gets GCd at the end of every execution, which gives you an unexpected error when running again with a recovery epoch:

```
thread '<unnamed>' panicked at 'Box<dyn Any>', src/operators/mod.rs:438:71
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
  File "/home/whoahbot/code/bytewax/bytewax/./examples/cart_join.py", line 77, in <module>
    run_main(flow, input_builder, output_builder, recovery_config=recovery_config)
  File "/home/whoahbot/code/bytewax/bytewax/./examples/cart_join.py", line 46, in joiner
    state["unpaid_order_ids"].remove(order_id)
ValueError: list.remove(x): x not in list
```

src/execution/mod.rs
```diff
@@ -161,7 +168,15 @@ pub(crate) fn run_main(
             timely::execute::execute_directly::<Result<(), String>, _>(move |worker| {
                 let (pump, probe) = Python::with_gil(|py| {
                     let flow = &flow.as_ref(py).borrow();
-                    build_dataflow(worker, py, flow, input_builder, output_builder)
+
+                    build_dataflow(
```
whoahbot, line 172: Is there a way for us to throw an error if you try and build a dataflow with the SQLite recovery backend and multiple processes?

whoahbot (approved):
Really happy with the approach so far. I left you a few questions that aren't blockers for this.

pysrc/bytewax/execution.py
```diff
@@ -189,7 +217,13 @@ def run_cluster(
         flow: Dataflow to run.
 
         inp: Input data. Will be reifyied to a list before sending to
```
blakestier, line 219: while you're here? `reifyied` -> `reified`

src/dataflow/mod.rs
```diff
@@ -198,9 +206,13 @@ impl Dataflow {
     ///
     /// It emits a `(key, updated_value)` tuple downstream for each
     /// input item.
-    #[pyo3(text_signature = "(self, builder, mapper)")]
-    fn stateful_map(&mut self, builder: TdPyCallable, mapper: TdPyCallable) {
-        self.steps.push(Step::StatefulMap { builder, mapper });
+    #[pyo3(text_signature = "(self, step_id, builder, mapper)")]
+    fn stateful_map(&mut self, step_id: String, builder: TdPyCallable, mapper: TdPyCallable) {
```
blakestier, line 210: the `step_id` is a name here, yes? Would it be clearer to call it `step_name` or `task_name`?

src/execution/mod.rs
blakestier: The docs say it will just be slow. If it still works we could allow multiple processes locally (however slow) but a warning does seem prudent at minimum. If we're going to throw an error we should be more finger-waggy in the docs

src/execution/mod.rs
```diff
@@ -218,22 +233,35 @@ pub(crate) fn run_main(
...
 ///     worker_count_per_proc: Number of worker threads to start on
 ///         each process.
-#[pyfunction]
+#[pyfunction(
+    flow,
+    input_builder,
+    output_builder,
+    addresses,
+    proc_id,
+    "*",
```
blakestier, line 259: what's this little asterisk bud telling me?

src/recovery/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+/// We'll implement this for each kind of recovery store we need to
+/// talk to. This is not exposed Python-side.
+pub(crate) trait RecoveryStore {
+    // TODO: I wanted to make a trait RecoveryStore statically typed, but I
+    // think that'll require GAT / generic associated types.
+
+    /// Build recovery interface for this step.
+    fn for_step(&self, step_id: &str) -> Box<dyn StepRecovery<u64, TdPyAny, TdPyAny>>;
+
+    // TODO: We can implement state compaction / GC once we know
```
blakestier, line 108: 👍

docs/articles/getting-started/recovery.md
```diff
@@ -0,0 +1,66 @@
...
+## Performing Recovery
+
+If a dataflow has failed, follow this high-level game plan to recover
+and have it resume processing from where it failed:
+
+1. Terminate the dataflow processes.
+
+2. Observe the previous dataflow output and note the final epoch
+   emitted from each process. Find the oldest / smallest epoch. The
+   epoch just _before_ this will be the recovery epoch.
```
blakestier, line 57: This is helpful

docs/examples/cart-join.md
```diff
@@ -0,0 +1,256 @@
+In this example, we're going to build a small online order fulfillment
+system. It will join two event streams: one containing customer orders
+and another containing successful payments, and then emit which orders
+for each customer have been paid. We're also going to design it to be
+**recoverable** and restartable after invalid data.
+
+## Sample Data
+
+First let's make a local file containig our sample event data in JSON
```
blakestier, line 9:
```suggestion
First let's make a local file containing our sample event data in JSON
```

docs/examples/cart-join.md
```diff
@@ -0,0 +1,256 @@
...
+format. Save this as `cart-join.jsons`.
```
blakestier, line 10: I usually see this as `json`, not `jsons`. Does `jsons` work?

docs/examples/cart-join.md
```diff
@@ -0,0 +1,256 @@
...
+    assert worker_index == 0  # We're not going to worry about multiple workers yet.
+    with open("cart-join.jsons") as f:
+        for epoch, line in enumerate(f):
+            obj = json.loads(line)
+            yield epoch, obj
+```
+
+## Dataflow
+
+Let's get our empty dataflow setup.
```
blakestier, line 43:
```suggestion
Let's get our empty dataflow set up.
```

docs/examples/cart-join.md
```diff
@@ -0,0 +1,256 @@
...
+```
+
+Now, our plan is to use the stateful map operator to actually do the
+join. All stateful operators require their input data to be a `(key,
+value)` tuple so that the system can route all keys so they access the
+same state.  Let's add that key field using the `user_id` field
+present in every event.
+
+```python
+def key_off_user_id(event):
```
blakestier, line 58: I like this name for other examples 🤔

src/recovery/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+            Python::with_gil(|py| key.extract(py)).expect("Key cannot be cast to string");
+        let state_pickled: Option<Vec<u8>> = state.as_ref().map(|state| {
+            Python::with_gil(|py| -> Result<Vec<u8>, PyErr> {
+                let pickle = py.import("dill")?;
+                Ok(pickle.call_method1("dumps", (state,))?.extract()?)
+            })
+            .expect("Error pickling state")
+        });
+
+        let future = query("INSERT INTO states (epoch, step_id, key, state) VALUES (?1, ?2, ?3, ?4) ON CONFLICT (epoch, step_id, key) DO UPDATE SET state = EXCLUDED.state")
```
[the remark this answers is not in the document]
davidselassie: I'm not totally clear on the context of your question. The high-level purpose of the `ON CONFLICT` clause is so that if a second dataflow execution re-calculates the state for some key at some epoch (which very easily could happen if we're restarting and one worker is ahead in an epoch than another; we have to go back to the oldest finalized epoch), the re-calculation is saved _overwriting_ the old value. I think overwriting is less surprising than ignoring since the resulting state on the second pass; it feels more correct that the results of the most recent input data should be used.

You might argue that we should hard error on this situation (or have a toggle to), but then we'd have to have a surefire way to know which recovery data to delete to allow a graceful restart (since some workers might be ahead of others). I'd guess that this would mean "delete everything after the latest finalized / sealed epoch" (once we implement some way of calculating that), but I'm not totally sure that's correct in the general case (especially with delays). This seems easier for now.

docs/examples/cart-join.md
```diff
@@ -0,0 +1,256 @@
...
+        for epoch, line in enumerate(f):
+            if epoch < RESUME_FROM_EPOCH:
+                continue
+            obj = json.loads(line)
+            yield epoch, obj
+```
+
+
+We need a **recovery store**. For this example we'll use a local
+[Sqlite]() database because we're running on a single machine.
```
blakestier, line 187: This needs a link still

src/execution/mod.rs
davidselassie: We could check the number of `addresses` (since each process gets one) and error out. It feels a little strange to prevent something that is correct but not performant. Where's your bar for "not performant"? Especially since any recovery store will have its performance trade offs and feels tough to encode all that into detectors and warnings.

docs/examples/cart-join.md
```diff
@@ -0,0 +1,256 @@
...
+
+```python
+from tempfile import TemporaryDirectory
+from bytewax.recovery import SqliteRecoveryConfig
+
+recovery_dir = TemporaryDirectory()  # We'll store this somewhere temporary for this test.
+recovery_config = SqliteRecoveryConfig(recovery_dir.name + "/state-recovery.sqlite3")
+```
+
+Now if we run the dataflow, every epoch the internal state will be
```
blakestier, line 198:
```suggestion
Now if we run the dataflow, every epoch in the internal state will be
```
?

src/dataflow/mod.rs
davidselassie: Bytewax doesn't care what you call it, just that it's unique per "version of the dataflow." (Which is something we'll need to suss out later, how we evolve this state over time.) To me, `id` implies uniqueness more than `name`.

davidselassie: Also we use in the docs `step` as a formal word to mean "a specific use of an operator with your custom logic".

src/execution/mod.rs
davidselassie: This is saying the remainder of the arguments are "keyword only" arguments which require the call site to name each argument. Thus it would be a... `TypeError` (I think) to call `cluster_main(flow, ib, ob, address, proc_id, recovery_store)`. You have to pass all arguments after the `*` by keyword `cluster_main(flow, ib, ob, address, proc_id, recovery_store=recovery_store)`.

The general idea is everything after the `*` is optional and non-hierarchical (e.g. specifying `worker_count_per_proc` doesn't imply that you should be setting `recovery-store`), so you need to pass it by keyword to make it so you don't have to repeat defaults for things you don't care about.

I couldn't find a concise summary of the different parameter passing "modes" but https://peps.python.org/pep-0570/ and https://peps.python.org/pep-3102/ discuss the details of `*` and `/`. Relatedly there's arg and kwarg unpacking with `*list_of_args`, and `**dict_of_kwargs`.

docs/articles/getting-started/recovery.md
```diff
@@ -0,0 +1,66 @@
...
+   (e.g. `cluster_main()`) you are using.
+   
+## Performing Recovery
+
+If a dataflow has failed, follow this high-level game plan to recover
+and have it resume processing from where it failed:
+
+1. Terminate the dataflow processes.
+
+2. Observe the previous dataflow output and note the final epoch
```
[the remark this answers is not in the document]
davidselassie: We will throw an error if you do not include a `capture()` step when building your dataflow. But, yes, the user could ignore it and never log it our output it and that would make recovery tough. I'll add a comment to log it somewhere. Figuring out the recovery epoch is important enough that we are going to build some machinery to calculate it automatically (also it allows us to garbage collect old recovery state) so I was thinking this would be somewhat of a temporary manual step.

pysrc/bytewax/execution.py
```diff
@@ -30,7 +37,12 @@ def run(flow: Dataflow, inp: Iterable[Tuple[int, Any]]) -> List[Tuple[int, Any]]
 
         flow: Dataflow to run.
 
-        inp: Input data.
+        inp: Input data. If you are recovering a stateful dataflow,
+            your input should resume from the last finalized epoch.
```
blakestier, line 41:
```suggestion
            your input should resume from the last finalized epoch specified.
```

docs/examples/cart-join.md
davidselassie: Hmmm. Perhaps we should provide a more formal way of saying "I know my recovery state is empty; create it anyway" and error out otherwise? This is a confusing manifestation of that.

src/recovery/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+///
+/// Returns:
+///
+///     Config object. Pass this as the `recovery_config` argument to
+///     your execution entry point.
+#[pyclass(module="bytewax.recovery", extends=RecoveryConfig)]
+#[pyo3(text_signature = "(db_file_name)")]
+pub(crate) struct SqliteRecoveryConfig {
+    #[pyo3(get)]
+    pub(crate) db_file_name: String,
```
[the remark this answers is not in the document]
davidselassie: Yes, this is a relative path. Will change.

docs/articles/getting-started/recovery.md
```diff
@@ -0,0 +1,66 @@
...
+   emitted from each process. Find the oldest / smallest epoch. The
+   epoch just _before_ this will be the recovery epoch.
+
+3. Restart your dataflow, but pass in this recovery epoch as the point
```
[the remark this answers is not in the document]
davidselassie: I bounced back and forth between threading a `resume_epoch` argument through Bytewax itself (as was in the first prototype) and being hands off about it. I think it comes down to whether we're positioning this as a library for "big infrastructure things" or "prototyping". If you just want to throw up a quick dataflow, it's annoying to have to interpret and use a `resume_epoch` argument in each input builder, or you need to manually add in assertions that "we can't seek" in this input stream. And you could forget to do those, does the library double check? Although if we hand off that responsibility to you, the input builder could still do the wrong thing just defaulting the other direction. It's also not totally clear to me how much coordination is necessary with the library to support recovery.

I think we should see what evolves in this area once we start handling automatic calculation of the finalized epoch.

docs/examples/cart-join.md
blakestier: I guess it must 🤯

src/execution/mod.rs
blakestier: Mostly I figure if we say "only do this locally or it's kinda wrong" then we should also log a warning, but we haven't really formalized our log criteria so 🤷‍♀️

src/execution/mod.rs
blakestier: Thank you. I googled "asterisk python arguments" and fainted at the number of results

docs/examples/cart-join.md
davidselassie: It does work, it's just a file name. I do this to remember that the file isn't a single JSON object, but is an object on each line. But I don't think that's a common convention. It's raising more questions than answers so I'll change this back.

docs/articles/getting-started/recovery.md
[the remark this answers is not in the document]
davidselassie: I was also kind of having these feelings surrounding "should Bytewax code have options to change logging verbosity". Is this a library which manages the full execution for you more framework-style? Or are you setting up the execution environment and you delegate to us when things are right? It feels wrong to have `bytewax.run()` call `logging.setLevel(DEBUG)` or whatever, but it also seems useful to provide some scaffolding surrounding it.

docs/examples/cart-join.md
davidselassie: I mean it is persisted each epoch. Will clarify.

pysrc/bytewax/execution.py
davidselassie: What is the `specified` getting at in this case? I imagine "last finalized epoch" is a fact about the dataflow, so only one exists at any given moment. And it's not part of the arguments here or something you need to coordinate between arguments.

pysrc/bytewax/recovery.py
```diff
@@ -0,0 +1,32 @@
...
+0. Pick a **recovery store** in which internal state will be
+   continously backed up. Different storage systems will have
+   different performance and reconfiguration tradeoffs.
+
+1. Create a **recovery config** for connecting to your recovery store.
+
+2. Pass that created config as the `recovery_config` argument to the
+   entry point running your dataflow (e.g. `bytewax.run_cluster()`).
+
+3. Run your dataflow from the beginning, or resuming input from the
```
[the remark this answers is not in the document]
davidselassie: A little bit of both!

docs/articles/getting-started/recovery.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Prerequisites
+
+Not all dataflows are possible to be recovered. There are a few
+requirements:
+
+1. _Stateful Operators_ - Only the operators
+   [reduce](./operators#reduce) and [stateful
+   map](./operators#stateful_map) aggregate state _between_ epochs. If
+   your dataflow is simple enough to not be using these,
```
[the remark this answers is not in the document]
davidselassie: I guess this point actually isn't true. You can recover a dataflow that isn't using any recoverable operators, just the execution won't result in any data being in the recovery store and so recovery is "trivial" in that you can just replay input items without setting up state again. Since Bytewax does all the right things for you, I think I'll just drop this point.

src/operators/mod.rs
```diff
@@ -55,51 +62,164 @@ pub(crate) fn inspect_epoch(inspector: &TdPyCallable, epoch: &u64, item: &TdPyAn
...
+                                    // If there's previous aggregator,
+                                    // reduce.
+                                    reducer(&key, aggregator_for_key, value)
+                                } else {
+                                    // If there's no previous aggregator,
+                                    // pass through the current value.
+                                    value
+                                };
+
+                            // Note which keys' state were modified to
```
blakestier, line 159:
```suggestion
                            // Note which keys' states were modified to
```

🤔

src/recovery/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+}
+
+struct NoOpStepRecovery {
+    step_id: String,
+}
+
+impl<T: Debug, K: Debug, D: Debug> StepRecovery<T, K, D> for NoOpStepRecovery {
+    fn recover_last(&self, current_epoch: &T, key: &K) -> Option<D> {
+        debug!(
+            "noop state fake-recovered step_id={} key={key:?}@epoch={current_epoch:?}",
```
blakestier, line 149: When would we want to see this? If we did like a dry run? I got stuck on the "fake-recovered" bit

src/recovery/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+                .busy_timeout(Duration::from_secs(5))
+                .journal_mode(sqlx::sqlite::SqliteJournalMode::Wal)
+                .locking_mode(sqlx::sqlite::SqliteLockingMode::Normal)
+                .synchronous(sqlx::sqlite::SqliteSynchronous::Normal)
+                .thread_name(|i| format!("sqlx-sqlite-{i}"));
+            let future = SqlitePoolOptions::new().connect_with(options);
+            rt.block_on(future)
+        })
+        .unwrap();
+        debug!("Opened Sqlite connection pool to {db_file_name}");
```
blakestier, line 189: I was tempted to raise a security concern here but probably a bit premature

docs/examples/cart-join.md
blakestier: makes sense

src/dataflow/mod.rs
blakestier: ID for a unique value makes sense.

Is uniqueness guaranteed, then? Since we're creating new tables I'm not sure if the primary key will bark.

pysrc/bytewax/execution.py
davidselassie: We reworded this to imply you have to make your input resume from the last finalized epoch.

blakestier (approved):
😅  Thank you so much for the big lift, the docs, the tests, and the commits. 🎉

src/recovery/mod.rs
davidselassie: What specific future concerns do you have?

blakestier: i'm a nervous nelly and get 🙈  about logging any db info. It's not like it's a password tho

src/recovery/mod.rs
davidselassie: It's "fake" in that the no-op store will never be able to recover state. It always returns "not found". I wanted to log that a query to the recovery store occurred, though. This is mostly useful for debugging Bytewax and state recovery itself. Let me use some clearer words here. Eventually we should namespace our logging so someone can tune the spew.

This touches on a good point in that I'm sort of overloading the concept of recovery here. There's nothing stopping one from "recovering" a dataflow execution using this no-op store and there being no errors, but you'd get different output because all of the state was reset to "new key". Although no one is stopping you from re-running your dataflow over different data or starting at a different point in the input and seeing different output. I guess it's important to realize that output is a function of:

1. Input values
2. Input order
3. Recovery store state

All at once. You can't really detect that one is "wrong".

src/recovery/mod.rs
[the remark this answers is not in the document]
whoahbot: Thanks, that makes sense. I think the context of my original question was a misunderstanding of the sqlite behavior of ON CONFLICT.

src/execution/mod.rs
whoahbot: If it's correct but not performant, I'm fine with not throwing an error here.

PR #58, opened 2022-04-18
Re-enable documentation tests and upgrade longform docs to 0.8.0

davidselassie:
Updates our docs to 0.8.0 API and gets all the documentation tests working again. I had to jump through some test runner and multiprocessing hoops to make that happen. Then something about changing the multiprocessing context caused the other execution tests to break, so had to fix them there too. Unfortunately, even though I fixed up the _docstring_ doctests and they pass locally, CI can't run them because the hacks to enable them in an installed package broke multiprocessing elsewhere; we should still run them locally during development with `pytest --doctest-modules`. See individual commits for compartmentalized changes and detailed descriptions.

### DIFF
--- a/docs/articles/getting-started/epochs.md
+++ b/docs/articles/getting-started/epochs.md
@@ -1,91 +1,162 @@
-Now that we've seen an end-to-end example running a dataflow program, it's time to discuss **epochs**.
+Now that we've seen an end-to-end example running a dataflow program, it's time to discuss epochs.
 
 To explain what an epoch represents, let's imagine an unending stream of lines that we want to count words for. If the stream has no end, we'll need to introduce another concept to decide when we want to emit the current counts of words from the stream.
 
-Epochs are what allow us to influence ordering and batching within a dataflow. Increment the epoch when you know you need some data to be processed strictly after, or separately from the proceeding data.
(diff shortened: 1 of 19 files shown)


pysrc/bytewax/testing.py
```diff
@@ -0,0 +1,52 @@
...
+
+doctest_ctx = multiprocessing.dummy
+"""Use this `multiprocessing` context when running in doctests.
+
+Pass to `bytewax.spawn_cluster()` and `bytewax.run_cluster()`.
+
+Spawning subprocesses is fraught in doctest contexts, so use this to
+demonstrate the API works, but not actually run via multiple
+processes. We have other normal `pytest` tests which actually test
+behavior. Don't worry.
```
blakestier, line 37: 😂

whoahbot (approved):

blakestier (approved):

docs/articles/getting-started/epochs.md
```diff
@@ -1,91 +1,162 @@
...
+('and', 1)
+('by', 1)
 ```
 
 Notice how there are multiple instances of counts of `to` and `and`, etc!
 
 ## Inspecting epochs
 
-To better illustrate that counts are per-line, we can use the [inspect epoch](/operators/operators#inspect-epoch) operator instead of [inspect](/operators/operators#inspect):
+To better illustrate that counts are per-line, we can print the epoch in addition to the output items when we run the dataflow:
```
blakestier, line 150: 🎉

PR #59, opened 2022-04-19
Frontier psychiatry

whoahbot:
## Overview

This PR adds two new primitives to bytewax: `AdvanceTo` and ~`Send`~ `Emit`. These two primitives expose two more fundamental constructs of Timely Dataflow to our users, and allow you to influence when work is performed in your dataflow.

## Rationale

In some of our examples that use Kafka, I discovered that outputs for a particular epoch would not be generated until new input was introduced into the dataflow. This coupling of inputs to outputs allows us to have a very fluent Python interface, but it led to some confusion when expecting the results of your latest input.

## Changes

All of the execution parts of the code, save `cluster_main` and `spawn_cluser` now utilize `AdvanceTo` and `Send` with their input builder. This allows us to keep almost the same interface for executors that use generators.

### DIFF
--- a/pysrc/bytewax/inputs.py
+++ b/pysrc/bytewax/inputs.py
@@ -15,7 +15,7 @@ from typing import Any, Callable, Iterable, Tuple
 
 def yield_epochs(fn: Callable):
     """A decorator function to unwrap an iterator of [epoch, item]
-    into successive `AdvanceTo` and `Send` classes with the
+    into successive `AdvanceTo` and `Emit` classes with the
     contents of the iterator.
 
     Use this when you have an input_builder function that returns
--- a/pysrc/bytewax/inputs.py
+++ b/pysrc/bytewax/inputs.py
@@ -13,6 +13,37 @@ from dataclasses import dataclass
 from typing import Any, Callable, Iterable, Tuple
 
 
+def yield_epochs(fn: Callable):
+    """A decorator function to unwrap an iterator of [epoch, item]
+    into successive `AdvanceTo` and `Send` classes with the
+    contents of the iterator.
+
+    Use this when you have an input_builder function that returns
+    a generator of (epoch, item) to be used with `cluster_main` or
+    `spawn_cluster`:
+
+    >>> from bytewax import Dataflow, cluster_main
+    >>> from bytewax.inputs import yield_epochs, fully_ordered
+    >>> flow = Dataflow()
+    >>> flow.capture()
+    >>> @yield_epochs
+    ... def input_builder(i, n):
+    ...   return fully_ordered(["a", "b", "c"])
+    >>> cluster_main(flow, input_builder, lambda i, n: print, [], 0, 1)
+    (0, 'a')
+    (1, 'b')
+    (2, 'c')
+    """
+
+    def inner_fn(worker_index, worker_count):
+        gen = fn(worker_index, worker_count)
+        for (epoch, item) in gen:
+            yield AdvanceTo(epoch)
+            yield Send(item)
+
+    return inner_fn
+
+
 def single_batch(wrap_iter: Iterable) -> Iterable[Tuple[int, Any]]:
     """All input items are part of the same epoch.
 
--- a/pysrc/bytewax/inputs.py
+++ b/pysrc/bytewax/inputs.py
@@ -3,8 +3,12 @@
 Use these to wrap an existing iterator which yields items.
 
 """
+
 import datetime
 import heapq
+
+from .bytewax import AdvanceTo, Send
+
 from dataclasses import dataclass
 from typing import Any, Callable, Iterable, Tuple
 
--- a/pysrc/bytewax/inputs.py
+++ b/pysrc/bytewax/inputs.py
@@ -7,7 +7,7 @@ Use these to wrap an existing iterator which yields items.
 import datetime
 import heapq
 
-from .bytewax import AdvanceTo, Send
+from .bytewax import AdvanceTo, Emit
 
 from dataclasses import dataclass
 from typing import Any, Callable, Iterable, Tuple
@@ -39,7 +39,7 @@ def yield_epochs(fn: Callable):
         gen = fn(worker_index, worker_count)
         for (epoch, item) in gen:
             yield AdvanceTo(epoch)
-            yield Send(item)
+            yield Emit(item)
 
     return inner_fn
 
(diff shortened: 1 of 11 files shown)


src/execution/mod.rs
```diff
@@ -42,10 +75,19 @@ impl Pump {
     fn pump(&mut self) {
         Python::with_gil(|py| {
             let mut pull_from_pyiter = self.pull_from_pyiter.0.as_ref(py);
-            if let Some(epoch_item_pytuple) = pull_from_pyiter.next() {
-                let (epoch, item) = with_traceback!(py, epoch_item_pytuple?.extract());
-                self.push_to_timely.advance_to(epoch);
-                self.push_to_timely.send(item);
+            if let Some(epoch_item) = pull_from_pyiter.next() {
```
davidselassie, line 78: `epoch_item` -> ....... `input_action`? What have you been calling `Send` and `AdvanceTo` as a group?

src/execution/mod.rs
```diff
@@ -42,10 +75,19 @@ impl Pump {
...
+                match epoch_item {
+                    Ok(item) => {
+                        if let Ok(send) = item.downcast::<PyCell<TdSend>>() {
+                            self.push_to_timely.send(send.borrow().item.clone());
+                        } else if let Ok(advance_to) = item.downcast::<PyCell<TdAdvanceTo>>() {
+                            self.push_to_timely.advance_to(advance_to.borrow().epoch);
+                        }
```
davidselassie, line 85: `else { panic!("Unknown input action") }`?

src/execution/mod.rs
```diff
@@ -42,10 +75,19 @@ impl Pump {
...
+                    }
+                    Err(err) => {
+                        std::panic::panic_any(err);
```
davidselassie, line 88: Is there a way to use `with_traceback!` instead here? This will be a Python error and we'd probably still want to log it before throwing using the same machinery.

src/execution/mod.rs
```diff
@@ -68,36 +110,29 @@ impl Pump {
...
-        && !interrupt_flag.load(Ordering::Relaxed)
-    {
+    while !interrupt_flag.load(Ordering::Relaxed) && !probe.done() {
         if !pumps_with_input_remaining.is_empty() {
+            // We have input remaining, pump the pumps, and step the workers while
+            // there is work less than the current epoch to do.
             let mut updated_pumps = Vec::new();
             for mut pump in pumps_with_input_remaining.into_iter() {
                 pump.pump();
+                worker.step_while(|| probe.less_than(&pump.push_to_timely.time()));
```
davidselassie, line 126: I'm really trying to wrap my head around this because I think it changes the behavior of our dataflows. Not that that's a bad thing! I just want to know for sure what behavior will manifest.

Doing a very careful read of [Timely's Monitoring Probes docs](https://timelydataflow.github.io/timely-dataflow/chapter_3/chapter_3_2.html): "Work until all data from before the input epoch has been captured."

This seems like the most intuitive behavior thinking of epochs as "batches" as long as there isn't any operators which "delay" in the dataflow (like anything that accumulated state but then emits at the oldest timestamp, not the latest). If there was, then this would cause hanging? Because then you'd need to keep introducing new data to get output, but this would prevent that. Perhaps this just generally a good constraint on our operators.

pysrc/bytewax/inputs.py
```diff
@@ -3,12 +3,47 @@
...
 import datetime
 import heapq
+
+from .bytewax import AdvanceTo, Send
+
 from dataclasses import dataclass
 from typing import Any, Callable, Iterable, Tuple
 
 
+def yield_epochs(fn: Callable):
```
davidselassie, line 16: I think the type signature on this would be `yield_epochs(iter: Iterable[Tuple[int, Any]]) -> Iterable[Any]`? Perhaps in the future we could add a base class for `AdvanceTo` and `Send` like `InputAction` which would help explain the final `Any`.

davidselassie (approved):
Looks good to me. Some random thoughts if you want to deal with in a followup PR or worry about later.

src/execution/mod.rs
```diff
@@ -8,13 +8,45 @@ use std::time::Duration;
...
+#[pyo3(text_signature = "(epoch)")]
+pub(crate) struct TdAdvanceTo {
+    #[pyo3(get)]
+    epoch: u64,
+}
+
+#[pymethods]
+impl TdAdvanceTo {
+    #[new]
+    fn new(epoch: Py<PyLong>) -> Self {
```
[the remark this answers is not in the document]
whoahbot: Good catch! It does indeed.

src/execution/mod.rs
davidselassie: Further head wrapping: does this make the dataflow... blocking? Non-concurrent? This is sort of saying that we won't have multiple epochs in flight inside the dataflow at the same time and I'm trying to think through what that means for resource utilization. Since every worker can do every task, maybe this isn't an issue; every worker churns through as much of an epoch as fast and thus if it finishes faster it'll get new input faster. But since progress tracking is across the whole cluster, imbalances in work could mean that some workers are waiting without work until a batch of slow keys on one worker is finished? I think that's less of a problem then the "each step is it's own compute" and provisioning for that. I don't think this is a reason to not make this change but it feels like there is another knob here we might need to eventually expose.

src/execution/mod.rs
whoahbot: I did give this a shot, but wasn't able to get the shape of the `Result` to fit with our macro. Happy to give it another go if you have a different construction I can try.

src/execution/mod.rs
whoahbot: > I'm really trying to wrap my head around this because I think it changes the behavior of our dataflows.

Agreed!

This was the most illuminating part of the documentation for me:

> `probe.less_than(input.time())`
> This checks if there are any times strictly less than what the input is positioned to provide next. If so, it means we could keep doing work and making progress, because we know that the system could catch up to input.time() as we can't produce anything less than this from input.

> This seems like the most intuitive behavior thinking of epochs as "batches"

Agreed. Without exposing probes, and the ability to step a worker, I'm not sure how else we can influence progress in the system. It may be that we need to revisit this when the need for a "delay" operator arises.

pysrc/bytewax/inputs.py
whoahbot: I think since we are using this as a decorator, it still has to be a Callable, since we call the fn to get the `Iterable`, but I can refine the types here to be `Callable -> Iterable[Tuple[int, Any]]`.

I did have a base class at one point that both Send and AdvanceTo derived from, to deal with some type inference problems I was running into, but removed them after solving that initial problem. I can add it back if you think it adds clarity here.

davidselassie (approved):

blakestier (approved):

PR #60, opened 2022-04-21
Move operator descriptions into API docs

davidselassie:
Move the detailed description of each operator into the docstrings / API docs and link to them from an overview in the longform docs. This'll fix https://github.com/bytewax/docs/issues/56#issuecomment-1102849381 and means we don't have to have duplicate detailed documentation for each operator.


konradsienkowski (commented):
Looks good to me, added an update of the `metadata.json` file with a new structure

whoahbot (approved):
:sparkles:

PR #63, opened 2022-04-27
Fix broken links in longform documentation

konradsienkowski:
Done:
- [x] Fix broken links in `Understanding epochs` and `A simple example` articles
- [x] Fix broken links in `Building Sessions from Search Logs` example
- [x] Add links to AdvanceTo and Emit dataclasses in  `Understanding epochs`
- [x] Add 'Benefits' H2 in `Overview` article

To do - we need to fix broken links in API Docs submodules:
- [x] All links in [bytewax.parse](https://docs-vue.staging.bytewax.io/apidocs/parse), removing 'index.html' from their target should solve that problem
- [x] Links in [bytewax.testing](https://docs-vue.staging.bytewax.io/apidocs/testing) require the same treatment as above

### DIFF
--- a/apidocs/html/bytewax/index.html
+++ b/apidocs/html/bytewax/index.html
@@ -114,10 +114,8 @@ locally on one machine.</p>
 <dd>Dataflow to run.</dd>
 <dt><strong><code>input_builder</code></strong></dt>
 <dd>Returns input that each worker thread should
-process. Should yield either <code><a title="bytewax.AdvanceTo" href="#bytewax.AdvanceTo">AdvanceTo</a></code> or <code>Send</code> to
-advance the epoch, or input new data into the dataflow.
-If you are recovering a stateful dataflow, you must ensure
-your input resumes from the last finalized epoch.</dd>
+process. Should yield either <code><a title="bytewax.AdvanceTo" href="#bytewax.AdvanceTo">AdvanceTo</a></code> or <code><a title="bytewax.Emit" href="#bytewax.Emit">Emit</a></code> to
+advance the epoch, or input new data into the dataflow.</dd>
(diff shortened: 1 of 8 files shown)


davidselassie (approved):
This looks good. I'm happy to show you how to update the docstrings for the API docs later.

davidselassie:
I did some digging, and it seems like the remaining broken links are between auto-generated links and auto-generated anchors by pdoc, and are cross-module. E.g. `bytewax.parse.cluster_args()` to `bytewax.run_cluster()`.

I don't think fixing these is possible without changing pdoc's code, modifying the URL rewriting in the Vue app (if that's what it's called?), or having an HTML post-processing step to munge URLs.

For example, it seems like the Vue app is rewriting the HTML file at the path `/apidocs/bytewax/inputs.html` to the public URL path `/apidocs/inputs/` but then it's also rewriting the path `/apidocs/bytewax/index.html` to `/apidocs/`. That is not the same pattern.

pdoc has no built in way of telling it to remember an arbitrary mapping of paths to make sure links are rewritten. So we could think about ways to add that to pdoc, or we have to add it at a later step, or have our final URLs match the relative structure of the output files.

I vote for the latter and I see that it will make the URLs on the API docs a little uglier, but I think it's the easiest of the options for now.

konradsienkowski:
@davidselassie you're right - the Vue app is modifying paths/URLs anyway, so I added functions to handle also that problem. Those are already deployed on staging 🙂 

We can also add the `bytewax.` prefix to the paths if we want now - previously it caused problems because Vite doesn't support dots inside URLs by default. However, I would suggest to stay with removing the '.html' extension from routes

PR #73, opened 2022-05-09
Automatic state recovery / garbage collection

davidselassie:
This is an overhaul of the state recovery system to support two big
new features:

1. _Automatic recovery_ - Bytewax will backup in the recovery store
the **dataflow frontier** which is the oldest in-progress epoch. This
represents the epoch that if the dataflow crashed right now, one
should resume from. The startup machinery now reads this epoch from
the recovery store and will pass it to the input builders so the
system will start up exactly where it failed.

2. _State garbage collection_ - By keeping track of the dataflow
frontier, we can can know what state is no longer necessary for
recovery and GC it from the state store.

See the module docstring in `bytewax.recovery` for what is necessary
from a user's perspective and see the rustdoc in the `recovery` module
to see documentation on how this works internally.

### DIFF
--- a/docs/articles/getting-started/execution.md
+++ b/docs/articles/getting-started/execution.md
@@ -91,19 +91,21 @@ Our next entry point introduces a more complex API to allow for more performance
 
 ### Builders
 
-You provide a input and output **builders**, functions that are called on each worker and will produce the input and handle the output for that worker. The input builder function should return an iterable that yields `(epoch, item)` tuples. The output builder function should return a callback **output handler** function that can be called with each epoch and item of output produced.
+You provide a input and output **builders**, functions that are called on each worker and will produce the input and handle the output for that worker. The input builder function should return an iterable that yields `Emit(item)` or `AdvanceTo(epoch)`. The output builder function should return a callback **output handler** function that can be called with each epoch and item of output produced.
+
+The input builder also recieves the epoch to resume processing on in the case of restarting the dataflow. The input builder must somehow skip ahead in its input data to start at that epoch.
 
 The per-worker input of `spawn_cluster()` has the extra requirement that the input data is not duplicated between workers. All data from each worker is introduced into the same dataflow. For example, if each worker reads the content of the same file, the input will be duplicated by the number of workers. Data sources like Apache Kafka or Redpanda can provide a partitioned stream of input which can be consumed by multiple workers in parallel without duplication, as each worker sees a unique partition.
 
-`run_cluster()` blocks until all output has been collected.
+`spawn_cluster()` blocks until all output has been collected.
 
 ```python doctest:SORT_OUTPUT
 from bytewax import spawn_cluster, AdvanceTo, Emit
 from bytewax.testing import test_print
 
 
-def input_builder(worker_index, worker_count):
-    for epoch, x in enumerate(range(3)):
+def input_builder(worker_index, worker_count, resume_epoch):
+    for epoch, x in enumerate(range(resume_epoch, 3)):
         yield AdvanceTo(epoch)
         yield Emit({"input_from": worker_index, "x": x})
 
@@ -169,8 +171,8 @@ The input builder function should return an iterable that yields `(epoch, item)`
 from bytewax import cluster_main
 
 
-def input_builder(worker_index, worker_count):
-    for epoch, x in enumerate(range(3)):
+def input_builder(worker_index, worker_count, resume_epoch):
+    for epoch, x in enumerate(range(resume_epoch, 3)):
         yield AdvanceTo(epoch)
         yield Emit({"input_from": worker_index, "x": x})
 
--- a/docs/articles/getting-started/execution.md
+++ b/docs/articles/getting-started/execution.md
@@ -91,7 +91,7 @@ Our next entry point introduces a more complex API to allow for more performance
 
 ### Builders
 
-You provide a input and output **builders**, functions that are called on each worker and will produce the input and handle the output for that worker. The input builder function should return an iterable that yields `Emit(item)` or `AdvanceTo(epoch)`. The output builder function should return a callback **output handler** function that can be called with each epoch and item of output produced.
+You provide input and output **builders**, functions that are called on each worker and will produce the input and handle the output for that worker. The input builder function should return an iterable that yields `Emit(item)` or `AdvanceTo(epoch)`. The output builder function should return a callback **output handler** function that can be called with each epoch and item of output produced.
 
 The input builder also recieves the epoch to resume processing on in the case of restarting the dataflow. The input builder must somehow skip ahead in its input data to start at that epoch.
 
--- a/pysrc/bytewax/recovery.py
+++ b/pysrc/bytewax/recovery.py
@@ -1,32 +1,52 @@
 """Bytewax's state recovery machinery.
 
+Preparation
+-----------
+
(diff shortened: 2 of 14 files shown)


davidselassie:
What I was previously referring to as the "final frontier" in my spec docs, I now want to call the **dataflow frontier**or "frontier of a dataflow". I think it is a better analogy to the "frontier of an input" that Timely already has. And I think the use of the word "final" made it seem like it was what was complete, but actually it's the last epoch that is incomplete.

davidselassie:
I'm going to actually work on a Kafka recovery store connector next if this API looks good.

docs/articles/getting-started/execution.md
```diff
@@ -91,19 +91,21 @@ Our next entry point introduces a more complex API to allow for more performance
 
 ### Builders
 
-You provide a input and output **builders**, functions that are called on each worker and will produce the input and handle the output for that worker. The input builder function should return an iterable that yields `(epoch, item)` tuples. The output builder function should return a callback **output handler** function that can be called with each epoch and item of output produced.
+You provide a input and output **builders**, functions that are called on each worker and will produce the input and handle the output for that worker. The input builder function should return an iterable that yields `Emit(item)` or `AdvanceTo(epoch)`. The output builder function should return a callback **output handler** function that can be called with each epoch and item of output produced.
+
+The input builder also recieves the epoch to resume processing on in the case of restarting the dataflow. The input builder must somehow skip ahead in its input data to start at that epoch.
```
awmatheson, line 96: If I am using kafka and I am using a windowed approach to my epochs. Does that mean that I have to track my minimum offset associated with that epoch somewhere else so that I can restart correctly from the minimum offset associated with the most recently completed window. Or am I misunderstanding?

awmatheson, line 96: If I am using kafka and I am using a windowed approach to my epochs. Does that mean that I have to track my minimum offset associated with that epoch somewhere else so that I can restart correctly from the minimum offset associated with the most recently completed window. Or am I misunderstanding?

davidselassie: This is a very good point. Perhaps. Recovery requires your input builder to be able to replay all input from a specific epoch (the one passed in in `resume_epoch`). So if you're reading from Kafka you effectively need a function like `offset_for(epoch)`. If you're windowing into fixed size buckets (10 msg / epoch), then it'd be possible to: `resume_offset = resume_epoch * 10`. But if you're windowing by time, then there's no deterministic way to go from epoch back to offset, and yes you'd have to keep track of this yourself, which is annoying.

Idea: perhaps we could have something like "annotated epochs". Where an epoch is kind of an `(int, Any)` tuple. Bytewax instructs Timely to only use the `int` part for ordering, but we persist the `Any` part as well and it's metadata that you can use for recovery.

(This also could be the system that'd enable something like transactional writes on output: you put the txn ID into the epoch metadata and the output handler can close the txn.) (I also think this would be unrelated to Timely's multi-dimensional timestamp / epoch support, even though I talk about it as a 2-tuple here.)

Thoughts on that? Let me have a think about this because this is will be a frustrating and very common problem.

src/operators/mod.rs
```diff
@@ -320,105 +366,141 @@ impl<S: Scope, K: ExchangeData + Hash + Eq, V: ExchangeData> StatefulMap<S, K, V
...
+
+                                let (updated_state_for_key, output) =
+                                    mapper(&key, state_for_key, value);
+
+                                downstream_output_session.give_iterator(
+                                    output.into_iter().map(|item| (key.clone(), item)),
+                                );
+
+                                // Save the state so we can use it if
+                                // multiple values within this epoch. We
```
whoahbot, line 453:
```suggestion
                                // there are multiple values within this epoch. We
```

src/operators/mod.rs
```diff
@@ -452,3 +534,255 @@ pub(crate) fn stateful_map(
...
+                        for (step_id, key, update_type) in tmp_recovery_keys.drain(..) {
+                            recovery_store_summary
+                                .entry((step_id, key))
+                                .or_default()
+                                .insert((epoch.clone(), update_type));
+                        }
+                    });
+                }
+
+                // TODO: This runs on every operator activation. We
```
whoahbot, line 717: I'm trying to reconcile this with what Frank says about operator scheduling, are we in the explicitly activating context?

https://gitter.im/timely-dataflow/Lobby?at=5ebe885c49a1b731847828ab

Specifically:

> Operators get scheduled if 1. they receive new input, 2. the frontiers of any of their inputs change, or 3. they are explicitly activated. (or in a few other cases that are optimistic versions of 1. and 2.).
Operators are shut down if 1. their input frontiers are empty, 2. they hold no capabilities, and 3. they returned "false" from their most recent scheduling.
Each time you call worker.step() the worker starts to move through the list of operators that should be scheduled (for any of the three reasons above). Having scheduled each worker, it tests the shutdown condition and zeros out the worker if it should be shut down.

pysrc/bytewax/recovery.py
```diff
@@ -1,32 +1,52 @@
...
 
-Bytewax will automatically save and recover state from the recovery
-store during the execution of stateful operators. You know an operator
-is stateful when it takes a `step_id` argument.
+Recovery
+--------
+
+If the dataflow fails, first you must fix whatever underlying fault
+caused the issue. That might mean deploying new code which fixes a
+bug, removing bad data from a Kafka stream, or resolving an issue with
```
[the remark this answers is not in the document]
davidselassie: Good point. I think I was thinking of bumping consumer group offsets to skip something, but that also isn't relevant here because we aren't using those. Will change.

src/operators/mod.rs
davidselassie: I'm not totally sure on the call stack that leads to this closure being called. But perhaps I'm being fast and loose r.e. "scheduling" and "activation". My understanding was that it was called on what Frank is calling every scheduling. Our goal is to only run this code on 2, but I assume it also runs on 1 & 3 because that's what a generic operator would need. Hence the desire to have an extra check.

Does that answer your question?

whoahbot (approved):

docs/articles/getting-started/execution.md
```diff
@@ -91,19 +91,21 @@ Our next entry point introduces a more complex API to allow for more performance
 
 ### Builders
 
-You provide a input and output **builders**, functions that are called on each worker and will produce the input and handle the output for that worker. The input builder function should return an iterable that yields `(epoch, item)` tuples. The output builder function should return a callback **output handler** function that can be called with each epoch and item of output produced.
+You provide a input and output **builders**, functions that are called on each worker and will produce the input and handle the output for that worker. The input builder function should return an iterable that yields `Emit(item)` or `AdvanceTo(epoch)`. The output builder function should return a callback **output handler** function that can be called with each epoch and item of output produced.
```
blakestier, line 94:
```suggestion
You provide input and output **builders**, functions that are called on each worker and will produce the input and handle the output for that worker. The input builder function should return an iterable that yields `Emit(item)` or `AdvanceTo(epoch)`. The output builder function should return a callback **output handler** function that can be called with each epoch and item of output produced.
```

While you're here 🤷

blakestier (approved):

docs/articles/getting-started/execution.md
davidselassie: I think we'll tackle this in more detail when we look at prefab input builders.

PR #77, opened 2022-05-17
Fix inputs helper

blakestier:
We forgot to update the `yield_epochs` decorator when we added the `resume_epoch` argument.

### DIFF
--- a/pytests/test_execution.py
+++ b/pytests/test_execution.py
@@ -215,3 +215,46 @@ def test_cluster_main_can_be_ctrl_c(mp_ctx):
 
         assert test_proc.exitcode == 99
         assert len(out) < 1000
+
+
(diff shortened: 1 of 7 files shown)


pytests/test_execution.py
```diff
@@ -215,3 +215,46 @@ def mapper(item):
 
         assert test_proc.exitcode == 99
         assert len(out) < 1000
+
+
+def test_yield_epochs_with_cluster_main(mp_ctx):
```
blakestier, line 220: is this more overhead than we need? Maybe. But this would have caught the missed signature change required.

whoahbot: I think it's great. Thanks for adding this test.

whoahbot (approved):

PR #84, opened 2022-06-08
Kafka Consumer

blakestier:
## Initial Kafka Consumer

This work introduces a Kafka Consumer written in Rust with minimal configuration options. 

In order to implement it, the shape of the API has changed in that commands formerly accepting an `input_builder` (`cluster_main`, for example) now accept a configuration object of type `KafkaInputConfig` or `ManualInputConfig` (the latter accepting an `input_builder` as its argument).

While we anticipate adding more configuration flexibility in the future, this made sense as a first pass.


docs/articles/getting-started/execution.md
```diff
@@ -109,6 +109,7 @@ def input_builder(worker_index, worker_count, resume_epoch):
         yield AdvanceTo(epoch)
         yield Emit({"input_from": worker_index, "x": x})
 
+input_config = ManualInputConfig(input_builder)
```
blakestier, line 112: Would this be cuter as a python input helper? So we could inline `manual_config(input_builder)` into the `spawn_cluster` call? Or do we want to keep config choices overt for now? (that's where i'm leaning atm tho it's kinda clunky)

src/input/mod.rs
```diff
@@ -0,0 +1,490 @@
...
+            rt,
+            consumer,
+            current_epoch: 0,
+            desired_batch_size: config.batch_size,
+            current_batch_size: 0,
+        }
+    }
+
+    pub fn next(&mut self) -> TimelyAction {
+        if self.current_batch_size == self.desired_batch_size {
```
blakestier, line 425: I'm not sure why but we see this for a few rounds before data starts coming in. As in, we don't really emit a `1` or `2` epoch...

src/input/mod.rs
```diff
@@ -0,0 +1,490 @@
...
+#[pyo3(text_signature = "(brokers, group_id, topics)")]
+pub(crate) struct KafkaInputConfig {
+    #[pyo3(get)]
+    pub brokers: String,
+    #[pyo3(get)]
+    pub group_id: String,
+    #[pyo3(get)]
+    pub topics: String,
+    #[pyo3(get)]
+    pub batch_size: u64,
```
blakestier, line 127: `epoch_size`?

pysrc/bytewax/inputs.py
```diff
@@ -8,9 +8,7 @@
 import heapq
 from dataclasses import dataclass
 from typing import Any, Callable, Iterable, Tuple
-
-from .bytewax import AdvanceTo, Emit
-
+from .bytewax import AdvanceTo, Emit, KafkaInputConfig, ManualInputConfig, InputConfig
```
blakestier, line 11: The module of "input" is ballooning a bit but I pushed on modularizing more since we still only have the one "source"

src/input/mod.rs
```diff
@@ -0,0 +1,490 @@
...
+/// manual input configuration.
+///
+///
+/// See also: `inputs.yield_epochs()`
+///
+/// >>> def input_builder(worker_index, worker_count, resume_epoch):
+/// ...     for i in range(10):
+/// ...         yield AdvanceTo(i) # Advances the epoch to i
+/// ...         yield Emit(i) # Adds the input i at epoch i
+#[pyclass(module = "bytewax")]
```
davidselassie, line 40: Might be out of scope for this PR, but maybe think about moving `bytewax.Emit` into `bytewax.inputs.Emit`. In that case, these Python module names will need to be changed.

src/lib.rs
```diff
@@ -7,6 +7,7 @@ use std::time::Duration;
 
 pub(crate) mod dataflow;
 pub(crate) mod execution;
+pub(crate) mod input;
```
davidselassie, line 10: Maybe call this Rust file / module `inputs` instead of `input` to match the resulting Python-side package?

src/input/mod.rs
davidselassie: Is that due to an empty poll incrementing the epoch?

davidselassie (commented):
This looks great! I have a smattering of comments, mostly smaller details that come to mind. I think the only design-level questions I have are surrounding deserialization and the keys of messages. I know there's much more work to be done re recovery, but I think you can work on that in a separate PR later.

src/input/mod.rs
```diff
@@ -0,0 +1,490 @@
...
+    }
+
+    /// Pickle as a tuple.
+    fn __getstate__(&self, py: Python) -> (&str, Py<PyAny>) {
+        ("ManualInputConfig", self.generator.clone_ref(py))
+    }
+
+    /// Egregious hack see [`SqliteRecoveryConfig::__getnewargs__`].
+    fn __getnewargs__(&self) -> (&str,) {
+        (&"",)
```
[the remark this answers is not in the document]
blakestier: yes, that's right. I'll see

src/input/mod.rs
```diff
@@ -0,0 +1,490 @@
...
+    }
+
+    fn input_remains(&mut self) -> bool {
+        !self.empty
+    }
+}
+
+/// Encapsulates the process of pulling data out of the input Python
+/// iterator and feeding it into Timely.
+struct PyPump {
```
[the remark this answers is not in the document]
blakestier: good call. that moved over time

src/lib.rs
blakestier: i was actually thinking of `inputs`-> `input` python-side, but that's more disruptive

src/input/mod.rs
```diff
@@ -0,0 +1,490 @@
...
+                match self.consumer.poll(Duration::from_millis(1000)) {
+                    None => {
+                        dbg!("No messages available, incrementing epoch");
+                        self.current_batch_size = 0;
+                        self.current_epoch += 1;
+                        TimelyAction::AdvanceTo(self.current_epoch)
+                    }
+                    Some(r) => match r {
+                        Err(e) => panic!("Kafka error! {}", e),
+                        Ok(s) => match s.payload_view::<str>() {
```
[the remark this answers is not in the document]
blakestier: Yep there's a TODO for that

src/lib.rs
davidselassie: It's not a super big deal, but singular [`input` is a Python built-in function](https://docs.python.org/3/library/functions.html#input) and sometimes it gets funny syntax highlighting, so that's why I originally had it as plural.

src/input/mod.rs
```diff
@@ -0,0 +1,490 @@
...
+            return TimelyAction::AdvanceTo(self.current_epoch);
+        } else {
+            // TODO async loop to populate batch buffer
+            self.rt.block_on(async {
+                // I have no sense of an appropriate timeout. I suppose users provide?
+                match self.consumer.poll(Duration::from_millis(1000)) {
+                    None => {
+                        dbg!("No messages available, incrementing epoch");
+                        self.current_batch_size = 0;
+                        self.current_epoch += 1;
```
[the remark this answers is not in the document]
blakestier: This should make more sense after I add in the batching loop, so we'll poll until the batch is full, then increment the epoch

src/lib.rs
blakestier: oh yep that rings a 🔔

src/input/mod.rs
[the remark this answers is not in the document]
blakestier: (I should say, there's a TODO for the bytes piece . . . I hadn't thought about the keys, but will do so now)

src/input/mod.rs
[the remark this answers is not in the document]
whoahbot: I think the semantics we discussed were "up to n messages per batch". Similar to the Python Kafka client batching semantics.

If we wait until the batch is full, we will be stuck waiting for new data when the topic has no new messages.

pysrc/bytewax/__init__.py
```diff
@@ -5,11 +5,13 @@
...
 
 """
-from .bytewax import AdvanceTo, cluster_main, Dataflow, Emit, run_main
+from .bytewax import AdvanceTo, cluster_main, Dataflow, Emit, KafkaInputConfig, ManualInputConfig, run_main
 from .execution import run, run_cluster, spawn_cluster
 
 __all__ = [
     "Dataflow",
+    "KafkaInputConfig",
+    "ManualInputConfig",
```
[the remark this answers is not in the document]
blakestier: I couldn't really decide on this. We directly import AdvanceTo and Emit and other "mission critical" kind of pieces of the dataflow under `bytewax`, so that seemed right for the two required configs, but I was thinking the parent struct would stay in `inputs`.  But maybe we should just move the other? Maybe I need to read about philosophies on namespacing? Maybe we need one?

src/input/mod.rs
blakestier: I don't see how yet. We don't increment the batch until we get a message, and the desired_batch_size defaults to `1`, so it's unclear to me how we fall in here

src/input/mod.rs
blakestier: I don't know. I'm undecided on our semantics around epochs I think. I started with `messages_per_epoch` but it felt clunky. I'll revisit it more

src/input/mod.rs
```diff
@@ -0,0 +1,490 @@
...
+///
+/// Returns:
+///
+///     Config object. Pass this as the `input_config` argument to
+///     your execution entry point.
+
+#[pyclass(module = "bytewax.inputs", extends = InputConfig)]
+#[pyo3(text_signature = "(generator)")]
+pub(crate) struct ManualInputConfig {
+    generator: Py<PyAny>,
```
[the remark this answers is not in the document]
blakestier: I think it makes sense to rename this bc of the ambiguity around `generator` as a function vs. `generator` as an object-- maybe just back to `input_builder` bc why not.

I am curious about examples where we are returning iterables that are not generators (in manual configuration scenarios using `spawn_cluster` and `cluster_main`), given the required use of `AdvanceTo` and `Emit` now. Do you have one in mind?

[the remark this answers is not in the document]
davidselassie: I guess the term I'm looking for here is [generator iterator](https://docs.python.org/3/glossary.html#term-generator-iterator). I've been using "generator" to refer to the iterator returned, but it looks like it more appropriately refers to the function. A generator is a subset of all functions, though, and we don't require the input builder to be a generator, just that it's a function that returns an iterable.

I don't think we currently have any examples that use non-generator iterators because all of them are using `run` or `run_main`. But when we drop those entry points, we'll be doing lots of things like this in tests:

```py
def ib(i, n, e):
    return [
        Emit("a"),
        Emit("b"),
        AdvanceTo(2),
        Emit("c"),
    ]
```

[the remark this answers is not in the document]
blakestier: Yeah thanks for that example. I can see how that maintains the contract-- just trying to imagine a user using anything other than a generator. But how often will they rely on manual input in general? 🤷 

Anyhow, back to `input_builder`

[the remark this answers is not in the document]
davidselassie: Yeah, I think long term this `input_builder` structure is not long for this world.

If you want to get creative, you could make a custom iterator transform that isn't a generator, and it's possible some of these exist:

```py
class WrapEmit:
    def __init__(self, wrap):
        self.wrap = wrap

    def next():
        return Emit(self.wrap.next())
```

Actually I'm pretty sure `map(wrap, lambda x: Emit(x))` doesn't result in a generator iterator actually.

```py
In [1]: wrap = [1, 2, 3]

In [2]: from bytewax import Emit

In [4]: inp = map(lambda x: Emit(x), wrap)

In [5]: inp
Out[5]: <map at 0x7f7331b48a60>
```

Looks like it's a `map` object.

None of that really matters, though.

src/inputs/mod.rs
```diff
@@ -262,8 +260,8 @@ impl Pump for KafkaPump {
         match self.kafka_consumer.next() {
             TimelyAction::AdvanceTo(epoch) => self.push_to_timely.advance_to(epoch),
             TimelyAction::Emit(item) => Python::with_gil(|py| {
-                let py_any_string = TdPyAny::from(PyString::new(py, &item));
-                self.push_to_timely.send(py_any_string);
+                let key_payload = vec![item.0, item.1].into_py(py);
```
davidselassie, line 263: Minor point, but does `(item.0, item.1)` work? Using a `Vec` and the resulting Python `list` allows / implies it to be resizable. Also you don't have to allocate a vec.

davidselassie (commented):
This is looking good! I have a few more smaller comments now that there's an example.

examples/simple_kafka_input.py
```diff
@@ -1,12 +1,19 @@
+import json
 from bytewax import cluster_main, Dataflow
 from bytewax.inputs import KafkaInputConfig
 
+
 def output_builder(worker_index, worker_count):
-    def output_fn(epoch_dataframe):
-        print(epoch_dataframe)
+    def output_fn(epoch_keypayload):
+        e, data = epoch_keypayload
```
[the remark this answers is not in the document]
blakestier: Wow i didn't even know about the latter 🌈

src/inputs/mod.rs
```diff
@@ -0,0 +1,451 @@
...
+pub(crate) struct InputConfig;
+
+/// Use [Kafka](https://kafka.apache.org) as the input source. Currently
+/// does not support recovery. Kafka messages will be passed through the dataflow
+/// as byte two-tuples of Kafka key and payload.
+///
+///
+/// Args:
+///
+///     brokers: (string) Comma-separated of broker addresses.
```
blakestier, line 104: We could make this a string array, too.

src/inputs/mod.rs
```diff
@@ -0,0 +1,451 @@
...
+
+impl KafkaPump {
+    fn new(config: PyRef<KafkaInputConfig>, push_to_timely: InputHandle<u64, TdPyAny>) -> Self {
+        let context = CustomContext;
+
+        let rt = tokio::runtime::Builder::new_current_thread()
+            .enable_all()
+            .build()
+            .unwrap();
+
```
blakestier, line 265: Link to [default values](https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md)?

src/inputs/mod.rs
```diff
@@ -0,0 +1,450 @@
...
+        let consumer: BaseConsumer<CustomContext> = rt.block_on(async {
+            ClientConfig::new()
+                .set("group.id", config.group_id.clone())
+                .set("bootstrap.servers", config.brokers.clone())
+                .set("enable.partition.eof", "false")
+                .set("session.timeout.ms", "6000")
+                .set("enable.auto.commit", "true")
```
blakestier, line 272: Dan suggested making this configurable so I'll add that.

Anything else seem like it should be configurable? 👀

davidselassie (approved):
This looks good to me! I'm sure we'll discover things we want to change  and config options to provide once we get more experience with it, but I think it's a good place to jump off from.

examples/feast/feast_dataflow.py
```diff
@@ -6,6 +6,7 @@
 from kafka import KafkaConsumer
 
 from bytewax import cluster_main, Dataflow, inputs
+from bytewax.inputs import ManualInputConfig
```
whoahbot, line 9: Do we want to re-create this example using the KafkaInputConfig?

whoahbot (approved):
Looks good! I left you a question about the Feast dataflow.

examples/feast/feast_dataflow.py
blakestier: It won't work without modifying the parquet file data bc we have so little epoch-management for this new consumer rn (so the sorting and the tumbling required aren't available). I also thought maybe it was fine to show this option until we have those ("use ours *or* make your own!")

PR #90, opened 2022-07-12
Windowing (Tumble, Batch)

blakestier:
Introduces TumblingWindow and Batching for input from _both_ manual input configuration and Kafka input configuration.

As a result, epoch assignment now lives in our rust code, and `AdvanceTo` and `Emit` have been removed.

The following tasks remain to complete this work:

- [ ] Rename configs-- they are a bit unwieldy
- [ ] Consider impact of epoch management moving for users wanting to recover from an epoch.
- [ ] Add back single batch. _I had removed it, but after looking at tests thought some users would need it back_
- [ ] Add tests (tumbling, batch, default)
- [ ] Return entire Kafka message? _Currently just returning payload_
- [ ] Modularize all the new stuff in [inputs.rs](http://inputs.rs/) (this file is out of hand)
- [ ] KafkaConsumer could probably go back into the iterator again 
- [ ] Make deserialize its own function? 
- [ ] Documentation


pytests/test_operators.py
```diff
@@ -25,8 +22,8 @@ def add_one(item):
     assert sorted(out) == sorted(
         [
             (0, 1),
```
blakestier, line 24: Here and elsewhere in this file I debated the value of checking the epochs at all.

davidselassie:
This ended up turning into https://github.com/bytewax/bytewax/pull/96

PR #93, opened 2022-07-13
Fix argument formatting in API docs

davidselassie:
- Fixes docstring argument formatting
- Updates API docs


konradsienkowski (approved):
Looks good to me 👍 Added two more things:
- 'Recovery' article to the metadata.json, as it's broken/orphaned now 
- H1 Header to the Shopping cart example

PR #95, opened 2022-07-22
Fold window operator

Psykopear:
# Fold Window Operator
This PR introduces the `fold_window` operator.

`fold_window`, like `reduce_window`, lets you combine all the items for a key within a given epoch into an accumulator.

The main difference is that `fold_window` accepts a `builder` function as first parameter, which is called the first time a new key appears in the configured window.

The second parameter is the `folder` function, which acts just like the `reducer` in `reduce_window`, takes the `accumulator` as first parameter and `current_value` as second one, and returns the updated `accumulator`.

The `builder` function takes the first key as parameter, and it should be used to build the empty accumulator.

This is an example usage (from the test for the python side):

```python
    def gen():
        yield from [
            {"user": "a", "type": "login"},
            {"user": "a", "type": "post"},
            {"user": "a", "type": "post"},
            {"user": "b", "type": "login"},
            {"user": "a", "type": "post"},
            {"user": "b", "type": "post"},
            {"user": "b", "type": "post"},
        ]

    def extract_id(event):
        return (event["user"], event)

    def build(key):
        return defaultdict(lambda: 0)

    def count(results, event):
        results[event["type"]] += 1
        return results

    clock_config = TestingClockConfig(item_incr=timedelta(seconds=4))
    window_config = TumblingWindowConfig(length=timedelta(seconds=10))

    out = []
    flow = Dataflow(TestingInputConfig(gen()))
    flow.map(extract_id)
    flow.fold_window("sum", clock_config, window_config, build, count)
    flow.capture(TestingOutputConfig(out))

    run_main(flow)

    assert len(out) == 3
    assert ("a", {"login": 1, "post": 2}) in out
    assert ("a", {"post": 1}) in out
    assert ("b", {"login": 1, "post": 2}) in out
```

## TODO
- [x] Write the docstring
- [x] Move from fold_epoch to fold_window
- [x] User the new input operators

### DIFF
--- a/src/execution/mod.rs
+++ b/src/execution/mod.rs
@@ -19,7 +19,7 @@ use crate::recovery::StateWriter;
 use crate::recovery::{build_recovery_readers, build_recovery_writers, FrontierUpdate};
 use crate::recovery::{default_recovery_config, ProgressWriter};
 use crate::recovery::{ProgressReader, StateCollector};
-use crate::window::{build_clock, build_windower, reduce_window, StatefulWindowUnary};
+use crate::window::{build_clock, build_windower, reduce_window, StatefulWindowUnary, fold_window};
 use log::debug;
 use pyo3::exceptions::{PyRuntimeError, PyValueError};
 use pyo3::prelude::*;
@@ -435,6 +435,33 @@ fn build_production_dataflow<A: Allocate>(
                 }
                 Step::Filter { predicate } => {
                     stream = stream.filter(move |item| filter(&predicate, item));
+                }
+                Step::FoldWindow { step_id, clock_config, window_config, builder, folder } => {
+                    let windower = build_windower(py, window_config)?;
+                    let clock = build_clock(py, clock_config)?;
+
+                    stream = stream
+                        .map(extract_state_pair)
+                        .stateful_window_unary(
+                            step_id,
+                            clock,
+                            windower,
+                            move |key, state, next_value| {
+                                fold_window(&builder, &folder, key, state, next_value)
+                            },
+                            StateKey::route,
+                            &state_loading_stream,
+                            &mut state_backup_streams,
+                        )
+                        .map(|(key, result)| {
+                            result
+                                .map(|value| (key.clone(), value))
+                                .map_err(|err| (key.clone(), err))
+                        })
+                        // For now, filter to just reductions and
+                        // ignore late values.
+                        .ok()
+                        .map(wrap_state_pair);
                 }
                 Step::Inspect { inspector } => {
                     stream = stream.inspect(move |item| inspect(&inspector, item));
--- a/src/window/mod.rs
+++ b/src/window/mod.rs
@@ -775,12 +775,14 @@ pub(crate) fn fold_window(
 ) -> (Option<TdPyAny>, impl IntoIterator<Item = TdPyAny>) {
     match next_value {
         Some(value) => Python::with_gil(|py| {
-            let initial_acc = acc
-                .unwrap_or_else(|| with_traceback!(py, builder.call1(py, (key.clone(),))).into());
             // Save the value's debug string here first, because its ownership
             // will be moved to the `folder` function.
             let value_dbg = format!("{value:?}");
-            let updated_acc = with_traceback!(py, folder.call1(py, (initial_acc, value))).into();
+            // Either take acc's value or build the initial one with the builder,
+            // then update it with the folder.
+            let acc = acc
+                .unwrap_or_else(|| with_traceback!(py, builder.call1(py, (key.clone(),))).into());
+            let updated_acc = with_traceback!(py, folder.call1(py, (acc, value))).into();
             debug!("fold_window for key={key:?}: builder={builder:?}, value={value_dbg}) -> updated_acc={updated_acc:?}",);
             (Some(updated_acc), None)
         }),
--- a/src/window/mod.rs
+++ b/src/window/mod.rs
@@ -765,6 +765,30 @@ pub(crate) fn reduce_window(
     }
 }
 
+/// Shim function for [`crate::dataflow::Dataflow::fold_window`].
+pub(crate) fn fold_window(
+    builder: &TdPyCallable,
+    folder: &TdPyCallable,
+    key: &StateKey,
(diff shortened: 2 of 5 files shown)


src/execution/mod.rs
```diff
@@ -146,6 +146,20 @@ fn build_production_dataflow<A: Allocate>(
                 Step::Filter { predicate } => {
                     stream = stream.filter(move |item| filter(&predicate, item));
                 }
+                Step::FoldEpoch { builder, folder } => {
+                    stream = stream.map(extract_state_pair).aggregate(
```
davidselassie, line 150: This comment isn't about this specific line of code, but is a general thing: Based on our discoveries as part of https://github.com/bytewax/bytewax/pull/96 , we probably won't have any operators that combine over epochs any more because epochs are going to be an implementation detail. So instead in the future this would become a `fold_window` operator. But! I don't think you should throw away this PR. We can move this over to use windowing stuff when it is done. I think Dan recorded a video of us discussing what it means to have "operational epochs" but I'm also happy to answer any questions.

davidselassie (commented):
This looks great.

pytests/test_operators.py
```diff
@@ -196,6 +196,41 @@ def count(count, event_count):
     )
 
 
+def test_fold_epoch():
+    def extract_id(event):
+        return (event["user"], event)
+
+    def builder(event):
+        return {event["type"]: 1}
```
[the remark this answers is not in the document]
Psykopear: Ok, will fix both the code and the test

src/execution/mod.rs
Psykopear: Yes I was talking with Dan about this yesterday. Let me know how you want to proceed, we can merge this PR, wait for the windowing one to be merged, and fix this in another PR, or we can wait for the windowing PR to be merged, fix this one and only after merge it.

src/operators/mod.rs
```diff
@@ -65,6 +65,39 @@ pub(crate) fn filter(predicate: &TdPyCallable, item: &TdPyAny) -> bool {
...
+        builder,
+        folder,
+        _key,
+        value,
+        acc
+    );
+    Python::with_gil(|py| {
+        let updated_acc = match acc {
+            Some(acc) => with_traceback!(py, folder.call1(py, (acc.clone_ref(py), value))).into(),
+            None => with_traceback!(py, builder.call1(py, (value, ))).into(),
```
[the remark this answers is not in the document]
Psykopear: Ok.
Just calling it like you propose makes the borrow checker not happy (cannot move out of `*acc` which is behind a mutable reference), and trying to borrow the option's value (like `acc.as_ref()....`) makes the `.into()` call not work because the trait `&TdPyAny: From<pyo3::Py<pyo3::PyAny>>` is not satisfied (only without the reference).

If I `.take()` the whole `Option` out it works though, and I can also avoid the `clone_ref(py)` when calling the folder:

```rust
let initial_acc = acc.take().unwrap_or_else(|| {
    with_traceback!(py, builder.call1(py, (value.clone_ref(py),))).into()
});
let updated_acc = with_traceback!(py, folder.call1(py, (initial_acc, value))).into();
*acc = Some(updated_acc);

```
Since I always set the new acc value, this should be fine.

src/window/mod.rs
```diff
@@ -701,6 +701,52 @@ pub(crate) fn reduce_window(
...
+) -> (Option<TdPyAny>, impl IntoIterator<Item = TdPyAny>) {
+    match next_value {
+        Some(value) => {
+            Python::with_gil(|py| {
+                let updated_acc: TdPyAny = match acc {
+                    // If there's no previous state for this key, use
+                    // the current value.
+                    None => {
+                        let initial_acc: TdPyAny =
+                            with_traceback!(py, builder.call1(py, (value.clone_ref(py),))).into();
```
davidselassie, line 720: Re: my mistake. To match the semantics of the `stateful_map` operator (which also has a `builder`), the builder should be called with the `key` and not the `value`.

pytests/test_operators.py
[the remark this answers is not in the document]
davidselassie: Err, sorry, I misspoke here. I meant `{}`, since this builder should just take the `key` and not the `value` (since the "empty" accumulator wouldn't be able to know about any values yet). The folder will have to add the event type because that's the first point with values.

Or if you want to get fancy, you could return `defaultdict(lambda: 0)` to streamline the folder code.

src/operators/mod.rs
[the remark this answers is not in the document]
davidselassie: Yep, that looks great. I need the borrow checker to check everything I write.

src/execution/mod.rs
davidselassie: Sorry I didn't get back to you sooner. I already merged the windowing and source PR. So that will mean the tests need to be fixed up slightly. But I think that'll mean that our public API has stabilized and there shouldn't be any more changes at least!

pytests/test_operators.py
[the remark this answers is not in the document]
Psykopear: That makes sense, it wasn't clear to me what was the advantage of taking the whole event in the builder, given we call the folder right after, but I assumed it could be useful in some cases. This makes the rust side nicer too, so I'm going to update the code, and then the description of the PR too.

> Or if you want to get fancy, you could return defaultdict(lambda: 0) to streamline the folder code.

Hey, that's a python stdlib piece I didn't know of, I'm going to use it immediately :)

src/execution/mod.rs
Psykopear: No problems, rebasing it on top of the new code made me understand some of the things going on in that PR, so I think that's good.

src/window/mod.rs
Psykopear: Yep, fixing it.

davidselassie (approved):
This looks great! Thanks for putting up with all the API changes and me being slow to respond during this process. Feel free to merge whenever you'd like. We probably won't cut a new release with this until we get recovery working properly again, but this is good to have.

whoahbot (approved):

Psykopear:
Great, I'll wait for the CI to pass after the rebase and merge it

PR #96, opened 2022-07-29
Windowing

davidselassie:
I am incapable of making a PR with less than 1500 lines of code.

Docstrings explain what is going on. Start with `bytewax.window` for
an overview of windowing concepts. Then read the Rust code in the
`window` module for how that's implemented.

There are a few changes to the recovery system you can see in diffs to
the Rust `recovery` module. The main thing is now we mostly are
passing around backups of `Vec<u8>` rather than explicit types so we
can mush all the backup types together.

The most horrible part of all the code is the renaming / re-invention
of the `StatefulMap` operator into `StatefulUnary` which very very
carefully lets you run arbitrary logic on incoming values, keeps track
of state per key, lets you "awaken" the operator per-key even if
there's no data, and in all that manages to send out state snapshots.

I disabled markdown tests until we figure out what we want for input
and how much we want to expose epochs, then we can turn them back
on. There's one pytest for windowing now, though.

The `KafkaPump` has been slightly modified to emit epochs periodically
and not deal with batch sizes, since we're moving in the direction of
heartbeat epochs.

### DIFF
--- a/pysrc/bytewax/window.py
+++ b/pysrc/bytewax/window.py
@@ -30,7 +30,7 @@ stateful operators, so you can read more on their methods on
 `bytewax.Dataflow`.
 
 4. Pass both these configs to the windowing operator of your
-choice. The **windowing operators** decided what kind of logic you
+choice. The **windowing operators** decide what kind of logic you
 should apply to values within a window and what should be the output
 of the window. E.g. `bytewax.Dataflow.reduce_window` combines all
 values in a window into a single output and sends that downstream.
--- a/pysrc/bytewax/window.py
+++ b/pysrc/bytewax/window.py
@@ -1,0 +1,75 @@
+"""Time-based windows.
+
+Bytewax provides some operators and pre-built configurations for
+easily grouping data into buckets called **windows** and running code
+on just the values in those windows.
+
+See the operator methods on `bytewax.Dataflow` with `_window` in the
+name for simple example use cases of each.
+
+Use
+---
+
+1. Pick a clock and create a config for it. A **clock** determines the
+time of each element and the current time used for closing each
+window. E.g. use the current system time. See the docs for each
+subclass of `ClockConfig` for options.
+
+2. Pick a windower and create a config for it. A **windower** defines
+how to take the values and their times and bucket them into
+windows. E.g. have tumbling windows every 30 seconds. See the docs for
+each subclass of `WindowConfig` for options.
+
+3. Pick a **key** to route the values for the window and make sure the
+input to the windowing operator you choose is a 2-tuple of `(key: str,
+value)`. Windows are managed independently for each key. If you need
+all data to be processed into the same window state, you can use a
+constant key like `("ALL", value)` but this will reduce the
+parallelism possible in the dataflow. This is similar to all the other
+stateful operators, so you can read more on their methods on
+`bytewax.Dataflow`.
+
+4. Pass both these configs to the windowing operator of your
+choice. The **windowing operators** decided what kind of logic you
+should apply to values within a window and what should be the output
+of the window. E.g. `bytewax.Dataflow.reduce_window` combines all
+values in a window into a single output and sends that downstream.
+
+You are allowed and encouraged to have as many different clocks and
+windowers as you need in a single dataflow. Just instantiate more of
+them and pass the ones you need for each situation to each windowing
+operator.
+
+Order
+-----
+
+Because Bytewax can be run as a distributed system with multiple
+worker processes and threads all reading relevant data simultaneously,
+you have to specifically collect and manually sort data that you need
+to process in strict time order.
+
+Recovery
+--------
+
+Bytewax's windowing system is built on top of its recovery system (see
+`bytewax.recovery` for more info), so failure in the middle of a
+window will be handled as gracefully as possible.
+
+Some clocks don't have a single correct answer on what to do during
+recovery. E.g. if you use `SystemClockConfig` with 10 minute windows,
+but then recover on a 15 minute mark, the system will immediately
+close out the half-completed window stored during recovery. See the
+docs for each `ClockConfig` subclass for specific notes on recovery.
+
+Recovery happens on the granularity of the _epochs_ of the dataflow,
+not the windows. See `bytewax.inputs` for more information on ways to
+adjust epochs.
+
+"""
+
+from .bytewax import (  # noqa: F401
+    ClockConfig,
+    SystemClockConfig,
+    TumblingWindowConfig,
+    WindowConfig,
+)
--- a/pysrc/bytewax/window.py
+++ b/pysrc/bytewax/window.py
@@ -70,6 +70,7 @@ adjust epochs.
 from .bytewax import (  # noqa: F401
     ClockConfig,
     SystemClockConfig,
+    TestingClockConfig,
     TumblingWindowConfig,
     WindowConfig,
 )
--- a/src/dataflow/mod.rs
+++ b/src/dataflow/mod.rs
@@ -1,4 +1,6 @@
 use crate::pyo3_extensions::TdPyCallable;
+use crate::window::ClockConfig;
+use crate::window::WindowConfig;
 use pyo3::exceptions::PyValueError;
 use pyo3::prelude::*;
 use pyo3::types::*;
@@ -18,6 +20,24 @@ use std::fmt::Display;
 
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
 pub(crate) struct StepId(String);
+
+impl<'source> FromPyObject<'source> for StepId {
+    fn extract(ob: &'source PyAny) -> PyResult<Self> {
+        Ok(StepId(ob.extract()?))
+    }
+}
+
+impl IntoPy<PyObject> for StepId {
+    fn into_py(self, py: Python) -> Py<PyAny> {
+        PyString::new(py, &self.0).into()
+    }
+}
+
+impl ToPyObject for StepId {
+    fn to_object(&self, py: Python) -> PyObject {
+        self.0.to_object(py)
+    }
+}
 
 impl Display for StepId {
     fn fmt(&self, fmt: &mut std::fmt::Formatter) -> Result<(), std::fmt::Error> {
@@ -251,11 +271,8 @@ impl Dataflow {
     ///
     /// It is a stateful operator. It requires the the input stream
     /// has items that are `(key: str, value)` tuples so we can ensure
-    /// that all relevant values are routed to the relevant
-    /// accumulator.
-    ///
-    /// It is a recoverable operator. It requires a step ID to recover
-    /// the correct state.
+    /// that all relevant values are routed to the relevant state. It
+    /// also requires a step ID to recover the correct state.
     ///
     /// It calls two functions:
     ///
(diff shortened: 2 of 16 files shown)


src/operators/mod.rs
```diff
@@ -279,173 +326,234 @@ where
...
+
+                        incoming_epoch_to_key_values_buffer
+                            .entry(epoch.clone())
+                            .or_insert_with(Vec::new)
+                            .extend(tmp_key_values.drain(..));
+
+                    });
+
+                    // TODO: Is this the right way to get the epoch
+                    // from a frontier? I haven't seen any examples
```
whoahbot, line 394: I also looked, and have not found any examples :(

src/operators/mod.rs
```diff
@@ -279,173 +326,234 @@ where
...
+                    // apply input in those epochs yet.
+                    activate_epochs_buffer.extend(incoming_epoch_to_backups_buffer.keys().cloned().filter(is_closed));
+                    activate_epochs_buffer.extend(incoming_epoch_to_key_values_buffer.keys().cloned().filter(is_closed));
+                    // Even if we don't have input data from an epoch,
+                    // we might have some output data from the
+                    // previous activation we need to output and the
+                    // frontier might have already progressed and it
+                    // would be stranded.
+                    activate_epochs_buffer.extend(outgoing_epoch_to_key_updates_buffer.keys().cloned().filter(is_closed));
+                    // Eagarly execute the frontier.
```
whoahbot, line 417:
```suggestion
                    // Eagerly execute the frontier.
```

src/window/mod.rs
```diff
@@ -0,0 +1,601 @@
...
+}
+
+pub(crate) fn build_clock<V>(
+    py: Python,
+    clock_config: Py<ClockConfig>,
+) -> Result<Box<dyn Clock<V>>, String> {
+    let window_config = clock_config.as_ref(py);
+
+    if let Ok(system_clock_config) = window_config.downcast::<PyCell<SystemClockConfig>>() {
+        let _system_clock_config = system_clock_config.borrow();
```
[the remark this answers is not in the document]
davidselassie: If we ever needed to pull config parameters out of the config object, yes. But there are no configuration parameters currently, so kinda no. I wanted to leave this no-op in there so it's a reminder that you need to do this for new config types. See https://github.com/bytewax/bytewax/blob/772c51f7c7cca09321aa86cecb34768148bd6b13/src/recovery/mod.rs#L672-L735 for the pattern I'm copying. It just looks goofy here because we don't have any other config types.

src/operators/mod.rs
```diff
@@ -279,173 +326,234 @@ where
...
-                            });
-                            state_backup_session.give_iterator(backups);
-                        }
-                    },
-                );
+                    // Schedule an activation at the next requested
+                    // wake up time.
+                    let now = Instant::now();
+                    if let Some(delay) = key_to_next_activate_at_buffer.values().map(|a| a.duration_since(now)).min() {
+                        activator.activate_after(delay);
```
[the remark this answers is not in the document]
davidselassie: > When we call `activate_after`, are we still only able to activate once `step` is called again?

I'm not sure if an operator can be activated multiple times on a single `step` call if the activation timeout passes while doing other work. Let me look into that. Although in our "source operators are inputs" future, I think that sidesteps this issue.

I think the main thing that could interact here is if we have an input source that is like "read data from Kafka but timeout after 1 sec". So then you could have the case where you read a ton of data but now the input source is empty so you get stuck waiting for new data from the input source when you actually have work to do downstream and that delays data entering a windowing operator resulting in windowing artifacts (like maybe random empty windows). I think the way around this is to design our input to never have blocking timeouts; if there's no useful work to do, yield back to Timely with a "wake me up later". This is a bummer in that we can't tap into a full async framework for this, but we'll see how bad it is to just have a delay wait. Also, if you're using event time, this isn't a problem.

> Is there a minimum time we should allow for windowing, based on the "heartbeat" of epochs flowing through the dataflow?

Windowing interacts with our "step while the probe is before the current input epoch" by introducing the input-to-probe processing latency as a delay between each epoch: e.g. when the input ticks over to epoch 5, the input is stalled until we see the last items of epoch 4 exit the output. That latency is not directly related to the heartbeat epoch length. But it would introduce some artifacts in windowing around that timescale if you're using a system clock to window. Again, if you were using event time, you wouldn't get those artifacts, though. And since this code is generic over clock type, I don't know we should enforce that here. Also since input-to-probe latency is an emergent property, it'd be hard to check this.

pysrc/bytewax/window.py
```diff
@@ -0,0 +1,75 @@
...
+input to the windowing operator you choose is a 2-tuple of `(key: str,
+value)`. Windows are managed independently for each key. If you need
+all data to be processed into the same window state, you can use a
+constant key like `("ALL", value)` but this will reduce the
+parallelism possible in the dataflow. This is similar to all the other
+stateful operators, so you can read more on their methods on
+`bytewax.Dataflow`.
+
+4. Pass both these configs to the windowing operator of your
+choice. The **windowing operators** decided what kind of logic you
```
blakestier, line 33:
```suggestion
choice. The **windowing operators** decide what kind of logic you
```

src/dataflow/mod.rs
```diff
@@ -313,38 +330,41 @@ impl Dataflow {
     ///     is_complete: `is_complete(updated_accumulator: Any) =>
     ///         should_emit: bool`
     #[pyo3(text_signature = "(self, step_id, reducer, is_complete)")]
-    fn reduce(&mut self, step_id: String, reducer: TdPyCallable, is_complete: TdPyCallable) {
+    fn reduce(&mut self, step_id: StepId, reducer: TdPyCallable, is_complete: TdPyCallable) {
```
blakestier, line 333: 👍

src/dataflow/mod.rs
```diff
@@ -313,38 +330,41 @@ impl Dataflow {
...
             reducer,
             is_complete,
         });
     }
 
-    /// Reduce epoch lets you combine all items for a key within an
-    /// epoch into an accumulator.
+    // TODO: Update this example once we have a better idea about
+    // input.
+    /// Reduce window lets you combine all items for a key within an
```
blakestier, line 343:
```suggestion
    /// Reduce window lets you combine all items for a key within a
```

src/inputs/mod.rs
```diff
@@ -100,12 +102,11 @@ pub(crate) struct InputConfig;
...
 /// Args:
 ///
-///     brokers: Comma-separated of broker addresses.
-///         E.g. "localhost:9092,localhost:9093"
+///     hosts: Broker addresses. E.g. "localhost:9092,localhost:9093"
 ///
 ///     group_id: Group id as a string.
 ///
-///     topic: Topic to which consumer will subscribe.
+///     topics: Topics to which consumer will subscribe.
```
blakestier, line 109: 👍

src/inputs/mod.rs
```diff
@@ -116,100 +117,134 @@ pub(crate) struct InputConfig;
...
 ///     your execution entry point.
 #[pyclass(module = "bytewax.inputs", extends = InputConfig)]
-#[pyo3(
-    text_signature = "(brokers, group_id, topics, offset_reset, auto_commit, messages_per_epoch)"
-)]
+#[pyo3(text_signature = "(hosts, group_id, topics, tail, offset_reset, auto_commit, epoch_length)")]
 pub(crate) struct KafkaInputConfig {
     #[pyo3(get)]
-    pub brokers: String,
+    pub hosts: Vec<String>,
```
blakestier, line 130: Why the rename? I favored `brokers` since that's what Kafka and `rdkafka` use. NBD just curious.

src/operators/mod.rs
```diff
@@ -89,138 +100,142 @@ pub(crate) fn inspect_epoch(inspector: &TdPyCallable, epoch: &u64, item: &TdPyAn
...
+        (acc, None, None)
+    }
 }
 
 /// Extension trait for [`Stream`].
 // Based on the good work in
 // https://github.com/TimelyDataflow/timely-dataflow/blob/0d0d84885672d6369a78cd9aff7beb2048390d3b/timely/src/dataflow/operators/aggregation/state_machine.rs#L57
-pub(crate) trait StatefulMap<S: Scope, V: ExchangeData> {
-    /// See [`crate::dataflow::Dataflow::stateful_map`] for semantics.
+pub(crate) trait StatefulUnary<S: Scope, V: ExchangeData> {
```
blakestier, line 152: Yay for rename

src/operators/mod.rs
```diff
@@ -89,138 +100,142 @@ pub(crate) fn inspect_epoch(inspector: &TdPyCallable, epoch: &u64, item: &TdPyAn
...
+
+            let should_emit_and_discard_acc: bool = with_traceback!(py, {
+                let should_emit_and_discard_acc_pybool: TdPyAny =
+                    is_complete.call1(py, (updated_acc.clone_ref(py),))?.into();
+                should_emit_and_discard_acc_pybool.extract(py).map_err(|_err| PyTypeError::new_err(format!("return value of `is_complete` in reduce operator must be a bool; got `{should_emit_and_discard_acc_pybool:?}` instead")))
+            });
+            debug!("reduce for key={key:?}: is_complete={is_complete:?}(updated_acc={updated_acc:?}) -> should_emit_and_discard_acc={should_emit_and_discard_acc:?}");
+
+            if should_emit_and_discard_acc {
+                (None, Some(updated_acc), None)
```
blakestier, line 139: I think it would probably be awkward to wrap this in another type but it's tricky to read the 3 tuple of options 🤔

src/operators/mod.rs
```diff
@@ -279,173 +326,234 @@ where
...
+                                    // signal that all keys have
+                                    // terminated.
+                                    keys_buffer.extend(key_values_buffer.iter().map(|(k, _v)| k).cloned());
+                                    keys_buffer.extend(state_cache.keys().cloned());
+                                    keys_buffer.dedup();
+                                    // Drain to re-use allocation.
+                                    key_values_buffer.extend(keys_buffer.drain(..).map(|k| (k.clone(), Poll::Ready(None))));
+                                } else {
+                                    // Otherwise, wake up any keys
+                                    // that are passed their requested
```
blakestier, line 476:
```suggestion
                                    // that are past their requested
```
?

src/window/mod.rs
```diff
@@ -0,0 +1,601 @@
...
+            )))
+        }
+    }
+}
+
+pub(crate) fn build_clock<V>(
+    py: Python,
+    clock_config: Py<ClockConfig>,
+) -> Result<Box<dyn Clock<V>>, String> {
+    let window_config = clock_config.as_ref(py);
```
blakestier, line 96: Should this var be `clock_config`? I'm not seeing where it's used for windowing here (yet)

whoahbot (approved):

src/operators/mod.rs
davidselassie: Not a bad idea. It's a little verbose, but we can try it out.

src/inputs/mod.rs
davidselassie: I had copy-pasted the Kafka code I used for recovery that didn't have async, and that happen to use `hosts` but `brokers` seems more specific. I'll change them both to that.

blakestier (approved):

PR #98, opened 2022-08-03
Input source operators

davidselassie:
Gets rid of `Pump`s and instead makes custom Timely source operators.

The source operator types control how often epochs happen and we're
defaulting to an epoch every 10 seconds.

The input components are just iterators and know nothing about epochs.

There are now output configs / components that you can use to
customize where capture data is being written.

Gets rid of `run` and `run_cluster`. Move the other execution entry
points into `bytewax.execution` only. `Dataflow` is moved into
`bytewax.dataflow.Dataflow`.

TODO
----

I haven't updated any of the doctests or examples. I think that'll be
a lot of work and want to wait until we've stabilized anything else
that needs to change here.

Cleaner docs on this. Specifically updating the long-form docs. We can
get rid of all the stuff teaching you about epochs.

Recovery does not currently recover from the failure epoch because we
don't have the offset system figured out, and the fact the input
system is isolated from epochs means the old "resume epoch" thing
doesn't make sense now.


src/dataflow/mod.rs
```diff
@@ -266,6 +285,38 @@ impl Dataflow {
...
+    ///
+    /// It calls a **mapper** function on each item.
+    ///
+    /// It emits each updated item downstream.
+    ///
+    /// It is commonly used for:
+    ///
+    /// - Extracting keys
+    /// - Turning JSON into objects
+    /// - So many things
```
blakestier, line 298: 😂

src/execution/mod.rs
```diff
@@ -441,33 +691,26 @@ fn worker_main<A: Allocate>(
...
 /// Args:
 ///
 ///     flow: Dataflow to run.
 ///
-///     input_config: Input config of type Manual or Kafka. See `bytewax.inputs`.
-///
-///     output_builder: Returns a callback function for each worker
-///         thread, called with `(epoch, item)` whenever and item
-///         passes by a capture operator on this process.
+///     epoch_config: A custom epoch config. You probably don't need
```
blakestier, line 701: This still feels funny to me, but i know we still have a lot of moving parts and maybe . . . maybe it just has to be funny, at least for now, but maybe eventually we'll be able to extract some of the testing mechanisms.

This is a 💭 only

src/execution/mod.rs
```diff
@@ -441,33 +691,26 @@ fn worker_main<A: Allocate>(
...
 ///         persisted.
 ///
-#[pyfunction(flow, input_config, output_builder, "*", recovery_config = "None")]
-#[pyo3(text_signature = "(flow, input_config, output_builder, *, recovery_config)")]
+#[pyfunction(flow, "*", epoch_length = "None", recovery_config = "None")]
+#[pyo3(text_signature = "(flow, *, epoch_config, recovery_config)")]
 pub(crate) fn run_main(
     py: Python,
     flow: Py<Dataflow>,
-    input_config: Py<InputConfig>,
```
blakestier, line 469: This def feels like the right direction for the API to me

src/inputs/mod.rs
```diff
@@ -256,90 +236,106 @@ impl KafkaInputConfig {
...
-///     your execution entry point.
-#[pyclass(module = "bytewax.inputs", extends = InputConfig)]
-#[pyo3(text_signature = "(input_builder)")]
-pub(crate) struct ManualInputConfig {
-    input_builder: Py<PyAny>,
+/// Defines out a single source of input reads data.
+pub(crate) trait InputReader<D> {
+    /// This has the same semantics as
+    /// [`std::async_iter::AsyncIterator::poll_next`]: no values ready
+    /// qyet ([`Poll::Pending`]), a new value has arrived
```
blakestier, line 243:
```suggestion
    /// yet ([`Poll::Pending`]), a new value has arrived
```

src/inputs/mod.rs
```diff
@@ -256,90 +236,106 @@ impl KafkaInputConfig {
...
+    /// ([`Poll::Ready`] with a [`Some`]), or the stream has ended and
+    /// there will be no more output ([`Poll::Ready`] with a
+    /// [`None`]).
```
blakestier, line 246: This comment is kind of awkward to read-- maybe pull each onto its own line?

src/execution/mod.rs
davidselassie: Yeah. I'm not sure what to do here because this is a subtle part of the API. Another idea might be to have this take the `ClockConfig` instead? Then you could use the `TestingClockConfig` to programmatically set the time in tests. But I don't think it could be exactly the same clock as the windowing because the idea of watermarks doesn't apply here. So then we'd be adding in a special new clock which seems actually more confusing than an `EpochConfig`. Perhaps we can streamline this later, but I do think the input-epoch separation is a step forward.

src/pyo3_extensions/mod.rs
```diff
@@ -309,20 +289,13 @@ pub(crate) fn wrap_state_pair(key_value: (StateKey, TdPyAny)) -> TdPyAny {
 ///
 /// Otherwise the GIL would be held for the entire life of the iterator.
 #[derive(Clone)]
-pub(crate) struct TdPyIterator(pub(crate) Py<PyIterator>);
```
blakestier, line 312: 🤔

src/pyo3_extensions/mod.rs
```diff
@@ -375,6 +348,12 @@ impl ToPyObject for TdPyCallable {
 ///
 /// Just pass through to [`Py`].
 impl TdPyCallable {
+    /// Create an "empty" [`Self`] just for use in `__getnewargs__`.
```
blakestier, line 351: 😎

src/execution/mod.rs
blakestier: Agree agree

blakestier (approved):
Looks great. This direction feels a lot cleaner and easier to reason about imo.

src/inputs/mod.rs
```diff
@@ -256,90 +236,109 @@ impl KafkaInputConfig {
...
-///
-/// Returns:
-///
-///     Config object. Pass this as the `input_config` argument to
-///     your execution entry point.
-#[pyclass(module = "bytewax.inputs", extends = InputConfig)]
-#[pyo3(text_signature = "(input_builder)")]
-pub(crate) struct ManualInputConfig {
-    input_builder: Py<PyAny>,
+/// Defines out a single source of input reads data.
```
blakestier, line 239:
```suggestion
/// Defines how a single source of input reads data.
```

whoahbot (approved):

PR #99, opened 2022-08-04
Cooperative manual input

davidselassie:
Asking an `InputReader` for the next piece of data should never
block. That's why it returns a `Poll<Option<T>`. But we need to expose
this Python-side if we want people to make well-behaved custom input
sources. This change allows that by intepreting any `None` in the
input stream as "no data, just yield and let other work happen".

You can see an example of how this would work in the unit test.


whoahbot (approved):
Lgtm

PR #103, opened 2022-08-16
Stateful input sources

davidselassie:
- Moves recovery state loading into its own separate dataflow that
  runs before the production one.

- Refactors stateful operators to use `StatefulLogic` impls that
  modify some internal state and periodic snapshotting. Same deal for
  windowing operators and `WindowLogic`. Both of these use the "resume
  state builder" pattern to re-create that internal state from a
  resume snapshot.

- Input is now specified on the `Dataflow.input` operator.

- Input is now stateful: Manual `input_builder`s accept old state and
  must yield a 2-tuple of `(state, item)`. Kafka input automatically
  manages its own state. This means recovery now resumes from the
  correct input location.

- Fixes up most? docstrings / doctests / Rustdocs. I didn't run the
  doctests because they're mostly in Rustdoc comments which doesn't
  work yet.

- Moves around operators to better separate concerns: code to
  implement Bytewax operators lives in `crate::operators`, utility
  Timely operators live in the crate that needs them (e.g. the
  recovery operators are moved into `crate::recovery`, like
  `StatefulUnary`).

TODO
----

In no particular order; some are aspirational.

- [x] Update long form documentation.

- [x] Fix up rest of examples.

- [x] Do we run our Rust tests in CI? We have some and we should.

- [x] Merge `fold_window` operator.

- [ ] More tests: Rust unit tests for `TumblingWindower`. Pytests which
  test that stateful input recovery is working (Like use Sqlite as the
  recovery store, but read from Kafka. Have a test setup write some
  data to a topic.). Add in testing the Kafka recovery store (can we
  get a small Kafka cluster in CI? Skip tests if it can't connect?).

- [ ] Break apart our huge Rust modules more. I think we should have a
  submodule per-config: e.g. `crate::recovery::stores::{sqlite,
  kafka}`.


pytests/test_operators.py
```diff
@@ -101,9 +114,21 @@ def session_complete(session):
         {"user": "a", "type": "logout"},
         {"user": "b", "type": "logout"},
     ]
-    flow = Dataflow(TestingInputConfig(inp))
+    flow.input("inp", TestingInputConfig(inp))
```
blakestier, line 117: 🤔 Time to make a `flow.output()` instead of a `capture`?

blakestier: (not in this pr)

src/execution/mod.rs
```diff
@@ -278,29 +338,53 @@ where
...
+
+    (resume_state_bytes, key)
 }
 
 fn build_input_source<S>(
     py: Python,
     epoch_config: Py<EpochConfig>,
     scope: &S,
+    step_id: StepId,
+    key: StateKey,
```
blakestier, line 377: I'm interested in calling this `state_key`

src/dataflow/mod.rs
```diff
@@ -2,135 +2,48 @@
...
-use sqlx::Type;
-use std::borrow::Cow;
-use std::fmt::Display;
-
-/// Newtype representing the unique name of a step in a dataflow.
-///
-/// Used as a key for looking up relevant state for different steps
-/// surrounding recovery.
-#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
-pub(crate) struct StepId(String);
```
davidselassie, line 33: Moved into `crate::recovery` since the reason it exists is to facilitate recovery. And then all the SQLx stuff can be isolated in there.

src/execution/mod.rs
davidselassie: I'd like to refactor this later to more of the builder pattern so each source is in control of the key fully round-tripped through the recovery store and we don't actually have to pass it in here.

src/execution/mod.rs
```diff
@@ -328,63 +450,6 @@ where
     }
 }
 
-/// Compile a dataflow which reads the progress data from the previous
-/// execution and calculates the resume epoch.
-fn build_resume_epoch_calc_dataflow<A: Allocate>(
```
davidselassie, line 333: Moved into `crate::recovery` so that all the reading and writing logic for state is in one spot.

src/operators/mod.rs
```diff
@@ -100,896 +71,225 @@ pub(crate) fn inspect_epoch(inspector: &TdPyCallable, epoch: &u64, item: &TdPyAn
...
-pub(crate) struct StatefulUnaryLogicReturn<R, D, I: IntoIterator<Item = R>> {
-    pub(crate) updated_state: Option<D>,
-    pub(crate) output: I,
-    pub(crate) activate_after: Option<Duration>,
-}
-
-/// Extension trait for [`Stream`].
-// Based on the good work in
-// https://github.com/TimelyDataflow/timely-dataflow/blob/0d0d84885672d6369a78cd9aff7beb2048390d3b/timely/src/dataflow/operators/aggregation/state_machine.rs#L57
-pub(crate) trait StatefulUnary<S: Scope, V: ExchangeData> {
```
davidselassie, line 170: All of the Timely operators just used for recovery I moved into `crate::recovery`. I think we'll just keep Bytewax operator definitions in this module.

src/pyo3_extensions/mod.rs
```diff
@@ -196,84 +183,19 @@ impl PartialEq for TdPyAny {
...
-///
-/// Yes, we lose a little bit of flexibility, but it makes usage more
-/// convenient.
-///
-/// You'll mostly interface with this via [`extract_state_pair`] and
-/// [`wrap_state_pair`].
-#[derive(
-    Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Deserialize, FromPyObject,
-)]
-pub(crate) struct StateKey(String);
```
davidselassie, line 216: Moved into `crate::recovery` since it only exists to support the recovery data model. Then we can keep all the SQLx stuff isolated in there too.

davidselassie:
I guess CI won't run if there's merge conflicts? Tests are passing locally.

src/inputs/mod.rs
```diff
@@ -2,17 +2,40 @@
...
+//!
+//! Architecture
+//! ------------
+//!
+//! The one extra quirk here is that input is completely decoupled
+//! from epoch generation. See [`crate::execution`] for how Timely
+//! sources are generated and epochs assigned. The only goal of the
+//! input system is "what's my next item for this worker?"
+//!
+//! Output is based around the core trait of [`InputReader`].  The
```
blakestier, line 14:
```suggestion
//! Input is based around the core trait of [`InputReader`].  The
```

src/recovery/mod.rs
```diff
@@ -9,42 +9,57 @@
...
+//! --------------
 //!
-//! Recovery is based around storing each worker's state updates and
-//! progress information. Each worker's recovery data is stored
-//! separately so that we can discern the exact state of each worker
-//! at the point of failure.
+//! To create a new stateful operator, create a new
+//! [`StatefulLogic`] impl and pass it to the
+//! [`StatefulUnary`] Timely operator. If you
+//! fullfil the API of [`StatefulLogic`], you will
```
blakestier, line 18:
```suggestion
//! fulfill the API of [`StatefulLogic`], you will
```

src/recovery/mod.rs
```diff
@@ -9,42 +9,57 @@
...
+//! get proper recovery behavior.
 //!
-//! The **worker frontier** represents the oldest epoch for which
-//! items are still in the process of being output or state updates
-//! are being backed up.
+//! The general idea is that you pass a **logic builder** which takes
+//! any previous state snapshots from the last exectuion and builds an
```
blakestier, line 22:
```suggestion
//! any previous state snapshots from the last execution and builds an
```

src/window/mod.rs
```diff
@@ -116,25 +124,42 @@ impl TestingClockConfig {
     const __test__: bool = false;
 
     #[new]
-    #[args(item_incr)]
-    fn new(item_incr: pyo3_chrono::Duration) -> (Self, ClockConfig) {
-        (Self { item_incr }, ClockConfig {})
+    #[args(item_incr, start)]
```
blakestier, line 127: start_at

src/window/mod.rs
```diff
@@ -96,6 +99,9 @@ impl ClockConfig {
 ///     item_incr (datetime.timedelta): Amount to increment "now"
 ///         after each item.
 ///
+///     start_at (datetime.datetime): Initial "now" / time of first
+///         item. If you set this and use a window config
```
blakestier, line 103: fix this doc

src/window/mod.rs
```diff
@@ -594,6 +633,179 @@ pub(crate) enum WindowError<V> {
     Late(V),
 }
 
+/// Impl this trait to create a windowing operator.
+///
+/// A separate instance of this will be create for each window a
```
blakestier, line 638:
```suggestion
/// A separate instance of this will be created for each window a
```

docs/examples/cart-join.md
```diff
@@ -118,42 +117,43 @@ def format_output(user_id__joined_state):
...
 On that note, we'll serialize the final output as JSON and print it.
 
 ```python
 def output_builder(worker_index, worker_count):
-    def output_handler(epoch_item):
-        epoch, item = epoch_item
+    def output_handler(item):
         line = json.dumps(item)
-        print(epoch, line)
+        print(line)
```
blakestier, line 127: Maybe two captures in the future 🤔 Seems odd that we're not employing the `StdOut`

docs/examples/cart-join.md
```diff
@@ -175,21 +175,28 @@ Following our checklist in [Bytewax's documentation on
 recovery](/getting-started/recovery/) we need to enhance our input
 builder with a few things.
 
-First, we need the ability to resume from an arbitrary epoch. Since
-we're using the line in the file as the epoch, we can skip ahead to
-that line.
+First, we need the ability to resume our input from where we left off.
+When using the `ManualInputBuilder`, we emit a state object, that we
```
blakestier, line 179: Config?

docs/examples/search-session.md
```diff
@@ -61,46 +61,34 @@ Epochs
...
-order in your data.
-
-Since we are going to use the order in which events happen to divvy
-them up into user sessions, we need to take advantage of epochs at
-some level. In this basic example, we're going to label each input
-event with a monotonically increasing epoch. [There are some some
-scaling downsides to this approach](/getting-started/epochs/), but we will be
-using it here anyway to demonstrate the basics of epochs.
+def input_builder(worker_index, worker_count, resume_state):
+    state = resume_state or {}
```
blakestier, line 85: @whoahbot if you were actually using a state passed in, would this work?

docs/examples/cart-join.md
whoahbot: I wasn't able to get the doctests passing when using `StdOut`. I think it's something we can fix, but I didn't get a chance to dig into why it wasn't working.

docs/examples/cart-join.md
whoahbot: Not sure I understand what you are referring to here. Can you clarify?

docs/examples/search-session.md
whoahbot: If I understand the intent of your question, I think you are suggesting that using resume_state this way could be confusing.

Perhaps using something like:

``` python
def input_builder(worker_index, worker_count, resume_state):
    for line in IMAGINE_THESE_EVENTS_STREAM_FROM_CLIENTS:
        yield(resume_state, line)
```

Would be better. What do you think?

docs/examples/cart-join.md
blakestier: I think you want `ManualInputConfig` over `ManualInputBuilder`

docs/examples/search-session.md
blakestier: I prefer the `None` scenario so we don't have to think hard about the shape of a `state` that we're ignoring, though in general this concern points more to @davidselassie's desire to move away from this shape in general.

I'll update it to `None` for now but we can revisit when you're back.

docs/examples/cart-join.md
blakestier: I think we have to capture the stdout (like in the tests) for it to run successfully. that's arguably more awkward so I'm fine leaving it. We can use `StdOutputConfig` in the py examples still

src/inputs/mod.rs
```diff
@@ -129,114 +150,86 @@ impl ManualInputConfig {
...
-///     `bytewax.dataflow.Dataflow` constructor.
+///     `bytewax.dataflow.Dataflow.input`.
 #[pyclass(module = "bytewax.inputs", extends = InputConfig)]
-#[pyo3(text_signature = "(brokers, group_id, topics, tail, offset_reset, auto_commit)")]
+#[pyo3(text_signature = "(brokers, topic, tail, starting_offset)")]
 pub(crate) struct KafkaInputConfig {
     #[pyo3(get)]
     brokers: Vec<String>,
     #[pyo3(get)]
-    group_id: String,
```
Psykopear, line 161: Is there a reason why `group_id` should not be configurable anymore?

whoahbot (approved):

PR #109, opened 2022-08-25
Solved some clippy warnings, enabled doctests

Psykopear:
In this PR I solved some of the remaining clippy issues, and added the "rlib" crate type to the Cargo.toml so that doctests are ran when running `cargo test`.

The clippy issues where of four kinds:
1) Too complex types
2) Some uses of the `with_traceback!` macro where defining a function and immediately calling it without any advantage in doing so
3) Functions with too many arguments
4) PyO3 related warnings (deref on an immutable reference)

## Too complex types
To solve most of those issues, I created 2 type aliases:
- `StringResult<T>` which is just `Result<T, String>`
- `EpochData` which represents a tuple of `(StateKey, (StepId, StateUpdate))`

I'm not sure `EpochData` is a proper name though, so let me know if you can think of something better.
Some of the warnings remain, but I'd like some help on the naming there, because the intent is not always clear to me.

## with_traceback! macro
Spoiler: I wanted to remove the macro, I ended up writing 2 more...
The intention of the macros looks clearer to me this way, but if you don't like adding more macros than we already had I can revert these changes and either ignore the clippy issues or find a way to make it all work with a single macro.

The macro definition was quite simple:
```rust
macro_rules! with_traceback {
    ($py:expr, $pyfunc:expr) => {
        match (|| $pyfunc)() {
            Ok(r) => r,
            Err(err) => std::panic::panic_any(err),
        }
    };
}
```

Clippy was warning that sometimes defining the function and running it was not needed, so it suggested something like this:
```rust
macro_rules! with_traceback {
    ($py:expr, $pyfunc:expr) => {
        match  $pyfunc {
            Ok(r) => r,
            Err(err) => std::panic::panic_any(err),
        }
    };
}
```
The warning made sense, because we sometimes called the macro with just a simple expression inside.
The macro did not generate the warning when we used the `?` operator inside the expression (emulating a `try` block).
Also a lot of times when the macro was used, there was a mapping of the error to a `PyTypeError` at the calling site.

So I split what that macro was doing into 3 macros, documented and with doctests included.
Now we have:
```rust
macro_rules! unwrap_any {
    ($pyfunc:expr) => {
        $pyfunc.unwrap_or_else(|err| std::panic::panic_any(err))
    };
}
```
```rust
macro_rules! try_unwrap {
    ($pyfunc:expr) => {
        (|| $pyfunc)().unwrap_or_else(|err| std::panic::panic_any(err))
    };
}
```
```rust
macro_rules! py_unwrap {
    ($pyfunc:expr, $err_msg:expr) => {
        $pyfunc
            .map_err(|_err| PyTypeError::new_err($err_msg))
            .unwrap_or_else(|err| std::panic::panic_any(err))
    };
}
```
See the file `src/macros.rs` for documentation, examples and reference

## Functions with too many arguments
I didn't touch anything here, this would require to better define the types we move around and probably a non trivial refactoring, but let me know if you want me to work on that too.

## PyO3 related warnings
Those are already fixed in PyO3 0.17, but we need to wait until `pyo3-chrono` and `pyo3-log` release a new version based on PyO3 0.17 before upgrading.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,7 @@ edition = "2021"
 
 [lib]
 name = "bytewax"
-crate-type = ["cdylib"]
+crate-type = ["cdylib", "rlib"]
 
 [package.metadata.maturin]
 python-source = "pysrc"
--- a/src/execution/mod.rs
+++ b/src/execution/mod.rs
@@ -29,13 +29,13 @@
 //! for how to create a [`periodic_epoch_source`].
 
 use crate::dataflow::{Dataflow, Step};
+use crate::StringResult;
 use crate::inputs::build_input_reader;
 use crate::inputs::InputReader;
 use crate::operators::*;
 use crate::outputs::build_output_writer;
 use crate::outputs::capture;
 use crate::pyo3_extensions::{extract_state_pair, wrap_state_pair, TdPyAny};
-use crate::recovery::CollectGarbage;
 use crate::recovery::StatefulUnary;
 use crate::recovery::WriteProgress;
 use crate::recovery::WriteState;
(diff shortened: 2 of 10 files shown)


blakestier (approved):

whoahbot (approved):
Looks good to me. I think there might be a better name for `EpochData`, but I wasn't able to think of one.

Since it's an internal refactor, we can always come back and change it if we find something better.

PR #113, opened 2022-08-26
Break up large modules into smaller files.

whoahbot:
This PR is just a restructuring of our Rust code into submodules, it does not include any functional changes to the library.


Psykopear (approved):

src/execution/mod.rs
```diff
@@ -49,26 +52,29 @@ use crate::recovery::{ProgressReader, StateCollector};
 use crate::recovery::{RecoveryConfig, StepId};
 use crate::recovery::{RecoveryStoreSummary, StateWriter};
 use crate::recovery::{StateBytes, StateKey};
-use crate::recovery::{StateReader, StateUpdate};
+use crate::recovery::StateReader;
```
blakestier, line 55: I don't think this is related to your PR but i'm curious about the way we're importing things. Why bundle some imports in `{}` and then do it again? Is this a formatting bug?

src/execution/mod.rs
```diff
@@ -49,26 +52,29 @@ use crate::recovery::{ProgressReader, StateCollector};
...
+use std::time::Duration;
 use timely::communication::Allocate;
-use timely::dataflow::operators::generic::builder_rc::OperatorBuilder;
 use timely::dataflow::{operators::*, Stream};
 use timely::dataflow::{ProbeHandle, Scope};
 use timely::progress::Timestamp;
 use timely::worker::Worker;
-use timely::Data;
+
+use self::periodic_epoch::{periodic_epoch_source, PeriodicEpochConfig};
```
blakestier, line 73: 😳 what the heck.

i like that we modularize it. I hadn't seen this tho

src/inputs/manual_input.rs
```diff
@@ -0,0 +1,129 @@
+use std::task::Poll;
+
+use crate::pyo3_extensions::{TdPyAny, TdPyCallable, TdPyCoroIterator};
+use crate::recovery::StateBytes;
+use crate::{py_unwrap, try_unwrap};
+use pyo3::exceptions::{PyTypeError, PyValueError};
+use pyo3::prelude::*;
+
+use super::{InputConfig, InputReader};
```
blakestier, line 9: like drinking a cool glass of water

src/operators/mod.rs
```diff
@@ -10,17 +10,16 @@
...
-use crate::window::WindowLogic;
 use crate::try_unwrap;
 use crate::{log_func, unwrap_any};
 use log::debug;
 use pyo3::exceptions::PyTypeError;
 use pyo3::prelude::*;
-use std::task::Poll;
-use std::time::Duration;
+
+pub(crate) mod fold_window;
```
blakestier, line 19: 🤔 would it make sense for us to further compartmentalize stateful vs not stateful operators? Or too much?

blakestier (approved):
🥇 Thanks for doing this. A few questions, nothing blocking.

PR #115, opened 2022-08-30
Only use UTC datetimes

Psykopear:
This PR forces datetimes passed to bytewax as config parameters to be in UTC, using the new chrono integration added in PyO3 (not merged yet).

We'll need to wait for the PR on PyO3 to be merged so that we can open a PR on pyo3-log to make everything work.
The PyO3 author said he wants to add the chrono integration in PyO3 0.17.2 which is coming soon, so this should not take too long.

## Previous description (ignore this)
This PR makes all datetimes in Bytewax timezone aware.

It uses a non-released version of PyO3 that adds the `chrono` integration, and removes the dependency from `pyo3-chrono`, which was doing the python->rust conversion deleting timezone info, without properly converting to UTC first.

The approach taken in this PR is to add a `chrono_tz` integration on top of the one implemented in PyO3, so that any datetime in the codebase has timezone info attached, and can be correctly compared to any other datetime.

The only edge case we have to handle, is when we receive naive datetimes from the user in python.
Right now the approach is to convert the datetime to utc using the python interpreter.
This means that any naive datetime coming from python will be assumed to be local (relative to where the dataflow is ran) and converted to UTC before being passed down to rust. This could be wrong in case the datetime comes from data generated somewhere with a different timezone, but we don't have enough informations to make a proper conversion anyway. Another possible approach would be to assume the datetime is UTC rather than local, and we would attach the UTC timezone info without converting first, but this would be equally wrong most of the times. The application should warn anytime we receive a naive datetime, so that users are aware of the fact and can fix the data if needed.

For (de)serialization of datetimes, I had to implement serde's traits manually, since there is no standard way to serialize datetimes with IANA timezone info. To do this I just serialize the local naivedatetime and the timezone name separately, so that we can rebuild the same `ChronoDateTime` when deserializing.

The PR is still in Draft because I'm waiting for PyO3 to release the new version, and depending on the versioning chosen (0.17.2 or 0.18) we might also need to wait for an update on `pyo3-log` which depends on PyO3 0.17.

To make things easier, for the next release I propose to implement Consensus Time:
https://xkcd.com/2594/

## Original description (ignore this too)

The test at the end of the file should explain the situation I'm trying to fix here: `pyo3-chrono` ignores timezone info even if they are present.
I think we should instead take that into account.
With this conversion traits, we always convert datetimes to UTC internally.
If the user sends us timezone aware datetimes, we do the conversion.
If the user sends us naive datetimes, we assume they are already UTC (debatable).

This is not the only possible approach, and might not be the best one, so I'm opening this PR to discuss what you think about this.

My point is that I don't think we should ignore tzinfo if present, but the flow should still work if we receive naive datetimes (and maybe warn the user about the assumption).

This PR is rough, and if we want to replace `pyo3-chrono` I'll need to add the conversion traits for `Time` and `Date` too, and some more tests.

edit: I just found this PR: https://github.com/chronotope/chrono/pull/542 which does the conversion properly. Reading the discussion, it looks like they plan to integrate this into PyO3 (rather than in chrono) under a feature flag, so maybe we'll have this for free in the future

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,8 +15,13 @@ bincode = { version = "1.3.3" }
 chrono = { version = "0.4", features = [ "serde" ] }
 futures = { version = "0.3.21" }
 log = { version = "0.4" }
-pyo3 = { git = "https://github.com/psykopear/pyo3", branch = "chrono", version = "0.17", features = ["macros", "chrono"] }
-# pyo3-log = { version = "0.7" }
+# TODO: PyO3 pinned to current latest main commit, which includes pyo3->chrono integration
+#       Change to version "0.17.2" as soon as it is released
+pyo3 = { git = "https://github.com/pyo3/pyo3", rev = "86ce4d1a1", version = "0.17", features = ["macros", "chrono"] }
+# TODO: pyo3-log too is pinned to a fork, so that it compiles together with the unreleased PyO3 version.
+#       Change this to the version "0.7" as soon as PyO3 "0.17.2" itself is released.
+#       The current pyo3-log version should work as it is, since it depends on pyo3 "~0.17".
+pyo3-log = { git = "https://github.com/psykopear/pyo3-log", rev = "802b6f4", version = "0.7" }
 scopeguard = { version = "1.1.0" }
 send_wrapper = { version = "0.6.0" }
 serde = { version = "1.0.134" }
@@ -34,7 +39,7 @@ rdkafka = { version = "0.28.0", features = [ "cmake-build", "gssapi-vendored", "
 rdkafka = { version = "0.28.0", features = [ "cmake-build", "gssapi", "ssl" ] }
 
 [dev-dependencies]
-pyo3 = { git = "https://github.com/psykopear/pyo3", branch = "chrono", version = "0.17", default-features = false, features = ["macros", "chrono"] }
+pyo3 = { git = "https://github.com/pyo3/pyo3", rev = "86ce4d1a1", version = "0.17", default-features = false, features = ["macros", "chrono"] }
 
 [features]
 extension-module = ["pyo3/extension-module"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,9 +15,8 @@ bincode = { version = "1.3.3" }
 chrono = { version = "0.4", features = [ "serde" ] }
 futures = { version = "0.3.21" }
 log = { version = "0.4" }
-pyo3 = { version = "0.17.1" }
-pyo3-chrono = { version = "0.5.0" }
-pyo3-log = { version = "0.7.0" }
+pyo3 = { git = "https://github.com/psykopear/pyo3", branch = "chrono", version = "0.17", features = ["macros", "chrono"] }
+# pyo3-log = { version = "0.7" }
 scopeguard = { version = "1.1.0" }
 send_wrapper = { version = "0.6.0" }
 serde = { version = "1.0.134" }
@@ -35,7 +34,7 @@ rdkafka = { version = "0.28.0", features = [ "cmake-build", "gssapi-vendored", "
 rdkafka = { version = "0.28.0", features = [ "cmake-build", "gssapi", "ssl" ] }
 
 [dev-dependencies]
-pyo3 = { version = "0.17.1", default-features = false }
+pyo3 = { git = "https://github.com/psykopear/pyo3", branch = "chrono", version = "0.17", default-features = false, features = ["macros", "chrono"] }
 
 [features]
 extension-module = ["pyo3/extension-module"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,10 +17,12 @@ futures = { version = "0.3.21" }
 log = { version = "0.4" }
 # TODO: PyO3 pinned to current latest main commit, which includes pyo3->chrono integration
 #       Change to version "0.17.2" as soon as it is released
+# pyo3 = { version = "0.17.2", features = ["macros", "chrono"] }
 pyo3 = { git = "https://github.com/pyo3/pyo3", rev = "86ce4d1a1", version = "0.17", features = ["macros", "chrono"] }
 # TODO: pyo3-log too is pinned to a fork, so that it compiles together with the unreleased PyO3 version.
 #       Change this to the version "0.7" as soon as PyO3 "0.17.2" itself is released.
 #       The current pyo3-log version should work as it is, since it depends on pyo3 "~0.17".
+# pyo3-log = "0.7"
 pyo3-log = { git = "https://github.com/psykopear/pyo3-log", rev = "802b6f4", version = "0.7" }
 scopeguard = { version = "1.1.0" }
 send_wrapper = { version = "0.6.0" }
@@ -39,6 +41,9 @@ rdkafka = { version = "0.28.0", features = [ "cmake-build", "gssapi-vendored", "
 rdkafka = { version = "0.28.0", features = [ "cmake-build", "gssapi", "ssl" ] }
 
 [dev-dependencies]
+# TODO: PyO3 pinned to current latest main commit, which includes pyo3->chrono integration
+#       Change to version "0.17.2" as soon as it is released
+# pyo3 = { version = "0.17.2", default-features = false, features = ["macros", "chrono"] }
 pyo3 = { git = "https://github.com/pyo3/pyo3", rev = "86ce4d1a1", version = "0.17", default-features = false, features = ["macros", "chrono"] }
 
 [features]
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -13,12 +13,13 @@ python-source = "pysrc"
...
 chrono = { version = "0.4", features = [ "serde" ] }
-chrono-tz = { version = "0.6", features = [ "serde" ] }
+# chrono-tz = { version = "0.6", features = [ "serde" ] }
 futures = { version = "0.3.21" }
 log = { version = "0.4" }
-pyo3 = { version = "0.17.1" }
-pyo3-chrono = { version = "0.5.0" }
-pyo3-log = { version = "0.7.0" }
+# pyo3 = { version = "0.17.1", git = " }
+pyo3 = { path = "../../../repos/pyo3/" }
```
whoahbot: Is the branch for this pushed up to GitHub? I believe we can pin to a github repo and SHA, so CI tests can run.

Psykopear: I'm currently using a branch on my own fork because the current PR needs a modified version of `chrono` that was not pushed anywhere. I proposed a fix on the PR to make it work with the current version of `chrono`, but the original author prefers to fix things in `chrono` and leave the PR as it is, so I'm waiting to know where this is going first.

I'm keeping this PR in draft until I know more, and then will link to the specific commit once it's usable.

CI would not pass yet anyway, I'm fixing a problem with the Python->Rust conversion right now, trying to instance a `TumblingWindowConfig` with a `start_at` parameter fails.

src/lib.rs
```diff
@@ -37,7 +37,7 @@ fn sleep_release_gil(py: Python, secs: u64) {
 #[pymodule]
 #[pyo3(name = "bytewax")]
 fn mod_bytewax(py: Python, m: &PyModule) -> PyResult<()> {
-    pyo3_log::init();
+    // pyo3_log::init();
```
whoahbot, line 40: TODO: Don't forget to uncomment this when it's available again.

pytests/test_datetime.py
```diff
@@ -0,0 +1,166 @@
...
+    flow_1 = Dataflow()
+    build_flow(flow_1, window_config_1, out_1)
+
+    out_2 = []
+    flow_2 = Dataflow()
+    build_flow(flow_2, window_config_2, out_2)
+
+    run_main(flow_1)
+    run_main(flow_2)
+    assert len(out_1) == len(out_2)
```
davidselassie, line 166: Similar questions as above test.

src/pyo3_extensions/chrono_tz.rs
```diff
@@ -0,0 +1,232 @@
...
+    type Output = Duration;
+
+    #[inline]
+    fn sub(self, rhs: ChronoDateTime) -> Duration {
+        self.0 - rhs.0
+    }
+}
+
+impl ToPyObject for ChronoDateTime {
+    fn to_object(&self, py: Python<'_>) -> PyObject {
```
davidselassie, line 54: Is there a way to piggyback on the PyO3 built-in TZ-ignoring conversions and then fill in the TZ info here, rather than re-creating the rest of the conversion? Or does that make things more confusing?

src/pyo3_extensions/chrono_tz.rs
```diff
@@ -0,0 +1,232 @@
...
+                key.parse()
+                    .map_err(|_err| PyTypeError::new_err("Timezone not found"))?
+            } else {
+                let utc = ob
+                    .py()
+                    .import("datetime")?
+                    .getattr("timezone")?
+                    .getattr("utc")?;
+                dt = dt.getattr("astimezone")?.call1((utc,))?.cast_as()?;
+                chrono_tz::Tz::UTC
```
davidselassie, line 108: What's your thinking behind doing the conversion to UTC and not trying to create a `chrono::offset::FixedOffset` in this case? It feels like it would be nice to round-trip a TZ datetime without this conversion.

davidselassie (commented):
Sorry I might be coming into this halfway and perhaps you already hashed these out, but I have some questions. Thanks!

pytests/test_datetime.py
```diff
@@ -0,0 +1,166 @@
...
+
+    out_2 = []
+    flow_2 = Dataflow()
+    build_flow(flow_2, window_config_2, out_2)
+
+    run_main(flow_1)
+    run_main(flow_2)
+
+    assert len(out_1) == len(out_2)
+    assert out_1 == out_2
```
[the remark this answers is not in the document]
Psykopear: this is a regression test (maybe should add a comment about it). It is testing a behaviour that is supposed to be always correct, but if you run this test using the `pyo3-chrono` integration we used before, this test fails because even if `start_at` and `start_at_1` represents the same date in python, the incorrect conversion in `pyo3-chrono` made the two flow run with a different `start_at` parameter, and gave two different results.

pytests/test_datetime.py
Psykopear: Similar concept, maybe this test is reduntant, but the idea is that settings `start_at` to `now` in python, and not doing it (so setting it in rust) should give the same result, but it didn't with `pyo3-chrono` if the `now` in python had different timezone info

pytests/test_datetime.py
```diff
@@ -0,0 +1,166 @@
...
+        start_at = datetime.now(tz=timezone(timedelta(hours=-8)))
+
+    wc = TumblingWindowConfig(length=timedelta(seconds=2), start_at=start_at)
+
+    # The dates should always be equivalent
+    assert start_at == wc.start_at
+    # But if we are working with fixed offsets in python<3.9,
+    # we convert everything to UTC, so the tzinfo won't be the same.
+    if py39:
+        assert start_at.tzinfo == wc.start_at.tzinfo
```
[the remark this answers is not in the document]
Psykopear: This would fail due to the choice of always having timezoneaware datetimes internally, and the way I do this conversion (see comments below).
Since I made the ChronoDateTime be `chrono::DateTime<chrono_tz::Tz>`, I can't use a `chrono::FixedOffset`. So when I receive a datetime with a fixed offset from python, i convert it to utc first. This means the `tzinfo` of the object that comes back from rust actually changes, even though it still represents the same date.

src/pyo3_extensions/chrono_tz.rs
```diff
@@ -0,0 +1,232 @@
...
+        if let Some(tz_dt) = tz.from_local_datetime(&dt).earliest() {
+            Ok(ChronoDateTime(tz_dt))
+        } else {
+            Err(PyTypeError::new_err("Can't convert datetime"))
+        }
+    }
+}
+
+// Implement serde (de)serialization manually here, since there is no standard
+// way to serialize datetimes with IANA timezone info (yet?).
```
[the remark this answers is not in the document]
Psykopear: Your understanding is correct, the point of using tz rather than a fixed offset is to keep the same information if the user provided a timezone info rather than an offset, see answers below for more context.

src/pyo3_extensions/chrono_tz.rs
Psykopear: I'll look into it, it makes sense to reuse it if we can

Psykopear:
So, I'm having a second look at this PR, and after some thinking I agree with @davidselassie here that we should force users to send us UTC datetimes, at least for the `start_at` parameter and the return type of the `dt_getter` function in the `EventTimeClock`.

Supporting any possible kind of datetime here makes everything more complicated than it needs to be.

Only accepting UTC datetimes will make this PR much simpler, I can remove the whole `chrono_tz` integration and just use the one on PyO3 once it's merged, and we can just use `DateTime<Utc>` instead of having a wrapper type in our structures.
It's a minor annoyance for users since a crash would happen either at the initialization of the flow (in case of a `start_at` parameter) or at the first event received (in case of the `EventTimeClock`), so it shouldn't lead to unexpected crashes when the flow is deployed to production (which was my worry when I started thinking about this).

I'm going to update this PR and make it ready for review.

davidselassie (approved):
I wasn't suggesting this with my comment, but this seems like a great way forward! UTC4ever.

whoahbot (approved):
I think we ended up with the better approach.

Feel free to merge when logging is available again.

Psykopear:
So, to have everything working we just need PyO3 `0.17.2`. The chrono integration will be merged then, and this should happen soon.
For the pyo3-log crate, we won't need to do anything once PyO3 0.17.2 is released, since it only depends on `~0.17`.
But, since this PR is blocking another one, I temporarily forked pyo3-log to make it compatible with the yet unreleased version of PyO3 (simply pinning its PyO3 dependency to the same commit we use here), so that we can merge this now.
I added 2 `TODO` notes in the `Cargo.toml` with what to do as soon as PyO3 0.17.2 is released, so that we can just swap the uncommented lines with the commented ones and we should be good to go.

PR #117, opened 2022-09-01
Update KafkaInput to prevent listening to non-existent topics

### LINKED ISSUE #116: No error from KafkaInputConfig when topic doesn't exist
### Bug Description

When you start a dataflow and the topic does not exist the dataflow will not share any details about that and continue to run in silence.

3.9

0.11.0

### Relevant log output

_No response_

### Steps to Reproduce

start a dataflow with a topic that doesn't exist

awmatheson:
Current method of getting partitions from metadata will return `Metadata` for a topic regardless of whether or not the topic exists and the partition value will be 0. rdkafka does not ever error if the topic exists so the bytewax worker0 ends up listening to a non-existent topic. 

This fix will just check if there are 0 partitions and then panic and let the user know that the topic doesn't exist.

This will fix #116

### DIFF
--- a/src/inputs/kafka_input.rs
+++ b/src/inputs/kafka_input.rs
@@ -127,12 +127,19 @@ fn get_kafka_partition_count(brokers: &[String], topic: &str) -> i32 {
 
     let timeout = Some(Duration::from_secs(5));
 
+    println!("Attempting to fetch metadata from {}", topic);
     let metadata = consumer
         .fetch_metadata(Some(topic), timeout)
         .map_err(|e| e.to_string())
         .expect("Unable to fetch topic metadata for {topic}");
-    let topic = &metadata.topics()[0];
-    topic
+
+    let user_topic = &metadata.topics()[0];
+    
+    if user_topic.partitions().len() == 0 {
+        panic!("Topic does not exist, please create first");
+    }
+
+    user_topic
         .partitions()
         .len()
         .try_into()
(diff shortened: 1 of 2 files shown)


Psykopear (approved):

src/inputs/kafka_input.rs
```diff
@@ -200,6 +202,9 @@ impl KafkaInput {
             .expect("Error building input Kafka consumer");
 
         let partition_count = get_kafka_partition_count(brokers, topic);
+        if partition_count == 0 {
+            panic!("Topic does not exist, please create first");
```
whoahbot, line 206: Do you want to include the topic name here in this message? Might be helpful when debugging.

awmatheson: good call, added.

src/inputs/kafka_input.rs
```diff
@@ -119,20 +119,18 @@ impl KafkaInputConfig {
...
+    log::debug!("Attempting to fetch metadata from {}", topic);
+    
     let metadata = consumer
         .fetch_metadata(Some(topic), timeout)
         .map_err(|e| e.to_string())
         .expect("Unable to fetch topic metadata for {topic}");
-    let topic = &metadata.topics()[0];
-    topic
+
+    let user_topic = &metadata.topics()[0];
```
blakestier, line 132: Will this panic if the `topics` array is empty?

blakestier: Maybe we can assume it's not since we only get metadata if the topic exists?

blakestier (approved):

PR #118, opened 2022-09-01
Kafka Output

blakestier:
Allows users to `capture` output into a Kafka stream.


examples/simple_kafka_in_and_out.py
```diff
@@ -15,17 +15,27 @@
...
-
 def deserialize(key_bytes__payload_bytes):
     key_bytes, payload_bytes = key_bytes__payload_bytes
     key = json.loads(key_bytes) if key_bytes else None
     payload = json.loads(payload_bytes) if payload_bytes else None
     return key, payload
 
+def serialize_with_key(key_payload):
+    key, payload = key_payload
+    new_key_bytes = key if key else json.dumps("my_key").encode('utf-8')
```
whoahbot, line 26: I think we should note here and in `ins_and_outs.md` that the key is optional, and will result in all data going to a single partition.

whoahbot (approved):
Looks good to me. Left you one note about adding docs for an optional output key.

PR #127, opened 2022-09-13
Event time processing

Psykopear:
This PR introduces a new Clock Config: `EventClockConfig`.

`EventClockConfig` is a clock based on datetimes retrieved from events.
Whenever an event is received, its datetime is extracted through a user specified function.
Events are awaited until a user specified duration is passed reached, if an event comes late it's dropped.

This configuration requires the user to specify 3 parameters:
- `dt_getter`: a function that receives an event as input and should return a datetime extracted from the event
- `late`: a duration after which (in event-time, not realtime) the window for waiting late events is closed
- `system_clock`: by default the system clock is used internally, but a TestingClock can be passed here

~This PR is ready for review, but is blocked by~:
- https://github.com/PyO3/pyo3/pull/2612
- https://github.com/bytewax/bytewax/pull/115

### TODO
- [x] Do not rely on `time.sleep` in tests
- [x] Add documentation
- [x] Add an example
- [x] More tests

### DIFF
--- a/pytests/test_event_time.py
+++ b/pytests/test_event_time.py
@@ -27,17 +27,17 @@ def test_event_time_processing():
         # This too should be processed in the first window
         clock.now = start_at + timedelta(seconds=6)
         yield None, {"type": "temp", "time": start_at + s(2), "value": 2}
-        # This should be processed in the second window
-        clock.now = start_at + timedelta(seconds=10.1)
-        yield None, {"type": "temp", "time": start_at + s(7), "value": 200}
-        # This should be processed in the third window
-        clock.now = start_at + timedelta(seconds=11.1)
-        yield None, {"type": "temp", "time": start_at + s(12), "value": 17}
         # This should be dropped, because its event_time is before
         # the latest event time, and it arrives `late` seconds after
         # the closing of the window + the delay of the latest received item.
         clock.now = start_at + timedelta(seconds=19.1)
         yield None, {"type": "temp", "time": start_at + s(1), "value": 200}
+        # This should be processed in the second window
+        clock.now += timedelta(seconds=1)
+        yield None, {"type": "temp", "time": start_at + s(7), "value": 200}
+        # This should be processed in the third window
+        clock.now += timedelta(seconds=1)
+        yield None, {"type": "temp", "time": start_at + s(12), "value": 17}
 
     def extract_sensor_type(event):
         return event["type"], event
--- a/pytests/test_event_time.py
+++ b/pytests/test_event_time.py
@@ -3,7 +3,7 @@ from datetime import datetime, timedelta, timezone
 from bytewax.dataflow import Dataflow
 from bytewax.execution import run_main
 from bytewax.inputs import ManualInputConfig
-from bytewax.window import TumblingWindowConfig, EventClockConfig, TestingClockConfig
+from bytewax.window import TumblingWindowConfig, EventClockConfig
 from bytewax.outputs import TestingOutputConfig
 from bytewax.testing import TestingClock
 
@@ -33,13 +33,9 @@ def test_event_time_processing():
         # This should be processed in the third window
         clock.now = start_at + timedelta(seconds=11.1)
         yield None, {"type": "temp", "time": start_at + s(12), "value": 17}
-        # This should be dropped, because its received "late".
-        # Calling "x" the time between latest_event_time (start_at + 2s) and
-        # the time we received the event at (start_at + 4s), events for the
-        # first window are late after:
-        # late + windows_length + x = 19 seconds
-        # ^      ^                ^
-        # 5s     10s              4s
+        # This should be dropped, because its event_time is before
+        # the latest event time, and it arrives `late` seconds after
+        # the closing of the window + the delay of the latest received item.
         clock.now = start_at + timedelta(seconds=19.1)
         yield None, {"type": "temp", "time": start_at + s(1), "value": 200}
 
--- a/pytests/test_event_time.py
+++ b/pytests/test_event_time.py
@@ -48,7 +48,7 @@ def test_event_time_processing():
 
     cc = EventClockConfig(
         lambda event: event["time"],
-        late_after_system_duration=late,
+        wait_for_system_duration=late,
         system_clock=clock
     )
     wc = TumblingWindowConfig(start_at=start_at, length=window_length)
--- a/pytests/test_event_time.py
+++ b/pytests/test_event_time.py
@@ -1,0 +1,77 @@
+from datetime import datetime, timedelta
+
+from bytewax.dataflow import Dataflow
+from bytewax.execution import run_main
+from bytewax.inputs import ManualInputConfig
+from bytewax.window import TumblingWindowConfig, EventClockConfig, TestingClockConfig
+from bytewax.outputs import TestingOutputConfig
+
+
+def test_event_time_processing():
+    """
+    Test used to validate the EventClockConfig workings.
+    """
+    start_at = datetime.now()
+    late = timedelta(seconds=10)
+    window_length = timedelta(seconds=5)
+
+    # Utility functions to generate "temp" events
+    def temp(val, seconds):
+        time = start_at + timedelta(seconds=seconds)
+        return None, {"type": "temp", "time": time, "value": val}
+
+    # Utility functions to generate wait events (used for advancing the clock)
+    def wait():
+        return None, {"type": "temp", "time": start_at}
+
+    def input_builder(worker_index, worker_count, state):
+        # Each yield advances the clock by 1 second
+        # This should be processed in the first window
+        yield temp(1, 1)
+        # This too should be processed in the first window
+        yield temp(2, 2)
+        # Wait 10 seconds
+        for _ in range(10):
+            yield wait()
+        # This should be dropped, because its received is after start_at + late
+        yield temp(200, 1)
+        # This should be processed in the second window
+        yield temp(200, 7)
+        # This should be processed in the third window
+        yield temp(17, 12)
+
+    def extract_sensor_type(event):
+        return event["type"], event
+
+    def acc_values(acc, event):
+        if "value" in event:
+            acc.append(event["value"])
+        return acc
+
+    cc = EventClockConfig(
+        lambda event: event["time"],
+        late_after_system_duration=late,
+        system_clock_config=TestingClockConfig(start_at=start_at, item_incr=timedelta(seconds=1))
+    )
+    wc = TumblingWindowConfig(start_at=start_at, length=window_length)
+
+    flow = Dataflow()
+    flow.input("input", ManualInputConfig(input_builder))
+
+    flow.map(extract_sensor_type)
+    flow.fold_window("running_average", cc, wc, list, acc_values)
+    flow.map(lambda x: {f"{x[0]}_avg": sum(x[1]) / len(x[1])})
+
+    out = []
+    flow.capture(TestingOutputConfig(out))
+    run_main(flow)
+
+    expected = [
+        {"temp_avg": 1.5},
+        {"temp_avg": 200},
+        {"temp_avg": 17},
+    ]
+    assert len(out) == 3
+    assert expected[0] in out
+    assert expected[1] in out
+    assert expected[2] in out
--- a/pytests/test_event_time.py
+++ b/pytests/test_event_time.py
@@ -1,67 +1,67 @@
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 
 from bytewax.dataflow import Dataflow
 from bytewax.execution import run_main
 from bytewax.inputs import ManualInputConfig
 from bytewax.window import TumblingWindowConfig, EventClockConfig, TestingClockConfig
 from bytewax.outputs import TestingOutputConfig
+from bytewax.testing import TestingClock
 
 
 def test_event_time_processing():
     """
     Test used to validate the EventClockConfig workings.
     """
-    start_at = datetime.now()
+    start_at = datetime(2022, 1, 1, 12, 0, 0, tzinfo=timezone.utc)
+    clock = TestingClock(start_at)
     late = timedelta(seconds=10)
     window_length = timedelta(seconds=5)
 
-    # Utility functions to generate "temp" events
-    def temp(val, seconds):
-        time = start_at + timedelta(seconds=seconds)
-        return None, {"type": "temp", "time": time, "value": val}
-
-    # Utility functions to generate wait events (used for advancing the clock)
-    def wait():
-        return None, {"type": "temp", "time": start_at}
+    def s(val):
+        return timedelta(seconds=val)
 
     def input_builder(worker_index, worker_count, state):
-        # Each yield advances the clock by 1 second
+        clock.now = start_at + timedelta(seconds=2)
         # This should be processed in the first window
-        yield temp(1, 1)
+        yield None, {"type": "temp", "time": start_at, "value": 1}
         # This too should be processed in the first window
-        yield temp(2, 2)
-        # Wait 10 seconds
-        for _ in range(10):
-            yield wait()
-        # This should be dropped, because its received is after start_at + late
-        yield temp(200, 1)
+        clock.now = start_at + timedelta(seconds=6)
+        yield None, {"type": "temp", "time": start_at + s(2), "value": 2}
         # This should be processed in the second window
-        yield temp(200, 7)
+        clock.now = start_at + timedelta(seconds=10.1)
+        yield None, {"type": "temp", "time": start_at + s(7), "value": 200}
         # This should be processed in the third window
-        yield temp(17, 12)
+        clock.now = start_at + timedelta(seconds=11.1)
+        yield None, {"type": "temp", "time": start_at + s(12), "value": 17}
+        # This should be dropped, because its received "late".
+        # Calling "x" the time between latest_event_time (start_at + 2s) and
+        # the time we received the event at (start_at + 4s), events for the
+        # first window are late after:
+        # late + windows_length + x = 19 seconds
+        # ^      ^                ^
+        # 5s     10s              4s
+        clock.now = start_at + timedelta(seconds=19.1)
+        yield None, {"type": "temp", "time": start_at + s(1), "value": 200}
 
     def extract_sensor_type(event):
         return event["type"], event
 
     def acc_values(acc, event):
-        if "value" in event:
-            acc.append(event["value"])
+        acc.append(event["value"])
         return acc
 
     cc = EventClockConfig(
         lambda event: event["time"],
         late_after_system_duration=late,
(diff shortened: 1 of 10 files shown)


Psykopear:
@davidselassie I updated the PR following yesterday's conversation, using SystemTime do determine events' lateness.
I'm going to work on the possibility of using a TestingClock in tests for the EventTimeConfig, so that we don't need to rely on `time.sleep` in CI

src/window/event_time_clock.rs
```diff
@@ -0,0 +1,225 @@
...
+///
+/// Args:
+///   dt_getter: Python function to get a datetime from an event.
+///   late: How much (system) time to wait before considering an event late.
+///
+/// Returns:
+///   Config object. Pass this as the `clock_config` parameter to your
+///   windowing operator.
+#[pyclass(module="bytewax.window", extends=ClockConfig)]
+#[pyo3(text_signature = "(item_incr)")]
```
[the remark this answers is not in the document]
Psykopear: Right, fixing it

src/window/event_time_clock.rs
```diff
@@ -0,0 +1,225 @@
...
+#[derive(Clone)]
+pub(crate) struct EventClockConfig {
+    #[pyo3(get)]
+    pub(crate) dt_getter: TdPyCallable,
+    #[pyo3(get)]
+    pub(crate) late: chrono::Duration,
```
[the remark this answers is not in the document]
Psykopear: Sure, changed to `late_after_system_duration`

src/window/event_time_clock.rs
```diff
@@ -0,0 +1,225 @@
...
+        match next_value {
+            // If there will be no more values, close out all windows.
+            Poll::Ready(None) => ChronoDateTime::max_utc(),
+            // If we received an event, set the new watermark and return it
+            Poll::Ready(Some(event)) => {
+                self.set_watermark(self.get_event_time(event));
+                self.watermark
+            }
+            // Return the saved watermark
+            _ => self.watermark,
```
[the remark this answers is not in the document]
Psykopear: I felt I was missing something here, thanks.
So, whenever the watermark is asked for, I advance it by the time elapsed from the system_time_of_last_event, and when a new event comes, I only update the watermark if the newly calculated one is later than the current one right? I'm going to push a fix, but please review it again.

src/window/event_time_clock.rs
```diff
@@ -0,0 +1,225 @@
...
+        }
+    }
+
+    fn time_for(&mut self, event: &V) -> ChronoDateTime {
+        self.get_event_time(event)
+    }
+
+    fn snapshot(&self) -> StateBytes {
+        StateBytes::ser(&(self.latest_event_time, self.watermark))
```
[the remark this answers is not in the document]
Psykopear: Right, I only added the ability to use the TestingClock later and forgot to add it to state recovery. I fixed it for now by manually selecting the fields to serialize depending on the clock's type, but I'm not sure this is the proper way to go at this point, see the comment below

src/window/event_time_clock.rs
```diff
@@ -0,0 +1,225 @@
...
+#[pyclass(module="bytewax.window", extends=ClockConfig)]
+#[pyo3(text_signature = "(item_incr)")]
+#[derive(Clone)]
+pub(crate) struct EventClockConfig {
+    #[pyo3(get)]
+    pub(crate) dt_getter: TdPyCallable,
+    #[pyo3(get)]
+    pub(crate) late: chrono::Duration,
+    #[pyo3(get)]
+    pub(crate) internal_clock: Option<TestingClockConfig>,
```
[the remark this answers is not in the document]
Psykopear: Makes sense, changing this

src/window/event_time_clock.rs
```diff
@@ -0,0 +1,225 @@
...
+        let now = self.clock.time();
+        self.watermark = now
+            .checked_add_signed(now - self.latest_event_time.unwrap())
+            .unwrap()
+            .checked_sub_signed(self.late)
+            .unwrap();
+    }
+}
+
+impl<V: ToPyObject> Clock<V> for EventClock {
```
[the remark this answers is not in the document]
Psykopear: Yep, that makes much more sense since I only need this constraint on the EventClock, fixing this

src/window/mod.rs
```diff
@@ -191,7 +196,7 @@ pub(crate) fn build_windower_builder(
 }
 
 /// Defines the sense of time for a windowing operator.
-pub(crate) trait Clock<V> {
+pub(crate) trait Clock<V: ToPyObject> {
```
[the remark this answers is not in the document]
Psykopear: Uh, didn't realize the repackaging as a pure Rust crate was a possibility, I'm going to be more careful about this in the future. Also, not changing the whole trait for just one kind of clock makes sense, fixed it.
I still need the `build_clock_builder` function require that `V` is a `TdPyAny` for any clock, not sure how to avoid this for now, but it's a more contained requirement at least.

Psykopear:
Thanks for the thorough review @davidselassie , this was needed.
I addressed most of the comments.

Right now the tests are broken due to the use of the `TestingClock` as the internal clock in the test, since we increment its internal counter more than once per event after the fix in `EventTimeClock::watermark`.

As I said there, I'm tempted to go back to having a `TestingEventTimeClock` for the sake of simplicity, but let me know if you prefer me to find a way to make it work as it is.

Do we have a use case outside of testing to have different kind of clocks for the `EventTimeClock`?

pytests/test_event_time.py
```diff
@@ -0,0 +1,77 @@
...
+        return None, {"type": "temp", "time": time, "value": val}
+
+    # Utility functions to generate wait events (used for advancing the clock)
+    def wait():
+        return None, {"type": "temp", "time": start_at}
+
+    def input_builder(worker_index, worker_count, state):
+        # Each yield advances the clock by 1 second
+        # This should be processed in the first window
+        yield temp(1, 1)
```
blakestier, line 30: Appreciate the comments! Optionally, you could modify the events to be strings instead of temp values, so their values could be more revealing, i.e. "window 1", "drop since late", "window 2".

src/window/event_time_clock.rs
```diff
@@ -0,0 +1,276 @@
...
+/// Use datetimes from events as clock.
+///
+/// If the dataflow has no more input, all windows are closed.
+///
+/// The watermark is the system time since the last element
+/// plus the value of `late`. It is updated every time an event
+/// with a newer datetime is processed.
+///
+/// Args:
+///   getter: Python function to get a datetime from an event.
```
blakestier, line 26: this is called `dt_getter` below.

src/window/event_time_clock.rs
```diff
@@ -0,0 +1,276 @@
...
+#[pyclass(module="bytewax.window", extends=ClockConfig)]
+#[pyo3(text_signature = "(getter, late_after_system_duration, system_clock_config)")]
+#[derive(Clone)]
+pub(crate) struct EventClockConfig {
+    #[pyo3(get)]
+    pub(crate) dt_getter: TdPyCallable,
+    #[pyo3(get)]
+    pub(crate) late_after_system_duration: chrono::Duration,
+    #[pyo3(get)]
+    pub(crate) system_clock_config: Option<TestingClockConfig>,
```
blakestier, line 42: We have an optional TestingClockConfig here to allow for testing windowing with event time?

src/window/event_time_clock.rs
```diff
@@ -0,0 +1,277 @@
...
+        // Assign it to self.latest_event_time
+        self.latest_event_time = Some(latest_event_time);
+
+        // Get current clock's time
+        let now = self.clock.time();
+        // Get watermark
+        let watermark = now
+            .checked_add_signed(now - latest_event_time)
+            // Panic if the watermark overflowed
+            .expect("watermark in date range (overflowed)")
```
davidselassie, line 209: It looks like `+` and `-` overloading internally does `check_add_signed` then `expect`. https://docs.rs/chrono/latest/src/chrono/datetime/mod.rs.html#853-860 So change all these to use the operators instead? Would make it more concise.

src/window/event_time_clock.rs
```diff
@@ -0,0 +1,277 @@
...
+        // one, since it can be advanced by system time too.
+        self.watermark = watermark.max(self.watermark);
+
+        // Also update self.system_time_of_last_event
+        self.system_time_of_last_event = Some(now);
+    }
+
+    fn advance_watermark_by_system_time(&mut self) {
+        let now = self.clock.time();
+        let elapsed = now - self.system_time_of_last_event.unwrap_or(self.watermark);
```
davidselassie, line 223: I believe you could drop the `Option` on `system_time_of_last_event` if you initialize it to the system time when the clock is created. This means that the watermark advances by system time even before the first event, which I think makes sense?

src/window/system_clock.rs
```diff
@@ -18,8 +18,15 @@ use super::{Clock, ClockConfig};
 ///   Config object. Pass this as the `clock_config` parameter to
 ///   your windowing operator.
 #[pyclass(module="bytewax.window", extends=ClockConfig)]
+#[derive(Clone, Copy)]
```
davidselassie, line 21: Why are these suddenly needed on our config types?

src/window/event_time_clock.rs
```diff
@@ -0,0 +1,277 @@
...
+            .expect("watermark in date range (overflowed)")
+            .checked_sub_signed(self.late)
+            // Panic if the watermark underflowed
+            .expect("watermark in date range (underflowed)");
+        // But only assign it to self.watermark if it's after the current
+        // one, since it can be advanced by system time too.
+        self.watermark = watermark.max(self.watermark);
+
+        // Also update self.system_time_of_last_event
+        self.system_time_of_last_event = Some(now);
```
davidselassie, line 218: I think we only want to save this whenever we get an event that is "advancing". Another way of saying, late events should not cause the watermark to go backwards.

src/window/event_time_clock.rs
```diff
@@ -0,0 +1,277 @@
...
+        let elapsed = now - self.system_time_of_last_event.unwrap_or(self.watermark);
+        self.watermark = self
+            .watermark
+            .checked_add_signed(elapsed)
+            .expect("watermark in date range (overflowed)");
+    }
+}
+
+impl Clock<TdPyAny> for EventClock {
+    fn watermark(&mut self, next_value: &Poll<Option<TdPyAny>>) -> DateTime<Utc> {
```
davidselassie, line 232: After thinking through all these other comments, perhaps a simpler way to phrase this is to store `late_time` which is a sort of "pure event time watermark", advancing when new (later) elements come in. Then always add the system time to that to generate the real watermark on the fly (since the system time is always changing on every call).

Is that generally right? Does that make things more or less clear? I think what you have here is also correct too.

```rs
match next_value {
    Poll::Ready(Some(event)) => {
        let event_late_time = self.time_for(event) - late;
        if event_late_time > self.late_time {
            self.late_time = event_late_time;
            self.system_time_of_last_event = now;
        }
    }
    Poll::Ready(None) => {
        self.late_time = MAX_UTC;
        self.system_time_of_last_event = now;
    }
    Poll::Pending => {}
}

let system_duration_since_last_event = now - self.system_time_of_last_event;
let watermark = self.late_time.checked_add_signed(system_duration_since_last_event).unwrap_or(MAX_UTC);
watermark
```

src/window/event_time_clock.rs
```diff
@@ -0,0 +1,277 @@
...
+            Self::Testing(clock) => clock.time_for(&()),
+            Self::System(clock) => clock.time_for(&()),
+        }
+    }
+}
+
+pub(crate) struct EventClock {
+    dt_getter: TdPyCallable,
+    late: chrono::Duration,
+    clock: InternalClock,
```
davidselassie, line 154: `system_clock`?

pytests/test_event_time.py
```diff
@@ -0,0 +1,77 @@
...
+        # This too should be processed in the first window
+        yield temp(2, 2)
+        # Wait 10 seconds
+        for _ in range(10):
```
davidselassie, line 34: Hmm. Not for this PR. But do you think it would be easier if `TestingClock` instead had a `time_getter` and let you explicitly set the time in Python in the tests? I think this pattern of needing items to implicitly advance the clock is getting more confusing than clear. I might prototype this out in another PR.

davidselassie (commented):
Thank you for working so hard on this. I try not to nitpick people's algorithms, but I wanted to really double check that this produced the correct behavior and happened to think of some maybe clearer phrasing? Feel free to discard this advice on your discretion.

pytests/test_event_time.py
davidselassie: Here I think this will help: https://github.com/bytewax/bytewax/pull/139/

src/window/system_clock.rs
Psykopear: So, this is due to the new `ClockBuilder` trait, which consumes `self`.

Previously in the `build_clock_builder` function we were manually cloning all the fields in the config, and passing them to the builder. With the new trait we clone the whole config instead (required by the `FromPyObject` trait, used in `extract`). So instead of implicitly requiring all the fields in the config to be `Clone`, we require the config itself to be cloneable.
Since we end up cloning all the fields in the end, I thought having this requirement here wouldn't do much of a difference.

This change is not needed for this PR, I can easily revert it, remove the `ClockBuilder` trait, and implement the `builder` function on the struct like it was previously. It's a refactoring I did along the way to make the pattern we use more explicit, and I feel it makes the code more readable, but it works the other way too.

src/window/event_time_clock.rs
Psykopear: Ok, now it is clear what we want here, fixed

src/window/event_time_clock.rs
Psykopear: yep, dt_getter was right, fixed it

src/window/event_time_clock.rs
```diff
@@ -0,0 +1,277 @@
...
+            watermark,
+            clock,
+        }
+    }
+
+    fn get_event_time(&self, event: &TdPyAny) -> DateTime<Utc> {
+        Python::with_gil(|py| {
+            self.dt_getter
+                // Call the event time getter function with the event as parameter
+                .call1(py, (event.to_object(py),))
```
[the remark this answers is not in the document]
Psykopear: fixed

src/window/event_time_clock.rs
```diff
@@ -0,0 +1,277 @@
...
+    #[pyo3(get)]
+    pub(crate) system_clock_config: Option<TestingClockConfig>,
+}
+
+impl ClockBuilder<TdPyAny> for EventClockConfig {
+    fn builder(self) -> Builder<TdPyAny> {
+        Box::new(move |resume_state_bytes: Option<StateBytes>| {
+            let (latest_event_time, system_time_of_last_event, watermark) = resume_state_bytes
+                .map(StateBytes::de)
+                .unwrap_or_else(|| (None, None, Utc::now()));
```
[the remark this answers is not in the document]
Psykopear: Fixed

src/window/event_time_clock.rs
```diff
@@ -0,0 +1,277 @@
...
+            dt_getter,
+            late,
+            latest_event_time,
+            system_time_of_last_event,
+            watermark,
+            clock,
+        }
+    }
+
+    fn get_event_time(&self, event: &TdPyAny) -> DateTime<Utc> {
```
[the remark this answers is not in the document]
Psykopear: Moved there, I initially had it in a separate function because I was using it in the `watermark` function, but that's not the case anymore

src/window/event_time_clock.rs
```diff
@@ -0,0 +1,277 @@
...
+            }
+        }
+    }
+
+    fn time_for(&mut self, event: &TdPyAny) -> DateTime<Utc> {
+        self.get_event_time(event)
+    }
+
+    fn snapshot(&self) -> StateBytes {
+        // Manually implement snapshotting here, because we'd have to
```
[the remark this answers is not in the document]
Psykopear: Ok, using the double serialization now, I also tested it to ensure it works, review the test too please

src/window/event_time_clock.rs
Psykopear: since the watermark calculation is changed, this comment is outdated. We do a `checked_add_signed` but handle the overflow case by unwrapping to MAX_UTC

src/window/event_time_clock.rs
Psykopear: Done

src/window/event_time_clock.rs
Psykopear: yep, fixed

pytests/test_event_time.py
Psykopear: Using the new testing clock now, so this is fixed

src/window/event_time_clock.rs
Psykopear: It's now slightly different, but yes, we still accept a parameter to use a testing clock rather than the default system clock in the event time clock. As I said in another comment, initially I implemented two separate clocks, `EventTimeClock` and `TestingEventTimeClock`, with the only difference between the two that the testing clock was using an auto incremented counter rather than system's `now`.
This meant keeping the two implementations in sync by hand, and could lead to problems if not (a test passing when it shouldn't because test and production use two different implementation of the event time clock), so I decided to find a way to reuse the same EventClock changing only the meaning of `now`.
I'm still not convinced it's the best approach, since there is a parameter needed only for testing inside a "production" object, but I felt this was better than the way I did it previously.

Open to suggestions though

src/window/event_time_clock.rs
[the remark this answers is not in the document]
Psykopear: Fixed with the double serialization pattern we discussed in other comments

whoahbot (approved):

src/window/event_time_clock.rs
```diff
@@ -0,0 +1,328 @@
...
+///   wait_for_system_duration: How much (system) time to wait before considering an event late.
+///
+///   system_clock: Optional configuration to use a PyTestingClock as the internal system_clock,
+///                 only used for testing purposes right now.
+///
+/// Returns:
+///   Config object. Pass this as the `clock_config` parameter to your
+///   windowing operator.
+#[pyclass(module="bytewax.window", extends=ClockConfig)]
+#[pyo3(text_signature = "(dt_getter, wait_for_system_duration, system_clock)")]
```
blakestier, line 39: I think `wait_for_system_duration` is explicit but perhaps leaking implementation. What about something like `lateness_buffer` or `late_after_duration` or something?

davidselassie (approved):

src/window/event_time_clock.rs
davidselassie: Yes, it's "internal implementation" but it's also the correct interpretation of this duration, as there's sort of multiple types of time going on.

PR #128, opened 2022-09-13
`downcast` instead of `extract`

whoahbot:
Based on some benchmarks, calling extract after pickling an object proved to be 1.3x slower than using `downcast` into a `PyBytes` struct and calling `as_bytes()`.

I struggled a bit with having to `map_err()` so often, due to lifetime constraints and would welcome suggestions on how to refactor that.

### DIFF
--- a/src/pyo3_extensions/mod.rs
+++ b/src/pyo3_extensions/mod.rs
@@ -115,13 +115,18 @@ impl serde::Serialize for TdPyAny {
     where
         S: serde::Serializer,
     {
-        let bytes: Result<Vec<u8>, PyErr> = Python::with_gil(|py| {
+        Python::with_gil(|py| {
             let x = self.as_ref(py);
-            let pickle = py.import("dill")?;
-            let bytes = pickle.call_method1("dumps", (x,))?.extract()?;
-            Ok(bytes)
-        });
-        serializer.serialize_bytes(bytes.map_err(S::Error::custom)?.as_slice())
+            let pickle = py.import("dill").map_err(S::Error::custom)?;
+            let bytes = pickle
+                .call_method1("dumps", (x,))
+                .map_err(S::Error::custom)?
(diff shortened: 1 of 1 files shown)


src/pyo3_extensions/mod.rs
```diff
@@ -115,13 +115,18 @@ impl serde::Serialize for TdPyAny {
...
     {
-        let bytes: Result<Vec<u8>, PyErr> = Python::with_gil(|py| {
+        Python::with_gil(|py| {
             let x = self.as_ref(py);
-            let pickle = py.import("dill")?;
-            let bytes = pickle.call_method1("dumps", (x,))?.extract()?;
-            Ok(bytes)
-        });
-        serializer.serialize_bytes(bytes.map_err(S::Error::custom)?.as_slice())
+            let pickle = py.import("dill").map_err(S::Error::custom)?;
```
davidselassie, line 120: I think you could get away with only one `map_err` if you use a closure and immediately call it (one day we can use [try blocks](https://doc.rust-lang.org/beta/unstable-book/language-features/try-blocks.html)) in order to "capture" the `PyResult` error type first. It's a little annoying that you'll probably need type annotations, though. Not sure if this is more clear or not.

```rust
Python::with_gil(|py| {
    let pybytes: PyResult<PyBytes> = || {
        let pickle = ...?;
        pickle...
            .downcast::<PyBytes>()?
    }();
    serializer.serialize_bytes(bytes.map_err(S::Error::custom)?.as_bytes())
})
```

davidselassie (approved):
This looks good to me. One trick that I think works, but you can decide if its too tricky.

src/pyo3_extensions/mod.rs
whoahbot: I fooled around with it a bit more, but ran into other problems that make me think I'll just wait for try blocks. Thanks for the suggestion!

PR #129, opened 2022-09-13
Only snapshot for recovery once per state key

davidselassie:
We only need to snapshot the logic once per state key. Stores
activated state keys in a `HashSet` to do this. Previously, we were
building up a `Vec`, pushing the state key onto the end with each
input data with that state key; this means we would snapshot once for
each input item.

Takes a look at all of the persistent data structures as part of the
`StatefulUnary` operator and changes them to be their most semantic
versions. Hopefully that'll prevent other bugs like this here:

- `activate_epochs_buffer` to a sorted set (we should activate the
  epochs in order and just once).

- `keys_buffer` to a set (we just need one copy of each key).

- `activated_epochs_to_keys_buffer` to a map to a set (the original
  problem).

### DIFF
--- a/src/recovery/mod.rs
+++ b/src/recovery/mod.rs
@@ -220,7 +220,9 @@ use serde::Serialize;
 use std::any::type_name;
 use std::collections::hash_map::DefaultHasher;
 use std::collections::hash_map::Entry;
+use std::collections::BTreeSet;
 use std::collections::HashMap;
+use std::collections::HashSet;
 use std::fmt::Debug;
 use std::fmt::Display;
 use std::hash::Hash;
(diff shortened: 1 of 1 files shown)


src/recovery/mod.rs
```diff
@@ -1276,15 +1279,11 @@ where
...
                     );
                     // Eagerly execute the frontier.
-                    activate_epochs_buffer.push(frontier_epoch);
-                    activate_epochs_buffer.dedup();
-                    // We promise to execute epochs in order so the
-                    // state is deterministic.
-                    activate_epochs_buffer.sort();
+                    activate_epochs_buffer.insert(frontier_epoch);
 
                     // Drain to re-use buffer. For each epoch in
```
whoahbot, line 1284: Comment still mentions that we drain the buffer here, is that referring to the `keys_buffer` below?

whoahbot (approved):
Left you one non-blocking comment.

blakestier (approved):

PR #130, opened 2022-09-13
Opt out of snapshots

blakestier:
Even if a dataflow does not opt into recovery, we still `snapshot` state and therefore do some unnecessary serialization. 

This PR adds a bool to avoid `snapshotting` if recovery is disabled.

### DIFF
--- a/src/recovery/mod.rs
+++ b/src/recovery/mod.rs
@@ -1088,7 +1088,9 @@ pub(crate) trait StatefulUnary<S: Scope, V: ExchangeData> {
     /// This is the core Timely operator that all Bytewax stateful
     /// operators are implemented in terms of. It is awkwardly generic
     /// because of that. We do this so we only have to implement the
-    /// very tricky recovery system interop once.
+    /// very tricky recovery system interop once. That said, setting
+    /// `take_snapshot` to false will prevent unnecessary serialization
+    /// in the case that the dataflow opts out of recovery.
     ///
     /// # Input
     ///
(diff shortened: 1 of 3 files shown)


whoahbot (approved):

src/recovery/mod.rs
```diff
@@ -1392,16 +1396,18 @@ where
...
-                                for key in keys {
-                                    if let Some(logic) = logic_cache.remove(&key) {
-                                        let state_bytes = logic.snapshot();
-                                        // Retain logic if not a
-                                        // reset.
-                                        if let StateUpdate::Upsert(_) = state_bytes {
-                                            logic_cache.insert(key.clone(), logic);
+                                if take_snapshot {
+                                    for key in keys {
+                                        if let Some(logic) = logic_cache.remove(&key) {
```
davidselassie, line 1401: Thinking about this more, this will break the behavior of stateful operators. We use the return value of `logic.snapshot()` to determine if the entire logic is thrown away within `StatefulUnary`; if we don't throw it away, then state will remain even though the logic wanted it gone. Sorry I didn't catch this earlier.

This behavior requires something tricky because no method on `StatefulLogic` can cause it to "drop itself", it needs to signal one level up to `StatefulUnary` to do so.

Not sure what I was thinking before, but a better API here would maybe be to have `StatefulLogic::exec` return a third tuple item `bool` which `true` signals "throw me away". And then `StatefulLogic::snapshot` wouldn't need to deal with `StateBytes::Reset` at all. With a little more thinking, there might be something even cleaner than that. But in any case this becomes a much bigger change, though...

davidselassie (changes requested):

src/recovery/mod.rs
blakestier: Hmm, yeah, that was my concern that I mentioned yesterday around stateful operators. There's no way to peel that apart? Well, we can discuss as a team.

davidselassie: I think our takeaway from meeting was "we should do a bigger change that disables all recovery code eventually, but not today and instead should look into some other performance improvements first".

PR #137, opened 2022-09-20
Introduce fate

davidselassie:
Previously I made the questionable API design of having
`StatefulLogic::snapshot` return a `StateUpdate`. This meant that the
snapshotting process affected the behavior of the stateful operator,
since it could return `Reset` to signal up to `StatefulUnary` to
discard the logic.

This unwinds that tangle by introducing a new method with a perhaps
overly poetic name `StatefulLogic::fate` which should return what
`StatefulUnary` should do with the logic when it's done processing via
a `LogicFate` enum. There are three options:

1. Retain it until a new item for this key comes in.
2. Retain it and awaken it after a timeout.
3. Discard it.

`fate` is attempting to encapsulate the problem of `StatefulUnary` is
the owner of the logics, so they can't drop themselves. And since
awakening timeouts are part of that process, they're in there too.

This is nice because it simplifies the return value of `exec` to just
output. The awaken delay is handled in `fate`.

It also uses this pattern within `Windower::fate` with `WindowerFate`
for the same kind of problem: the `WindowStatefulLogic` is the owner,
and we need to communicate back when it's safe to discard a
`Windower`. This fixes the bug of never discarding window state if a
key is never seen again: we now discard that state whenever all
windows for a key are closed.

A few other small changes:

- Standardises on the language of "awaken" a logic. Still uses
  Timely's "activate" for a Timely operator, though. Renames
  `StatefulLogic::exec` to `StatefulLogic::awake_with` to make more
  explicit when it is called. Renames `WindowLogic::exec` to
  `WindowLogic::with_next` to make explicit when it's called.

- All stateful operator tests should test recovery as part of testing
  the logic. We should do this to excercise the serde round-trip. I
  found three? bugs via this where recovery would panic because the
  deserialization wasn't to the correct type. As part of this, I added
  explicit type annotations to all `StateBytes::ser` and
  `StateBytes::de` calls so you can compare them.

- Clarifies more comments in the giant chunk of code for
  `StatefulUnary::stateful_unary`. I was also able to optimize it a
  little and get rid of two of our temporary buffers. I think it makes
  the process slightly clearer.

### DIFF
--- a/pytests/test_operators.py
+++ b/pytests/test_operators.py
@@ -3,7 +3,7 @@ from collections import defaultdict
 from datetime import datetime, timedelta, timezone
 from threading import Event
 
-from pytest import fixture, raises
+from pytest import fixture, mark, raises
 
 from bytewax.dataflow import Dataflow
 from bytewax.execution import run_main, TestingEpochConfig
@@ -341,6 +341,10 @@ def test_stateful_map_error_on_non_string_key():
         run_main(flow)
 
 
+@mark.skip(
+    "This test will not work with system time consistently until we mock the awaken "
+    "times in StatefulUnary."
+)
 def test_reduce_window(recovery_config):
     start_at = datetime(2022, 1, 1, tzinfo=timezone.utc)
     clock = TestingClock(start_at)
@@ -405,6 +409,10 @@ def test_reduce_window(recovery_config):
     assert sorted(out) == sorted([("ALL", 3), ("ALL", 1)])
 
 
+@mark.skip(
+    "This test will not work with system time consistently until we mock the awaken "
+    "times in StatefulUnary."
+)
 def test_fold_window(recovery_config):
     start_at = datetime(2022, 1, 1, tzinfo=timezone.utc)
     clock = TestingClock(start_at)
@@ -463,7 +471,7 @@ def test_fold_window(recovery_config):
         counts[typ] += 1
         return counts
 
-    flow.fold_window("sum", clock_config, window_config, dict, count)
+    flow.fold_window("count", clock_config, window_config, dict, count)
 
     out = []
     flow.capture(TestingOutputConfig(out))
--- a/src/operators/fold_window.rs
+++ b/src/operators/fold_window.rs
@@ -37,14 +37,14 @@ impl FoldWindowLogic {
 }
 
 impl WindowLogic<TdPyAny, TdPyAny, Option<TdPyAny>> for FoldWindowLogic {
-    fn exec(&mut self, next_value: Option<TdPyAny>) -> Option<TdPyAny> {
+    fn with_next(&mut self, next_value: Option<TdPyAny>) -> Option<TdPyAny> {
         match next_value {
             Some(value) => Python::with_gil(|py| {
                 let acc: TdPyAny = self
                     .acc
                     .take()
                     .unwrap_or_else(|| unwrap_any!(self.builder.call1(py, ())).into());
-                // Call the folder with the initialized accumulator
+                // Call the folder with the initialized accumulator.
                 let updated_acc = unwrap_any!(self
                     .folder
                     .call1(py, (acc.clone_ref(py), value.clone_ref(py))))
--- a/src/recovery/mod.rs
+++ b/src/recovery/mod.rs
@@ -209,6 +209,8 @@
 //! resume with the state from the end of the epoch just before
 //! failure, with the input resuming from beginning of the next epoch.
 
+use chrono::DateTime;
+use chrono::Utc;
 use log::debug;
 use pyo3::exceptions::PyTypeError;
 use pyo3::exceptions::PyValueError;
@@ -229,7 +231,6 @@ use std::hash::Hash;
 use std::hash::Hasher;
 use std::task::Poll;
 use std::time::Duration;
-use std::time::Instant;
 use timely::communication::Allocate;
 use timely::dataflow::channels::pact;
 use timely::dataflow::channels::pact::ParallelizationContract;
@@ -472,12 +473,30 @@ impl StateBytes {
 
 /// The two kinds of actions that logic in a stateful operator can do
 /// to each [`StateKey`].
+///
+/// These are the manifestations of the options in [`LogicFate`] for
+/// the recovery system.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub(crate) enum StateUpdate {
-    /// Save an updated state snapshot.
+    /// The logic was retained and this is the snapshot of the current state.
     Upsert(StateBytes),
-    /// Note that the state for this key was reset to empty.
-    Reset,
+    /// The logic was discarded. No need to keep the latest snapshot
+    /// of the state.
+    Discard,
+}
+
+/// If a [`StatefulLogic`] for a key should be retained by
+/// [`StatefulUnary::stateful_unary`].
+///
+/// See [`StatefulLogic::fate`].
+pub(crate) enum LogicFate {
+    /// This logic for this key should be retained and used again
+    /// whenever new data for this key comes in.
+    Retain,
+    /// The logic for this key is "complete" and should be
+    /// discarded. It will be built again if the key is encountered
+    /// again.
+    Discard,
 }
 
 /// Impl this trait to create an operator which maintains recoverable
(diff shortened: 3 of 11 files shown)


davidselassie:
This is going to definitely clash with https://github.com/bytewax/bytewax/pull/115 and https://github.com/bytewax/bytewax/pull/127 so I'm happy to wait until those are merged to figure this out.

davidselassie:
I'm actually going to poke at this a little further because I think persisting awake times per key is pretty closely tied to this. You can review if you want, but I'm going to add some new commits.

whoahbot (approved):

pytests/test_operators.py
```diff
@@ -262,3 +338,137 @@ def running_count(type_to_count, event):
...
+    assert sorted(out) == sorted([])
+
+    # Disable bomb
+    armed.clear()
+    out.clear()
+
+    # Recover
+    run_main(flow, epoch_config=epoch_config, recovery_config=recovery_config)
+
+    # But it remembers the first two items in the first window.
```
blakestier, line 399: Thanks for the comments. So helpful

pytests/test_operators.py
```diff
@@ -262,3 +338,137 @@ def running_count(type_to_count, event):
...
+    assert sorted(out) == sorted([("ALL", 3), ("ALL", 1)])
+
+
+def test_fold_window(recovery_config):
+    flow = Dataflow()
+
+    # Remember that clocks are built per-key so the `TestingClock` in
```
blakestier, line 406: Probably a smell but let's table that for now 🤔

src/operators/fold_window.rs
```diff
@@ -35,14 +37,14 @@ impl FoldWindowLogic {
 }
 
 impl WindowLogic<TdPyAny, TdPyAny, Option<TdPyAny>> for FoldWindowLogic {
-    fn exec(&mut self, next_value: Option<TdPyAny>) -> Option<TdPyAny> {
+    fn with_next(&mut self, next_value: Option<TdPyAny>) -> Option<TdPyAny> {
```
blakestier, line 40: why name this one `with_next` and the other `awake_with`?

src/recovery/mod.rs
```diff
@@ -477,18 +477,39 @@ pub(crate) enum StateUpdate {
     /// Save an updated state snapshot.
     Upsert(StateBytes),
     /// Note that the state for this key was reset to empty.
```
blakestier, line 479: Should this comment change?

src/recovery/mod.rs
```diff
@@ -1254,35 +1274,31 @@ where
...
-
-                    // Try to process all the epochs we have data for.
-                    // Filter out epochs that are ahead of the
-                    // frontier. The state at the beginning of those
-                    // epochs are not truly known yet, so we can't
-                    // apply input in those epochs yet.
-                    activate_epochs_buffer.extend(
+                    assert!(closed_epochs_buffer.is_empty());
+
+                    // On the last activation, we eagerly executed the
```
blakestier, line 1279: This comment confuses me in conjunction with the comment below (on 1296). At this point in the code, have we already executed the frontier? Or are we about to, like in 1298?

src/recovery/mod.rs
```diff
@@ -778,7 +797,7 @@ pub(crate) fn build_state_loading_dataflow<A: Allocate>(
 
                                 match update {
                                     StateUpdate::Upsert(state) => resume_state.insert(key, state),
-                                    StateUpdate::Reset => resume_state.remove(&key),
+                                    StateUpdate::Discard => resume_state.remove(&key),
```
blakestier, line 800: When i read this, I keep expecting the logic to be something like "Is there an updated state? great, add it to the stateful collection. Otherwise, discard it". Otherwise, I'm thinking there would be a 3rd possibility of no state update? I think it's the `Update` that's confusing me. Like it's either a `State::Upsert(state)` or a `State::Discard` but maybe the word `State` is too loaded to stand alone

src/window/tumbling_window.rs
```diff
@@ -131,16 +133,17 @@ impl Windower for TumblingWindower {
...
 
-    fn activate_after(&mut self, watermark: &NaiveDateTime) -> Option<Duration> {
-        let watermark = *watermark;
-        self.close_times
-            .values()
-            .cloned()
-            .map(|t| t - watermark)
-            .min()
+    fn fate(&self) -> WindowerFate {
+        if let Some(next_close) = self.close_times.values().cloned().min() {
```
blakestier, line 137: why did i think these were sorted already? 🤔

blakestier (approved):
I left some comments around names to highlight places that felt a little sticky for me, but I think this is a great change and I appreciate the tests, the untangling, and the dramatic introduction of `fate`

src/operators/fold_window.rs
davidselassie: Mmmm. Maybe that's more misleading then helpful. I was trying to remind us that the contexts in which these are called by the framework are slightly different:

- `StatefulLogic::exec` could happen at any old random time (although probably just your scheduled time) and you might not even have new data (but also probably do)! This is a random awakening in the night and you need to be prepared to do nothing useful.

- `WindowLogic::exec` specifically only happens when the window has new data (or we know the data stream is permanently over). This is like the "inverse"? the "b-side"? of `Iterator::next` -- you will be called with the result.

Yeah, the types `Poll<Option<V>>` vs `Option<V>` tell this to you as well, but I felt like the name `exec` was too generic anyway.

Something like `awaken` feels very appropriate for the `StatefulLogic::exec` because of the timeout dimension, but it feels not appropriate for the `WindowLogic::exec` because time is irrelevant and the iterator part seems the most important. (I couldn't find any prior art for this sort of iterator inversion of control in the std lib, otherwise I would have stolen that name.)

Perhaps `StatefulLogic::awake_with_next` and `StatefulLogic::with_next` to show the inner part is related, but stateful logic has that other timeout component?

pytests/test_operators.py
davidselassie: This is sort of changed in https://github.com/bytewax/bytewax/pull/139.

Not the per-key thing, but it'll be less confusing? Although I'm having some trouble with the `TestingClockConfig` in another PR, so the smell does remain...

src/recovery/mod.rs
davidselassie: Possibly both! If the frontier hasn't changed between activations, we keep running it multiple times merging in new data in that singular frontier epoch until it is closed.

E.g: We have a single stateful operator. It has a single input.

1. Wake up. Poll the input. It returns some data in epoch 4, 5, 6. The input's frontier is epoch 6. That means epoch 6 is _not_ closed. But 4,5 < 6 so they are closed so process them in order. Also process 6! 6 isn't closed but by definition we won't see data before it ever again, so we can partially complete processing on it. Go to sleep.
2. Wake up. Poll the input. It returns more data in epoch 6. The input's frontier is still 6. Process 6 again; not because it is closed, but because we're still "in order" and we can partially complete work. Go to sleep.

So in that case we eagerly executed the same frontier on both awakenings.

---

But also the frontier might have changed between activations. In that case, the answer to the question is still "both" but we already executed the old frontier and we're about to execute the new frontier. The exact epoch of "frontier" has changed.

---

Perhaps it would be clarifying here to put "the frontier on the last activation" and in the next comment to put "the current frontier" (and those might be the same and that's ok).

src/window/tumbling_window.rs
davidselassie: Maybe you're remembering that we use `BTreeMap` elsewhere? We can't here because we need to order by value, not key. We could cache the min close time, but I don't know that it would help that much.

src/recovery/mod.rs
davidselassie: FWIW on the reading side (since this linked snippet is the loading during recovery), that situation is already handled: if we didn't write the update, then there wasn't an update, so don't do anything.

src/operators/fold_window.rs
blakestier: `exec_with` and `awake_with`? This explanation is helpful-- not too fussed about it, and i think the "sometimes there's a thing with the next" is the part that gets sticky for me.

appreciate the walkthrough

pytests/test_operators.py
davidselassie: I figured out what was confusing me. It's that we were recovering and using a half-consumed generator. https://github.com/bytewax/bytewax/pull/144 has the tests out of this working.

davidselassie:
Rebased this. 

The changes from the https://github.com/bytewax/bytewax/pull/143 and https://github.com/bytewax/bytewax/pull/144 are thus no longer in this PR. 

Broke out `LogicFate::AwakeAfter` into its own function `StatefulLogic::next_awake` so that the "do I keep this around" question is totally separate from the "when do I wake up next" question.

In debugging that I think we're going to have to take an overhaul of how system time works: currently because there are interactions between the times returned by the clock in the window operators and the system time used by the Timely scheduler, we can't really deterministically unit test system time still. This is tough because the behavior _is_ determined by when some code runs / windows close. I think this looks like using the `Clock` in `StatefulUnary` for run time and assuming the Timely scheduler activation delays will match up, but I'm not sure.

PR #139, opened 2022-09-21
Allow manually adjusting `TestingClockConfig`

davidselassie:
For more complicated windowing tests, we need more control over the
way that the `TestingClock` determines the current time. Instead of
incrementing after each item, allow you to modify the current time
Python-side.

Has the `python.window.TestingClockConfig` take a wrapper Python class
called `bytewax.testing.TestingClock` which lets you modify a
`datetime.datetime` however you need after each item of input.

Modifies our two tests to demonstrate this.

### DIFF
--- a/pysrc/bytewax/testing.py
+++ b/pysrc/bytewax/testing.py
@@ -4,6 +4,8 @@ import multiprocessing.dummy
 from contextlib import contextmanager
 from threading import Lock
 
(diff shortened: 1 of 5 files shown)


whoahbot (approved):

Psykopear (approved):

Psykopear:
Yes, this makes testing much more clear

PR #142, opened 2022-09-23
Fixed examples for dataflow reduce and fold window.

PiePra:
Stumbled upon some non working code examples in docs and a small typo.

### DIFF
--- a/apidocs/html/bytewax/dataflow.html
+++ b/apidocs/html/bytewax/dataflow.html
@@ -144,7 +144,7 @@ world
 window into an accumulator, using a function to build its initial value.</p>
 <p>It is like <code>bytewax.Dataflow.reduce_window()</code> but uses a function to
 build the initial value.</p>
-<p>It is a stateful operator. It requires the the input stream
+<p>It is a stateful operator. It requires the input stream
 has items that are <code>(key: str, value)</code> tuples so we can ensure
 that all relevant values are routed to the relevant state. It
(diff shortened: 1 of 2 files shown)


apidocs/html/bytewax/dataflow.html
```diff
@@ -179,7 +179,7 @@ <h2 id="args">Args</h2>
 &gt;&gt;&gt; def count(results, event):
 ...     results[event[&quot;type&quot;]] += 1
 ...     return results
-&gt;&gt;&gt; clock_config = TestingClockConfig(item_incr=timedelta(seconds=4))
+&gt;&gt;&gt; clock_config = TestingClockConfig(start_at=datetime(2022, 1, 1, 13), item_incr=timedelta(seconds=4))
```
whoahbot, line 182: These docs are auto-generated from the source files in [src/dataflow/mod.rs](https://github.com/bytewax/bytewax/blob/main/src/dataflow/mod.rs#L407), if you make the changes there, we will re-generate the html documentation when making our next release.

whoahbot (changes requested):
Thanks so much for taking the time to fix these! I left you some notes on how to make your changes are permanent.

apidocs/html/bytewax/dataflow.html
PiePra: Made the changes in mod.rs within the second commit.

whoahbot (approved):

PR #143, opened 2022-09-26
Fixes recovery serde

davidselassie:
Makes all types for recovery serialization and deserialization
explicit and fixes them to match each other for each recoverable
operator.

This will fix panics during recovery loading.

This was also fixed in https://github.com/bytewax/bytewax/pull/137/
but I'm breaking it out here separately because I keep discoverying
smaller recovery bugs and am fixing them separately.

### DIFF
--- a/src/operators/reduce_window.rs
+++ b/src/operators/reduce_window.rs
@@ -20,7 +20,9 @@ pub(crate) struct ReduceWindowLogic {
 impl ReduceWindowLogic {
     pub(crate) fn builder(reducer: TdPyCallable) -> impl Fn(Option<StateBytes>) -> Self {
         move |resume_acc_bytes| {
-            let acc = resume_acc_bytes.map(|resume_acc_bytes| resume_acc_bytes.de());
+            let acc = resume_acc_bytes
+                .map(StateBytes::de::<Option<TdPyAny>>)
+                .flatten();
             Python::with_gil(|py| Self {
                 reducer: reducer.clone_ref(py),
                 acc,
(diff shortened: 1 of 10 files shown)


src/operators/reduce_window.rs
```diff
@@ -20,7 +20,9 @@ pub(crate) struct ReduceWindowLogic {
 impl ReduceWindowLogic {
     pub(crate) fn builder(reducer: TdPyCallable) -> impl Fn(Option<StateBytes>) -> Self {
         move |resume_acc_bytes| {
-            let acc = resume_acc_bytes.map(|resume_acc_bytes| resume_acc_bytes.de());
+            let acc = resume_acc_bytes
+                .map(StateBytes::de::<Option<TdPyAny>>)
+                .flatten();
```
whoahbot, line 25: Why wasn't the `flatten()` call needed before?

whoahbot (approved):
:+1: Just left you a question for my own curiosity.

src/operators/reduce_window.rs
davidselassie: It was actually needed, hence the panic, but type inference made the compiler assume the wrong types since it can't know what's in those bytes.

The previous version without the flatten was causing the compiler to infer the incorrect type for the deserialization step. So for reduce window, `snapshot` was serializing with `StateBytes::ser(&self.acc)` which was inferred (correctly) to be `StateBytes::ser::<Option<TdPyAny>>(&self.acc)` _but_ the builder was deserializing with `resume_acc_bytes.map(|resume_acc_bytes| resume_acc_bytes.de())` which was inferred to be `resume_acc_bytes.map(|resume_acc_bytes| resume_acc_bytes.de::<TdPyAny>())` but by (bad) luck the final type of `acc` was correct, so it was like cool this must be correct.

By making the deserialization type explicit and match the serialization type, now if you forget the `flatten` it's a type error.

PR #144, opened 2022-09-26
Better recovery tests

davidselassie:
As part of testing the functionality of recoverable operators, we
should actually do a recovery to ensure serde is correct. This
modifies the operator tests to do that (and moves an operator test out
of `test_window` into `test_operator`).

When using `TestingClockConfig`, you have to use a generator to ensure
the input code is modifying the state of the global clock. But! We
need to make sure that the generator is re-initialized upon the second
startup of the dataflow. Otherwise, since the test is happening with a
single `Dataflow` instance and the process wasn't restarted, the
generator is half-consumed upon recovery loading and you don't get the
correct behavior. To facilitate this, adds a
`TestingBuilderInputConfig` which re-builds the generator on each
start so it'll be re-run from top to bottom.

Also fixes some minor docstring issues.

And moves `ManualInputConfig` to the top of the module source to match
all the other PyO3 containing files.

### DIFF
--- a/pysrc/bytewax/inputs.py
+++ b/pysrc/bytewax/inputs.py
@@ -7,19 +7,19 @@ Use
 ---
 
 Create an `InputConfig` subclass for the source you'd like to read
-from. Then pass that config object to the `bytewax.dataflow.Dataflow`
-constructor.
+from. Then pass that config object as the `input_config` argument of
+the `bytewax.dataflow.Dataflow.input` operator.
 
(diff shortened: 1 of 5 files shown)


davidselassie:
This is just the unit test changes out of https://github.com/bytewax/bytewax/pull/137 but separate, and actually working.

whoahbot (approved):
LGTM

PR #146, opened 2022-10-03
Event time in tests

### LINKED ISSUE #145: Convert window tests to use event time
### Bug Description

As alluded to in this comment https://github.com/bytewax/bytewax/pull/137#issuecomment-1260093091, system time unit tests are very finicky to setup and aren't working yet for two reasons:

1. Part of system time windowing is operators scheduling wake ups. The order of those wake ups affects behavior (that's ok, because the desired behavior is based on time). We currently mock out `Clock` in the `WindowLogic` via `TestingClock` to allow us to arbitrarily change the time in a deterministic manner, _but_ we don't do the same for the time internally to `StatefulUnary` used to schedule the wake ups. Thus when Timely decides to wake up the operator and we decide "wakeup time is past: run the logic!" is based on an inconsistent sense of system time compared to the windowing system.

2. Even if we could synchronize both times to a `TestingClock`, the deterministic modification of that clock via `clock.now = whatever` is very challenging to schedule at just the right moment. Timely is a dataflow framework, so if you try to do something out-of-band (like modifying a shared global state of the current time), you have to really understand the behavior of the Timely scheduler and activators to know when code will run. This means that unit tests that change the clock will have to rely on very specific activation orders and are thus hard to write.

For example, I wrote [the following test](https://github.com/bytewax/bytewax/blob/8da9636cd498f19f0d8d492f71d14facbd410889/pytests/test_operators.py#L416-L497) with the following input:

```py
    def gen():
        clock.now = start_at  # +0 sec; reset on recover
        yield {"user": "a", "type": "login"}  # Epoch 0
        clock.now += timedelta(seconds=4)  # +4 sec
        yield {"user": "a", "type": "post"}  # Epoch 1
        clock.now += timedelta(seconds=4)  # +8 sec
        yield {"user": "a", "type": "post"}  # Epoch 2, Instant A
        clock.now += timedelta(seconds=4)  # +12 sec
        # First 10 sec window closes during processing this input.
        # Instant B
        yield {"user": "b", "type": "login"}  # Epoch 3, 
        yield "BOOM"  # Epoch 4
        clock.now += timedelta(seconds=4)  # +16 sec
        yield {"user": "a", "type": "post"}
        clock.now += timedelta(seconds=4)  # +20 sec
        # Second 10 sec window closes during processing this input.
        yield {"user": "b", "type": "post"}
        clock.now += timedelta(seconds=4)  # +24 sec
        yield {"user": "b", "type": "post"}
        clock.now += timedelta(seconds=4)  # +28 sec
```

Note that instant A and instant B are different times, and Timely could decide to run operators in these interleaving spaces. And since the order of events affects system time, you need to know the intricacies of the scheduler to make this test work.

Personally, I think the best way to deal with this is to minimize reliance on system time in unit tests. We should write as many of our unit tests as possible using just event time and infinite late windows (since that also uses system time), since that means that the ordering of the input is the only determinant of the output. In very specific instances where we need to test a system time thing, we should be careful about it and perhaps find a way to test the component Rust-side at a lower level.

Python 3.10.4

bytewax 0.11.2

### Relevant log output

_No response_

### Steps to Reproduce

See example above. Or run the test in https://github.com/bytewax/bytewax/blob/8da9636cd498f19f0d8d492f71d14facbd410889/pytests/test_operators.py#L416-L497.

Psykopear:
Closes #145 

Use `EventTimeClock` in tests, remove the now unused `TestingClock`, also remove the ability to mock `system_clock` in the `EventTimeClock` since it was causing more problems than it solved.

### Original description, ignore this
The `TestingClock` now behaves similarly to the `EventTimeClock`, but with either an "infinite" lateness (actually 1000 years), or a zero one.

In the context of a windowing operator, this means to either wait until the end of the stream before closing any window, or closing any window as soon as possible (as as soon as an event with a timestamp of a later window arrives).
The second option was added to test partial results before a crash/recovery (like in `test_reduce_window`).

Since we don't want to allow the previous approach of mocking the `EventTimeClock` in tests, I also removed the api for that, which made the clock itself simpler.


pytests/test_operators.py
```diff
@@ -379,24 +368,31 @@ def trigger(item):
 
     flow.flat_map(trigger)
 
-    clock_config = TestingClockConfig(clock)
+    clock_config = TestingClockConfig(lambda e: e["time"], wait_until_end=False)
```
davidselassie, line 371: What is your thinking behind using the `TestingClockConfig` here rather than the `EventTimeClockConfig`? My thinking was that the operator functionality can still be tested using event time and so we should use that clock directly to make it clear that is what's happending. We'll save the `TestingClock` as is for very special (and confusing) cases where we need to mock system time, which I don't think we have to do for any of these tests.

pytests/test_operators.py
```diff
@@ -456,16 +443,17 @@ def trigger(item):
     flow.flat_map(trigger)
 
     def key_off_user(event):
-        return (event["user"], event["type"])
+        return (event["user"], event)
 
     flow.map(key_off_user)
 
-    clock_config = TestingClockConfig(clock)
+    clock_config = TestingClockConfig(lambda e: e["time"], wait_until_end=False)
```
davidselassie, line 450: Same comment r.e. `TestingClockConfig` vs `EventTimeClockConfig`.

pytests/test_window.py
```diff
@@ -1,46 +1,37 @@
...
         # First 10 second window should close just before processing this item.
-        yield ("ALL", 1)
-        clock.now += timedelta(seconds=4)  # +16 sec
+        yield ("ALL", {"time": start_at + timedelta(seconds=12), "val": 1})
+        yield ("ALL", {"time": start_at + timedelta(seconds=16), "val": 1})
 
     flow.input("inp", TestingBuilderInputConfig(gen))
 
-    clock_config = TestingClockConfig(clock)
+    clock_config = TestingClockConfig(lambda e: e["time"], wait_until_end=True)
```
davidselassie, line 25: Same comment r.e. `TestingClockConfig` vs `EventTimeClockConfig`.

pytests/test_operators.py
Psykopear: So, the `TestingClockConfig` is technically an `EventClockConfig` with a fixed configuration.
Since we are not testing the clock, but the operators (and/or recovery), we do this by using a clock that can represent the two use cases we have in tests:
- We want to wait untill the end of the input before emitting anything
- We want to emit anything as soon as possible (to test for recovery)

We could do that with a normal `EventclockConfig`, but then we would have to be aware of never using it with a `wait_for_system_duration` that is not `0` or `longer than the timespan of all event times`.

So, I removed the old TestingClock since the idea of this PR was to move all tests to use an event time clock, and created a specific event time clock that only allows the behaviours we are currently using in tests.

This way we avoid giving the possibility of partially mocking system time (which creates more problems than it solves), and we can use event time without worrying about flaky behaviour due to system timing.

src/window/testing_clock.rs
```diff
@@ -5,118 +5,48 @@ use pyo3::{exceptions::PyValueError, prelude::*};
...
+/// If wait_until_end is False, windows are closed as soon as an event
+/// of a later window is received.
 ///
 /// Args:
 ///
-///   clock (TestingClock): Query this `TestingClock` for the current
-///       "now".
+///  dt_getter: function used to retrieve a UTC datetime from each event.
+///
+///  wait_until_end: wether to wait for all data to be collected, or
```
blakestier, line 22: When I read this comment, I wonder how we know "all" data has been collected.

I'm also wondering if it makes sense to default this to `false` and only explicitly set it for cases where we want to allow infinite lateness, since it seems like default behavior is "close the window when i say the window is closed (i.e. after 10 min).

Also, maybe that points to renaming this, since it's about waiting for all data, or allowing a wide wait band for lateness.

src/window/event_time_clock.rs
```diff
@@ -29,43 +25,30 @@ use super::{
...
 ///
-///   system_clock: Optional configuration to use a PyTestingClock as the internal system_clock,
-///                 only used for testing purposes right now.
-///
 /// Returns:
 ///   Config object. Pass this as the `clock_config` parameter to your
 ///   windowing operator.
 #[pyclass(module="bytewax.window", extends=ClockConfig)]
-#[pyo3(text_signature = "(dt_getter, wait_for_system_duration, system_clock)")]
+#[pyo3(text_signature = "(dt_getter, wait_for_system_duration)")]
```
blakestier, line 32: Removing that clock feels right 🎉

davidselassie:
Correct. I don't think we need `TestingClock` at all if we're not using it. Once we have a better sense of where are the spots we're going to use system time, we could perhaps resurrect it. Maybe with a clearer name too.

I'll merge #147 shortly.

davidselassie (approved):

PR #147, opened 2022-10-03
Persist awake times

davidselassie:
Prep
----

This turned into a sort of refresh of types used in the recovery
system to better match the model of what is actually happening.

The recovery store is modeled as two key-value DB tables, progress and
state, and the stateful operators generate CDC-style updates which are
applied to the tables. This PR renames the structs involved to match
that process. There's `ProgressUpdate` and `StateUpdate` which
correspond to KV changes. And each is a 2-tuple struct with 2 fields:
`ProgressRecoveryKey` and `ProgressOp` or `StateRecoveryKey` and
`StateOp`, with the `Op` enums representing what kind of change
happened to the value for each key (e.g. update or delete). Within
each `Op` is a `Progress` or `State` type that actually captures the
new value.

Stateful operators now emit fully formed `StateUpdate`s instead of the
difficult-to-name `EpochData`, which was the `(step_id, (state_key,
update))` tuple. This makes a nice symmetry between the backup phase
and the recovery loading, since everything uses the same types.

A lot of systemic renaming from `backup` to `update`, `update` to
`op`, and `state_bytes` to `state` to match this new convention.

Next Awake
----------

Now that we have a `State` type, we can add the `next_awake:
DateTime<Utc>` to that struct so we will save the next awake time for
each `StateKey`. Fills out that whenever we do a snapshot of the state
data in `StatefulUnary`.

Since all fields of `State` need to be written to each recovery store,
adds the `chrono` feature of `sqlx` to enable writing `DateTime` as a
row value. Updates the schema of the state table to add a `next_awake`
column.

Progress
--------

Makes both the SQLite and Kafka recovery stores actually store
progress data under the `ProgressRecoveryKey` to keep the parallel
construction between the state and progress recovery
tables. Previously, we were using the fixed string `"worker_frontier"`
which produces the same behavior when cluster size is static across
recovery, but this will set us up for our rescaling future where data
from multiple workers is mixed together. Instead the worker frontier
is one of the fields in the value (the only one as of now).

Adds a little `WorkerIndex` newtype to label when an int is meant to
be interpreted as the worker number.

As part of clarifying the datatypes needed to store progress, we don't
store the full antichain in the recovery store anymore; instead we
collapse the antichain into the minimum value during writing and only
store that single timestamp as the frontier. This helps with debugging
as you can now store the worker frontier as a single int, which is
much easier to parse out of the recovery store by hand rather than a
serde blob of bytes.

Reduces the logging level of recovery stores down to trace.

### DIFF
--- a/src/execution/mod.rs
+++ b/src/execution/mod.rs
@@ -39,7 +39,7 @@ use crate::operators::*;
 use crate::outputs::build_output_writer;
 use crate::outputs::capture;
 use crate::pyo3_extensions::{extract_state_pair, wrap_state_pair, TdPyAny};
-use crate::recovery::StatefulUnary;
+use crate::recovery::StateReader;
 use crate::recovery::WriteProgress;
 use crate::recovery::WriteState;
 use crate::recovery::{
@@ -52,7 +52,7 @@ use crate::recovery::{ProgressReader, StateCollector};
 use crate::recovery::{RecoveryConfig, StepId};
 use crate::recovery::{RecoveryStoreSummary, StateWriter};
 use crate::recovery::{StateBytes, StateKey};
-use crate::recovery::{StateReader, StateUpdate};
+use crate::recovery::{StateUpdateStream, StatefulUnary};
 use crate::window::{build_clock_builder, build_windower_builder, StatefulWindowUnary};
 use crate::StringResult;
 use log::debug;
@@ -177,7 +177,7 @@ fn build_source<S>(
     reader: Box<dyn InputReader<TdPyAny>>,
     start_at: S::Timestamp,
     probe: &ProbeHandle<S::Timestamp>,
-) -> StringResult<(Stream<S, TdPyAny>, Stream<S, StateUpdate<S::Timestamp>>)>
+) -> StringResult<(Stream<S, TdPyAny>, StateUpdateStream<S>)>
 where
     S: Scope<Timestamp = u64>,
 {
--- a/src/execution/mod.rs
+++ b/src/execution/mod.rs
@@ -523,11 +523,7 @@ fn build_and_run_state_loading_dataflow<A: Allocate>(
 
     run_until_done(worker, interrupt_flag, probe);
 
-    let mut step_to_key_to_resume_state = HashMap::new();
-    while let Ok((step_id, key_to_resume_state_bytes)) = step_to_key_to_resume_state_rx.recv() {
-        step_to_key_to_resume_state.insert(step_id, key_to_resume_state_bytes);
-    }
-
+    let step_to_key_to_resume_state = step_to_key_to_resume_state_rx.into_iter().collect();
     let recovery_store_summary = recovery_store_summary_rx
         .recv()
         .expect("Recovery store summary not returned from loading dataflow");
--- a/src/execution/mod.rs
+++ b/src/execution/mod.rs
@@ -224,7 +224,7 @@ fn build_production_dataflow<A: Allocate>(
     worker: &mut Worker<A>,
     epoch_config: Py<EpochConfig>,
     resume_epoch: u64,
-    mut step_to_key_to_resume_state: HashMap<StepId, HashMap<StateKey, State>>,
+    mut resume_state: HashMap<StepId, HashMap<StateKey, State>>,
     recovery_store_summary: RecoveryStoreSummary<u64>,
     flow: Py<Dataflow>,
     progress_writer: Box<dyn ProgressWriter<u64>>,
@@ -258,11 +258,9 @@ fn build_production_dataflow<A: Allocate>(
                     step_id,
                     input_config,
                 } => {
-                    let (resume_state, recovery_key) = resume_input_state(
+                    let (step_resume_state, recovery_key) = resume_input_state(
                         worker_index,
-                        step_to_key_to_resume_state
-                            .remove(&step_id)
-                            .unwrap_or_default(),
+                        resume_state.remove(&step_id).unwrap_or_default(),
                     );
 
                     let input_reader = build_input_reader(
@@ -270,7 +268,7 @@ fn build_production_dataflow<A: Allocate>(
                         input_config,
                         worker_index,
                         worker_count,
-                        resume_state,
+                        step_resume_state,
                     )?;
                     let (downstream, update_stream) = build_source(
                         py,
@@ -302,9 +300,7 @@ fn build_production_dataflow<A: Allocate>(
                     builder,
                     folder,
                 } => {
-                    let key_to_resume_state = step_to_key_to_resume_state
-                        .remove(&step_id)
-                        .unwrap_or_default();
+                    let step_resume_state = resume_state.remove(&step_id).unwrap_or_default();
 
                     let clock_builder = build_clock_builder(py, clock_config)?;
                     let windower_builder = build_windower_builder(py, window_config)?;
@@ -315,7 +311,7 @@ fn build_production_dataflow<A: Allocate>(
                             clock_builder,
                             windower_builder,
                             FoldWindowLogic::new(builder, folder),
-                            key_to_resume_state,
+                            step_resume_state,
                         );
 
                     stream = downstream
@@ -342,15 +338,13 @@ fn build_production_dataflow<A: Allocate>(
                     reducer,
                     is_complete,
                 } => {
-                    let key_to_resume_state_bytes = step_to_key_to_resume_state
-                        .remove(&step_id)
-                        .unwrap_or_default();
+                    let step_resume_state = resume_state.remove(&step_id).unwrap_or_default();
 
                     let (downstream, update_stream) =
                         stream.map(extract_state_pair).stateful_unary(
                             step_id,
                             ReduceLogic::builder(reducer, is_complete),
-                            key_to_resume_state_bytes,
+                            step_resume_state,
                         );
                     stream = downstream.map(wrap_state_pair);
                     state_update_streams.push(update_stream);
@@ -361,9 +355,7 @@ fn build_production_dataflow<A: Allocate>(
                     window_config,
                     reducer,
                 } => {
-                    let key_to_resume_state_bytes = step_to_key_to_resume_state
-                        .remove(&step_id)
-                        .unwrap_or_default();
+                    let step_resume_state = resume_state.remove(&step_id).unwrap_or_default();
 
                     let clock_builder = build_clock_builder(py, clock_config)?;
                     let windower_builder = build_windower_builder(py, window_config)?;
@@ -374,7 +366,7 @@ fn build_production_dataflow<A: Allocate>(
                             clock_builder,
                             windower_builder,
                             ReduceWindowLogic::builder(reducer),
-                            key_to_resume_state_bytes,
+                            step_resume_state,
                         );
 
                     stream = downstream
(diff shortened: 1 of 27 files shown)


src/recovery/kafka.rs
```diff
@@ -169,6 +172,25 @@ pub(crate) fn create_kafka_topic(brokers: &[String], topic: &str, partitions: i3
     }
 }
 
+fn to_bytes<T>(obj: &T) -> Vec<u8>
```
davidselassie, line 175: These are now only needed for Kafka recovery, since SQLite recovery has a reasonable type for each column.

src/recovery/mod.rs
```diff
@@ -467,213 +478,205 @@ impl StateBytes {
...
 }
 
-/// Impl this trait to create an operator which maintains recoverable
-/// state.
-///
-/// Pass a builder of this to [`StatefulUnary::stateful_unary`] to
-/// create the Timely operator. A separate instance of this will be
-/// created for each key in the input stream. There is no way to
-/// interact across keys.
-pub(crate) trait StatefulLogic<V, R, I>
```
davidselassie, line 509: This trait got moved down just before `StatefulUnary`, but is unchanged.

whoahbot (approved):

Psykopear (commented):
The PR looks good, I just added a couple of non blocking comments about structure/variables names.

The `execution` and `recovery` module feel hard to read, and trying to understand why, I noticed that a lot of times the same structure has different variable name, complex types are passed down through different functions, and the purpose of the types is not clear unless you have been following the code all along.

For example `let step_to_key_to_resume_state: HashMap<StepId, HashMap<StateKey, State>> = HashMap::new()` or similar lines, make the shape of the data clear, but not what it is used for.
This PR also shows that the shape of the data is leaking into variable names too, and changing the shape keeping the same meaning should not require us to change the name of all the variables.

For this reason I think using a newtype to represent our `state cache` and `state cache per step`, and renaming the variables to avoid leaking implementation details in the name would help improve readability.

A similar point for the streams. We reuse some stream shapes in several position, and I feel that giving them a name representing what we are using the stream for would help for the same reasons. The names given to variables/arguments/methods in those cases are already good, so I used that to find a name for each kind of stream.

If this is out of scope for this PR though, we can discuss those changes elsewhere.

Anyway I'd wait for this to be merged before merging #146 since this solves a bug I faced there

src/execution/mod.rs
```diff
@@ -130,16 +131,40 @@ pub(crate) fn default_epoch_config() -> Py<EpochConfig> {
     })
 }
 
+/// Integer representing the index of a worker in a cluster.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Deserialize)]
+pub(crate) struct WorkerIndex(pub(crate) usize);
+
+impl WorkerIndex {
+    /// Tell Timely to route to this worker.
```
blakestier, line 139: Should we test it with some debuggery?

src/execution/mod.rs
```diff
@@ -130,16 +131,40 @@ pub(crate) fn default_epoch_config() -> Py<EpochConfig> {
...
+        self.0.into_py(py)
+    }
+}
+
 /// Generate the [`StateKey`] that represents this worker and
 /// determine if there's any resume state.
 fn resume_input_state(
-    worker_index: usize,
+    worker_index: WorkerIndex,
     _worker_count: usize,
```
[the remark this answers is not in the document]
davidselassie: Yeah, I guess not at this point. I think an old version of the stateful input code had this function actually doing the input partitioning per worker, and so needed to know both worker index and count to do the partitioning correctly.

I can take it out.

src/execution/mod.rs
```diff
@@ -481,34 +506,33 @@ fn build_and_run_state_loading_dataflow<A: Allocate>(
...
     run_until_done(worker, interrupt_flag, probe);
 
-    let mut step_to_key_to_resume_state_bytes = HashMap::new();
-    while let Ok((step_id, key_to_resume_state_bytes)) = step_to_key_to_resume_state_bytes_rx.recv()
-    {
-        step_to_key_to_resume_state_bytes.insert(step_id, key_to_resume_state_bytes);
+    let mut step_to_key_to_resume_state = HashMap::new();
+    while let Ok((step_id, key_to_resume_state_bytes)) = step_to_key_to_resume_state_rx.recv() {
+        step_to_key_to_resume_state.insert(step_id, key_to_resume_state_bytes);
     }
```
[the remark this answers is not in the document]
davidselassie: Oh, great!

src/execution/mod.rs
davidselassie: Good idea.

src/execution/mod.rs
```diff
@@ -200,14 +225,14 @@ fn build_production_dataflow<A: Allocate>(
     worker: &mut Worker<A>,
     epoch_config: Py<EpochConfig>,
     resume_epoch: u64,
-    mut step_to_key_to_resume_state_bytes: HashMap<StepId, HashMap<StateKey, StateBytes>>,
+    mut step_to_key_to_resume_state: HashMap<StepId, HashMap<StateKey, State>>,
```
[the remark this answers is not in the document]
davidselassie: Good point: the types are there. I tried to take a pass doing a more semantic rename. I think the big things were to distinguish "resume state" from "current state". Hopefully this helps.

src/execution/mod.rs
```diff
@@ -153,7 +178,7 @@ fn build_source<S>(
     reader: Box<dyn InputReader<TdPyAny>>,
     start_at: S::Timestamp,
     probe: &ProbeHandle<S::Timestamp>,
-) -> StringResult<(Stream<S, TdPyAny>, Stream<S, EpochData>)>
+) -> StringResult<(Stream<S, TdPyAny>, Stream<S, StateUpdate<S::Timestamp>>)>
```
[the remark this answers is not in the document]
davidselassie: Great idea. I think this makes sense for `StateUpdateStream`, `ProgressUpdateStream`, and `StatefulStream`. But I think we need to still wait and see how windowing and errors work out, and `Stream<S, TdPyAny>` we only end up using in one place explicitly.

src/execution/mod.rs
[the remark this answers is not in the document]
davidselassie: I didn't end up using the type aliases, since there are so many variations on storing state that don't compound easily. Let's see how the naming changes feel.

src/operators/fold_window.rs
```diff
@@ -23,8 +23,8 @@ impl FoldWindowLogic {
         builder: TdPyCallable,
         folder: TdPyCallable,
     ) -> impl Fn(Option<StateBytes>) -> Self {
-        move |resume_acc_bytes| {
-            let acc = resume_acc_bytes
+        move |resume_snapshot| {
+            let acc = resume_snapshot
                 .map(StateBytes::de::<Option<TdPyAny>>)
                 .flatten();
```
Psykopear, line 29: Suggested by clippy, this could be:
```rust
let acc = resume_snapshot.and_then(StateBytes::de::<Option<TdPyAny>>)
```

src/operators/reduce.rs
```diff
@@ -28,8 +28,8 @@ impl ReduceLogic {
         reducer: TdPyCallable,
         is_complete: TdPyCallable,
     ) -> impl Fn(Option<StateBytes>) -> Self {
-        move |resume_acc_bytes| {
-            let acc = resume_acc_bytes
+        move |resume_snapshot| {
+            let acc = resume_snapshot
                 .map(StateBytes::de::<Option<TdPyAny>>)
                 .flatten();
```
Psykopear, line 34: Same

src/operators/reduce_window.rs
```diff
@@ -19,8 +19,8 @@ pub(crate) struct ReduceWindowLogic {
 
 impl ReduceWindowLogic {
     pub(crate) fn builder(reducer: TdPyCallable) -> impl Fn(Option<StateBytes>) -> Self {
-        move |resume_acc_bytes| {
-            let acc = resume_acc_bytes
+        move |resume_snapshot| {
+            let acc = resume_snapshot
                 .map(StateBytes::de::<Option<TdPyAny>>)
                 .flatten();
```
Psykopear, line 25: Same

src/recovery/mod.rs
```diff
@@ -1041,22 +1089,24 @@ impl<S: Scope> CollectGarbage<S> for Stream<S, StateBackup<S::Timestamp>> {
...
-                        recovery_store_summary.insert(state_backup);
+                    for state_backup in tmp_incoming_state.drain(..) {
+                        recovery_store_summary.insert(&state_backup);
                     }
                 });
 
                 // Tell the notificator to trigger on dataflow
                 // frontier advance.
-                dataflow_frontier_input.for_each(|cap, data| {
+                dataflow_frontier_input.for_each(|cap, incomfing| {
```
Psykopear, line 1104: incoming?

src/recovery/mod.rs
```diff
@@ -1135,44 +1250,41 @@ where
...
-        L: StatefulLogic<V, R, I> + 'static,       // Logic
-        LB: Fn(Option<StateBytes>) -> L + 'static, // Logic builder
-    >(
+    fn stateful_unary<R, I, L, LB>(
         &self,
         step_id: StepId,
         logic_builder: LB,
-        resume_state: HashMap<StateKey, StateBytes>,
-    ) -> (Stream<S, (StateKey, R)>, Stream<S, EpochData>)
+        key_to_resume_state: HashMap<StateKey, State>,
```
Psykopear, line 1257: `resume_state` here too I think

src/recovery/mod.rs
```diff
@@ -1195,12 +1307,30 @@ where
...
+                current_logic.insert(key.clone(), logic_builder(Some(snapshot)));
+                match next_awake {
+                    Some(next_awake) => {
+                        current_next_awake.insert(key.clone(), next_awake);
+                    }
+                    None => {
+                        current_next_awake.remove(&key);
+                    }
+                }
+            }
```
Psykopear, line 1333: may I suggest this?
```rust
            let mut current_logic = HashMap::new();
            let mut current_next_awake = HashMap::new();

            for (key, state) in state_cache.into_iter() {
                current_logic.insert(key.clone(), logic_builder(Some(state.state_bytes)));

                if let Some(next_awake) = state.next_awake {
                    current_next_awake.insert(key.clone(), next_awake);
                } else {
                    current_next_awake.remove(&key);
                }
            }
```

src/recovery/mod.rs
```diff
@@ -1401,69 +1535,64 @@ where
...
+                                        current_logic.insert(state_key.clone(), logic);
+                                        match next_awake {
+                                            Some(next_awake) => {
+                                                current_next_awake
+                                                    .insert(state_key.clone(), next_awake);
+                                            }
+                                            None => {
+                                                current_next_awake.remove(&state_key);
+                                            }
                                         }
```
Psykopear, line 1564: Same here, the match could be an `if let`, you get one less indentation level:
```rust
                                        if let Some(next_awake) = next_awake {
                                            current_next_awake
                                                .insert(state_key.clone(), next_awake);
                                        } else {
                                            current_next_awake.remove(&state_key);
                                        }
```

Psykopear (approved):
Left some more comments, mostly readability suggestions.

For the new commits, that is exactly what I was thinking of, the naming is much more consistent and meaningful now, thanks!

src/operators/fold_window.rs
davidselassie: Oh! There's `flat_map`. Thanks.

src/recovery/mod.rs
davidselassie: I want to explicitly destructure `State` so that if we add more state fields, we get a compiler error reminding us to unpack the new state into the operator. But I'll make it less of a giant line. And use `if let Some`.

PR #150, opened 2022-10-06
Capture output to DynamoDB

whoahbot:
This PR adds a DynamoDB output adapter, and an example combining that with the new event time window.

- Adds anyhow as a dependency.
- Adds proptest as a dev dependency.

### DIFF
--- a/docs/examples/sensor-data.md
+++ b/docs/examples/sensor-data.md
@@ -5,7 +5,7 @@ In this example, we will walk through processing sensor data from IOT devices us
 
 In this scenario, IOT devices are streaming real-time data that is being written to Kafka. Our goals are to:
 * Read data from Kafka
-* Build up a window of events from each sensor, based on it's id
+* Build up a window of events from each sensor, based on its id
 * Write those updated windows to DynamoDB every 5 seconds
 
 Let's start by setting up Bytewax to read from Kafka. We'll begin by importing some dependencies and instantiating our Bytewax Dataflow.
--- a/docs/examples/sensor-data.md
+++ b/docs/examples/sensor-data.md
@@ -31,7 +31,7 @@ def deserialize(key_bytes__payload_bytes):
 flow.map(deserialize)
 ```
 
-The goal of our example Dataflow is to accumulate sensor data over a defined period of time to produce an average reading. To do that, we'll use one of Bytewax's stateful windowing fold_window`.
+The goal of our example Dataflow is to accumulate sensor data over a defined period of time to produce an average reading. To do that, we'll use one of Bytewax's stateful windowing operators, `fold_window`.
 
 In this example, we want our sensor readings to be grouped together based on the timestamp that the sensor recorded when the data was produced. To do that, we need to define a function that will read that time from the sensor event and convert it to UTC.
 
--- a/docs/examples/sensor-data.md
+++ b/docs/examples/sensor-data.md
@@ -42,7 +42,13 @@ def get_event_time(event):
     return datetime.fromisoformat(event["time"])
 ```
 
-There are two important considerations when using event times to construct windows. The first is that we can't guarantee that sensor events will be read from Kafka in order, and the second is that we can't guarantee that we will have received all of the events that we should consider in a timely fashion. Some sensors could lose their connection to the Internet and send data well after it was recorded. In this way, event time processing is a tradeoff between latency and correctness.
+There are two important considerations when using event times to construct windows:
+
+ Event Orderliness: We cannot guarantee that sensor events will be read from Kafka in order.
+ 
+ Event Lateness: We can't guarantee that we will have received all of the events that we should consider in a timely fashion, since sensors could lose their connection to the Internet and send data well after it was recorded. 
+ 
+ In this way, event time processing is a tradeoff between latency and correctness.
 
 We need a way to tell our window operator how long we are willing to wait for all of the data to arrive before returning windows. To do that, we create an `EventClockConfiguration` that uses our time getting function, and defines how long we are willing to wait, in system time.
 
--- a/docs/examples/sensor-data.md
+++ b/docs/examples/sensor-data.md
@@ -70,7 +70,7 @@ start_at = datetime.now(timezone.utc)
 wc = TumblingWindowConfig(start_at=start_at, length=timedelta(seconds=5))
 ```
 
-Our stateful window operator, `fold_window` takes two functions in it's list of arguments, a builder and a folder. The builder function is called when a new key is encountered for a window, and the folder is called on every item in the window.
+Our stateful window operator, `fold_window` takes two functions in its list of arguments: a builder and a folder. The builder function is called when a new key is encountered for a window, and the folder is called on every item in the window.
 
 In our case, the folder function and the builder function are simple. We would like to accumulate all of the events for a sensor into a single list. Our builder function can be the built-in Python function `list`, and our folder can just call `append` on that list with every event within a window.
 
--- a/docs/examples/sensor-data.md
+++ b/docs/examples/sensor-data.md
@@ -1,0 +1,115 @@
+Analyzing Sensor Data With Kafka and DynamoDB
+================================================
+
+In this example, we will walk through processing sensor data from IOT devices using Bytewax.
+
+In this scenario, IOT devices are streaming real-time data that is being written to Kafka. Our goals are to:
+* Read data from Kafka
+* Build up a window of events from each sensor, based on it's id
+* Write those updated windows to DynamoDB every 5 seconds
+
+Let's start by setting up Bytewax to read from Kafka. We'll begin by importing some dependencies and instantiating our Bytewax Dataflow.
+
+``` python
+import json
+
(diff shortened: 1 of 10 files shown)


docs/examples/sensor-data.md
```diff
@@ -0,0 +1,115 @@
...
+``` python
+def deserialize(key_bytes__payload_bytes):
+    _key_bytes, payload_bytes = key_bytes__payload_bytes
+    payload = json.loads(payload_bytes)
+    return payload["id"], payload
+
+flow.map(deserialize)
+```
+
+The goal of our example Dataflow is to accumulate sensor data over a defined period of time to produce an average reading. To do that, we'll use one of Bytewax's stateful windowing fold_window`.
```
blakestier, line 34:
```suggestion
The goal of our example Dataflow is to accumulate sensor data over a defined period of time to produce an average reading. To do that, we'll use one of Bytewax's stateful windowing operators, `fold_window`.
```

docs/examples/sensor-data.md
```diff
@@ -0,0 +1,115 @@
+Analyzing Sensor Data With Kafka and DynamoDB
+================================================
+
+In this example, we will walk through processing sensor data from IOT devices using Bytewax.
+
+In this scenario, IOT devices are streaming real-time data that is being written to Kafka. Our goals are to:
+* Read data from Kafka
+* Build up a window of events from each sensor, based on it's id
```
blakestier, line 8:
```suggestion
* Build up a window of events from each sensor, based on its id
```

docs/examples/sensor-data.md
```diff
@@ -0,0 +1,115 @@
...
+    _key_bytes, payload_bytes = key_bytes__payload_bytes
+    payload = json.loads(payload_bytes)
+    return payload["id"], payload
+
+flow.map(deserialize)
+```
+
+The goal of our example Dataflow is to accumulate sensor data over a defined period of time to produce an average reading. To do that, we'll use one of Bytewax's stateful windowing fold_window`.
+
+In this example, we want our sensor readings to be grouped together based on the timestamp that the sensor recorded in the data when the data was produced. To do that, we need to define a function that will read that time from the sensor event and convert it to UTC.
```
blakestier, line 36:
```suggestion
In this example, we want our sensor readings to be grouped together based on the timestamp that the sensor recorded when the data was produced. To do that, we need to define a function that will read that time from the sensor event and convert it to UTC.
```

docs/examples/sensor-data.md
```diff
@@ -0,0 +1,115 @@
...
+
+``` python
+from datetime import datetime
+
+def get_event_time(event):
+    return datetime.fromisoformat(event["time"])
+```
+
+There are two important considerations when using event times to construct windows. The first is that we can't guarantee that sensor events will be read from Kafka in order, and the second is that we can't guarantee that we will have received all of the events that we should consider in a timely fashion. Some sensors could lose their connection to the Internet and send data well after it was recorded. In this way, event time processing is a tradeoff between latency and correctness.
```
blakestier, line 45: We should probably pull this into a separate doc, but I think it's worthwhile to call this out even more. I'll post a rough example of what i mean

blakestier, line 45:
```suggestion
There are two important considerations when using event times to construct windows:

 Event Orderliness: We cannot guarantee that sensor events will be read from Kafka in order, so...
 
 Event Lateness: We can't guarantee that we will have received all of the events that we should consider in a timely fashion, since sensors could lose their connection to the Internet and send data well after it was recorded. 
 
 In this way, event time processing is a tradeoff between latency and correctness.
```

Optionally we incorporate how we address ^^ in here

src/outputs/dynamodb_output.rs
```diff
@@ -0,0 +1,308 @@
...
+/// - List (containing the types in this list)
+/// - Dict (where keys are strings, and values are types in this list)
+///
+/// Writes to DynamoDB must be authenticated using AWS credentials.
+/// For more information on specifying your AWS credentials and configuring
+/// a region, see https://docs.aws.amazon.com/sdk-for-rust/latest/dg/credentials.html
+///
+/// Args:
+///
+///   table (str): Name of the table to write records to.
```
davidselassie, line 41: Must the table already exist? I didn't see anything like a "create table" call.

src/outputs/dynamodb_output.rs
```diff
@@ -0,0 +1,308 @@
...
+        Python::with_gil(|py| {
+            let (key, payload): (Option<TdPyAny>, TdPyAny) = py_unwrap!(
+                key_payload.extract(py),
+                format!(
+                    "DynamoDBOutput requires a `(key, payload)` 2-tuple; \
+                    got `{key_payload:?}` instead"
+                )
+            );
+            let p: &PyDict = py_unwrap!(
+                payload.extract(py),
```
davidselassie, line 170: `downcast` because you're requesting a specific Python type?

docs/examples/sensor-data.md
```diff
@@ -0,0 +1,115 @@
...
+
+``` python
+from datetime import timezone
+from bytewax.window import TumblingWindowConfig
+
+start_at = datetime.now(timezone.utc)
+wc = TumblingWindowConfig(start_at=start_at, length=timedelta(seconds=5))
+```
+
+Our stateful window operator, `fold_window` takes two functions in it's list of arguments, a builder and a folder. The builder function is called when a new key is encountered for a window, and the folder is called on every item in the window.
```
blakestier, line 67:
```suggestion
Our stateful window operator, `fold_window` takes two functions in its list of arguments: a builder and a folder. The builder function is called when a new key is encountered for a window, and the folder is called on every item in the window.
```

docs/examples/sensor-data.md
```diff
@@ -0,0 +1,115 @@
...
+
+In our case, the folder function and the builder function are simple. We would like to accumulate all of the events for a sensor into a single list. Our builder function can be the built-in Python function `list`, and our folder can just call `append` on that list with every event within a window.
```
blakestier, line 69:
```suggestion
In our case, the folder function and the builder function are straightforward. We would like to accumulate all of the events for a sensor into a single list. Our builder function can be the built-in Python function `list`, and our folder can call `append` on that list with every event within a window.
```

docs/examples/sensor-data.md
```diff
@@ -0,0 +1,115 @@
...
+
+``` python
+def acc_values(acc, event):
+    acc.append(event)
+    return acc
+```
+
+With those two functions and our time configuration, we can now use our `fold_window` operator:
```
blakestier, line 77:
```suggestion
With a name for the operation, those two functions, and our time configuration, we can now use our `fold_window` operator:
```

src/outputs/dynamodb_output.rs
```diff
@@ -0,0 +1,308 @@
...
+            .into_iter()
+            .map(|(k, v)| (k.to_string(), create_dynamodb_attribute(v).unwrap()))
+            .collect();
+        Ok(AttributeValue::M(attribute_map))
+    } else {
+        Err(anyhow!("Could not convert `{val:?}` into a DynamoDB type."))
+    }
+}
+
+// impl<'a> FromPyObject<'a> for &'a AttributeValue {
```
blakestier, line 153: 🗑️ ?

blakestier (approved):
I went full English major on the docs but I think they read well and I appreciate the addition.

Everything makes sense to me

src/outputs/dynamodb_output.rs
```diff
@@ -0,0 +1,308 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use proptest::prelude::*;
+
+    proptest! {
+        #[test]
+        fn test_dynamodb_64_numbers(n: u64) {
```
blakestier, line 213: Wondering if it makes sense to have a test calls out the "bool before long" rule

src/outputs/dynamodb_output.rs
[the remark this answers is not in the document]
whoahbot: I did spend some time writing a wrapper for `AttributeValue`, and implementing `FromPyObject` on it. In the end, since I was the only intended consumer of the trait, I wasn't sure it was worth the extra code. I could be convinced otherwise.

I did forget to delete this though. I'll do that.

src/outputs/dynamodb_output.rs
```diff
@@ -0,0 +1,308 @@
...
+            for (k, v) in p {
+                let key: String = k
+                    .extract()
+                    .expect("Keys must be formatted as Strings for Dynamodb. Got `{k:?}`");
+                let attr = create_dynamodb_attribute(&v).unwrap();
+                request = request.item(key, attr);
+            }
+
+            // Add the key to the request as the primary key for the table
+            if let Some(key) = key {
```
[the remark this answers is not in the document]
Psykopear: Looking here: https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_PutItem.html
It seems the primary key is needed for items, so this should fail on the `unwrap` below if the key is None?

PR #152, opened 2022-10-12
Add a DynamoDB output adapter.

whoahbot:
- Adds `boto3` as an optional dependency for bytewax.


pysrc/bytewax/dynamodb_output.py
```diff
@@ -0,0 +1,43 @@
+import boto3
+
+from bytewax.outputs import ManualOutputConfig
+
+
+def output_builder(table):
+    """Construct an output function to map Bytewax's (epoch, item) structure
+    to boto3's required structure"""
+    dynamodb = boto3.resource("dynamodb")
+    table = dynamodb.Table(table)
```
davidselassie, line 10: `logging.info(f"Writing to DynamoDB table {table}")`?

pyproject.toml
```diff
@@ -20,6 +20,10 @@ classifiers = [
     "Topic :: System :: Distributed Computing",
 ]
 dependencies = ["multiprocess>=0.70", "dill>=0.3.5"]
+extras = ["dynamodb"]
+
+[project.optional-dependencies]
+dynamodb = ["boto3[crt]>=1.15.0"]
```
davidselassie, line 26: Extras on extras on extras.

examples/dynamodb.py
```diff
@@ -0,0 +1,95 @@
...
+    )
+)
+
+recovery_config = KafkaRecoveryConfig(
+    ["127.0.0.1:9092"],
+    "dynamodb",
+)
+
+if __name__ == "__main__":
+    cluster_main(
```
davidselassie, line 90: Is there a reason for your examples to use `cluster_main` with a single worker rather than `run_main`? The traceback info is much clearer if something goes wrong.

pysrc/bytewax/dynamodb_output.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class DynamoDBOutputConfig(ManualOutputConfig):
+    """Write output of a Dataflow to [DynamoDB](https://aws.amazon.com/dynamodb/).
+
+    Creates a new item, or replaces an old item with a new item. If an item that has the same primary key as the new item already exists in the specified table, the new item completely replaces the existing item.
+
+    Items are written to DynamoDB using [boto3](https://boto3.amazonaws.com/v1/documentation/api/latest/index.html). For more information on authentication and configuration, please see the documentation for boto3.
+
+    Items must be formatted as a dictionary, with value types that are supported by boto3. For more information about value types, see the documentation for [put_item](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#DynamoDB.Client.put_item).
```
davidselassie, line 26: I'm not sure I could read just this and understand what items should be flowing into the capture operator here. Perhaps something slightly more explicit like "items flowing into the capture operator should be dictionaries and will be passed as the `Item` kwarg to `put_item`. See [`boto`'s docs]() for the schema of that dictionary."

Also note terminology clash: our docs use "item" to mean individual object flowing through a dataflow edge; DynamoDB uses "item" to mean something stored in table. Not sure if we need to disambiguate a little.

pysrc/bytewax/dynamodb_output.py
```diff
@@ -0,0 +1,43 @@
...
+
+def output_builder(table):
+    """Construct an output function to map Bytewax's (epoch, item) structure
+    to boto3's required structure"""
+    dynamodb = boto3.resource("dynamodb")
+    table = dynamodb.Table(table)
+
+    def put_item(item):
+        _epoch, item = item
+        table.put_item(Item=item)
```
davidselassie, line 14: Do you think it's worthwhile to make this

```py
_epoch, kwargs = epoch_kwargs
table.put_item(**kwargs)
```

Then this allows users of this output config to have access to the fancier stuff of `put_item` like `ConditionExpression`.

pysrc/bytewax/dynamodb_output.py
davidselassie, line 26: You probably want to link to `Table.put_item` rather than `Client.put_item`. https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#DynamoDB.Table.put_item It has slightly different args and requirements and is what you're using.

davidselassie (commented):
Looks great. I hope this feels a little better than the Rust lib version. Some thoughts:

pysrc/bytewax/dynamodb_output.py
whoahbot: Great suggestion!

.github/workflows/CI.yml
```diff
@@ -45,7 +45,7 @@ jobs:
       shell: bash
       run: |
         WHEEL_FILE=$(ls ./dist/*.whl)
-        pip install $WHEEL_FILE -v
+        pip install $WHEEL_FILE'[dynamodb]' -v
```
whoahbot, line 48: There may be a better way to install all extras for the purpose of testing.

davidselassie: Could perhaps do something like [this idea of using an `all` extra](https://stackoverflow.com/questions/64685527/pip-install-with-all-extras)? Unclear how we'd do it via `pyproject.toml` but might help. Could use something abominable like [YAML references](https://www.linode.com/docs/guides/yaml-anchors-aliases-overrides-extensions/) to not duplicate.

whoahbot: I think I found it. Since pip 21.2, you can refer to your own project as an optional dependency.

PR #153, opened 2022-10-13
Adding an article about gcp command

miccioest:
This PR adds an article describing how to use Waxctl to deploy Bytewax dataflows on GCP VMs.


whoahbot (approved):
Thank you for the level of detail included in your article. This is great.

PR #154, opened 2022-10-14
Add a Bigquery output adapter

blakestier:
Adds an output which writes to Google's BigQuery. 

Includes a dependency on `google-cloud-bigquery` as an internal client provider.


pysrc/bytewax/bigquery/bigquery_output.py
```diff
@@ -0,0 +1,44 @@
...
+
+def output_builder(table_ref):
+    """Constructs an output function to write to BigQuery"""
+    client = bigquery.Client()
+    table = client.get_table(table_ref)
+
+    def write_rows(rows):
+        errors = client.insert_rows_json(table, rows)  # Make an API request.
+        if errors != []:
+            logging.info("Errors while inserting rows: {}".format(errors))
```
blakestier, line 15: Is this what we want?

If a user passes each row individually, the row will an error will fail to insert. If the user passes an array of rows, the entire batch will fail to insert. 

Should they be able to choose to crash or log/where to log?

pysrc/bytewax/bigquery/bigquery_output.py
```diff
@@ -0,0 +1,45 @@
...
+from bytewax.outputs import ManualOutputConfig
+
+
+def output_builder(table_ref):
+    """Constructs an output function to write to BigQuery"""
+    client = bigquery.Client()
+    table = client.get_table(table_ref)
+
+    def write_rows(rows):
+        errors = client.insert_rows_json(table, rows)  # Make an API request.
```
davidselassie, line 14: What do you think about repeating the concept of https://github.com/bytewax/bytewax/pull/152#discussion_r993736853 ?

I think it's a cool trick to tell people to format their items as the kwargs to be passed to the library's writing function. Then they can use the fancier features of the library without us having to manually thread them through. It's a little bit of a balance in that we still don't support any arbitrary use of the library, like an output process that might involve other functions on `client`, but it does allow a lot more flexibility than only allowing the most vanilla of writes.

So something like `client.insert_rows_json(table=table, **kwargs)`.

pysrc/bytewax/bigquery/bigquery_output.py
davidselassie: I don't think it really matters at this point. I think I'd default to throwing an exception, but logging is also fine.

FWIW, I think the initial reason I had the capture operator emit downstream what comes in is just since we don't support non-linear dataflows yet, so it's hard to test stuff in-line. And longer term, I think a better way to handle this is to make the `capture` operator emit some sort of "Ok / Err" downstream and then the designer can be as fancy as they want dealing with errors. Then we can retrofit all of these to pass a relevant error object from the client lib API; maybe that's catching exceptions and wrapping them and sending them downstream as objects or maybe it's using the return value, but it'll depend on the API.

davidselassie (commented):
Thanks for doing this. I have a few comments, but most of them I really should have brought up yesterday as part of the Dynamo PR  https://github.com/bytewax/bytewax/pull/152 since they set some higher-level patterns and I guess I was sleepy then, so sorry.

pysrc/bytewax/bigquery/__init__.py
```diff
@@ -0,0 +1 @@
+from .bigquery_output import BigqueryOutputConfig  # noqa: F401
```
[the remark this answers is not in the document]
whoahbot: @davidselassie I ended up making a module structure that doesn't include DynamoDB by default in `output` because it crashes if you don't have boto3 installed, which you won't unless you install `bytewax[dynamodb]`.

Is there a way to optionally include that in the output module based on what you've installed?

pysrc/bytewax/bigquery/bigquery_output.py
whoahbot: +1 to throwing an exception here.

pysrc/bytewax/bigquery/__init__.py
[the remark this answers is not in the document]
whoahbot: That sounds good to me!

pysrc/bytewax/bigquery/bigquery_output.py
```diff
@@ -0,0 +1,45 @@
...
+
+from bytewax.outputs import ManualOutputConfig
+
+
+def output_builder(table_ref):
+    """Constructs an output function to write to BigQuery"""
+    client = bigquery.Client()
+    table = client.get_table(table_ref)
+
+    def write_rows(rows):
```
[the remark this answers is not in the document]
davidselassie: Lies on my part. Ignore this. `ManualOutputConfig` already strips out the epoch. Sorry.

pysrc/bytewax/connectors/bigquery/__init__.py
```diff
@@ -0,0 +1 @@
+from .bigquery_output import BigqueryOutputConfig  # noqa: F401
```
davidselassie, line 1: What's the thinking behind calling the submodule `bigquery_output`? The full path would then be `connectors.bigquery.bigquery_output`. I think we could drop the second `bigquery` for just `connectors.bigquery.output`.

davidselassie, line 1: Related: What is the thinking behind re-importing this symbol here? Now it's under both `bytewax.connectors.bigquery.BigqueryOutputConfig` and `bytewax.connectors.bigquery.bigquery_output.BigqueryOutputConfig`. It will also be duplicated in the API docs.

I think it's less confusing to have only a single canonical location. And I'd vote for `bytewax.connectors.bigquery.output.BigqueryOutputConfig`.

Not sure if this is clarifying, but: the reason we use this re-import pattern elsewhere is that Maturin doesn't let us build PyO3-based subpackages, so we have to shuffle around all the Rust-defined Python classes to their final destinations. I don't think we need that here.

pysrc/bytewax/connectors/bigquery/bigquery_output.py
```diff
@@ -0,0 +1,49 @@
+import logging
+
+from google.cloud import bigquery
+
+from bytewax.outputs import ManualOutputConfig
+
+
+class BigqueryOutputConfig(ManualOutputConfig):
```
davidselassie, line 8: Minor point, but call this `BigQueryOutputConfig` to make casing consistent? Then the casing matches `DynamoDBOutputConfig` (since it's not `DynamodbOutputConfig`). I guess this is only the second time we have an output config whose name is more than one word.

davidselassie (approved):
This looks good. I have one more module naming comment, but now I see I should have directed it to @whoahbot in the last PR. Sorry sorry.

davidselassie (approved):

davidselassie (approved):

PR #156, opened 2022-10-18
Prepare for v0.12.0 release.

whoahbot:


apidocs/generate.py
```diff
@@ -0,0 +1,34 @@
...
+pdoc.link_inheritance(context)
+
+# Use our templates
+pdoc.tpl_lookup.directories.insert(0, "templates")
+
+# HACK: These two functions aren't picked up by our module structure
+# as a part of bytewax.execution, so we inject them here directly.
+execution = bytewax.doc["execution"]
+execution.doc["run_main"] = pdoc.Function("run_main", execution, run_main)
+execution.doc["cluster_main"] = pdoc.Function("cluster_main", execution, cluster_main)
```
davidselassie, line 19: Are there any other functions that this happens for? Can you describe the general rule for when we need to hack this in? It's any function that is defined via PyO3 that we re-import into a different module?

davidselassie (approved):
Thanks for figuring out pdoc things.

apidocs/generate.py
whoahbot: Yes, this currently happens for `sleep_keep_gil` and `sleep_release_gil`. I think with our current package structure, you need to do this for any function, since we can't specify the module like we can for a class.

`sleep_keep_gil` is also an interesting case to look at. If you make a new virtualenv, and install Bytewax, you need to import it like so:

```python
In [1]: from bytewax import sleep_keep_gil
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
Cell In [1], line 1
----> 1 from bytewax import sleep_keep_gil

ImportError: cannot import name 'sleep_keep_gil' from 'bytewax' (/home/whoahbot/.pyenv/versions/3.9.13/envs/junk-drawer/lib/python3.9/site-packages/bytewax/__init__.py)

In [3]: from bytewax.bytewax import sleep_keep_gil
```

Perhaps there is a better way to structure our mixed Python/PyO3 project to avoid these hacks?

CHANGELOG.md
```diff
@@ -5,13 +5,15 @@
 __Add any extra change notes here and we'll put them in the release
 notes on GitHub when we make a new release.__
 
+## 0.12.0
+
```
blakestier, line 9: Mmm I neglected this in my PR but should we highlight new connectors?

RELEASE.md
```diff
@@ -16,9 +16,10 @@ Make a PR which commits the following:
 2. Commits updated API docs
 
    ```sh
-   (.venv) bytewax/apidocs $ maturin develop -E dev
+   (.venv) bytewax/ $ maturin develop -E dev
+   (.venv) bytewax/ $ cd apidocs
    (.venv) bytewax/apidocs $ rm -rf html/
-   (.venv) bytewax/apidocs $ ./build.sh
+   (.venv) bytewax/apidocs $ python ./generate.py
```
blakestier, line 22: 👍

pysrc/bytewax/execution.py
```diff
@@ -22,13 +22,26 @@
...
-    run_main,
     TestingEpochConfig,
 )
 
+# Due to our package structure, we need to define __all__
+# in any submodule as pdoc will not find the documentation
+# for functions imported here, but defined in another submodule.
+# See https://pdoc3.github.io/pdoc/doc/pdoc/#what-objects-are-documented
+# for more information.
+__all__ = [
```
davidselassie, line 37: My read of [the Python module docs for `__all__`](https://docs.python.org/3/tutorial/modules.html?highlight=__all__#importing-from-a-package) is that it is a manually maintained list of all public names this module contains, and if you define it, you're telling the module loader to not automatically determine public names. Thus it should contain `"spawn_cluster"` too, even though it is named as a "public" item.

Notice that it looks like the HTML docs now don't have spawn cluster.

whoahbot: Good catch, will fix.

PR #160, opened 2022-10-21
CDC-esque recovery operators

davidselassie:
This is some setup to help with re-scaling work.

The idea here is to have a more refined sense of the concept of a "state change" r.e. the recovery system. There's two levels of changes: changes in the current state within steps, then changes to the recovery store. By better separating these two layers, we can have our recovery partitions be separate from our workers.

Also, by unifying these concepts with some generic "keyed change" types, we can have a few primitive operators (like "read stream of changes" and "write stream of changes") and re-constitute state where we need it in a less confusing way. I really think reading through the way the recovery dataflows work is much more clear now.

Also because `recovery/mod.rs` was getting huge, I wanted to break it up into smaller sub-modules to make the structure more understandable. Since I was still working out the structure as I was making the above changes, I added a "retcon" commit that is a ton of bulk moving to the new module structure without changing any code. That will hopefully make the following commit with the meat of the changes easier to see what's going on.

New module structure within `recovery`:

- `python` for all the stuff related to reading configs.
- `dataflows` with what will be added to the dataflow for backup and how we do resume
- `model` with the types we use to model state and progress
- `store` with the implementations of the state and progress stores


src/recovery/model/change.rs
```diff
@@ -0,0 +1,139 @@
...
+        for change in batch {
+            self.write(change);
+        }
+    }
+}
+
+/// Allows reading keyed changes from a store.
+///
+/// This is not exactly the inverse of [`KWriter`]: a store is allowed
+/// to coallece writes that happen to the same key.
```
Psykopear, line 67: coalesce?

davidselassie: Yeah, this is a tricky concept that I should clarify. The reader and writer contracts do not say that "every write I promise to read back". They say "if you apply all reads to an empty K-V store, you'll get the same store state".

Example:

1. Write `Change::Upsert("a", 1)`, `Change::Upsert("b", 2)`, `Change::Upsert("a", 2)`.
2. That means the resulting store is something like `{"a": 2, "b": 2}`.
3. It's thus valid for a read of that store to result in `Change::Upsert("b", 2)`, `Change::Upsert("a", 2)`, since those are the only writes needed to recreate the same state.

"Coalesce" is a word that sort of gets at that: combining writes to only have the latest ones.

Let me try updating the rustdocs here to be more clear.

src/recovery/store/in_mem.rs
```diff
@@ -0,0 +1,353 @@
...
+        let mut garbage = Vec::new();
+
+        let mut empty_keys = Vec::new();
+        for (key, non_garbage_changes) in self.db.iter_mut() {
+            // This now contains `(epoch, op)` in epoch order where
+            // epoch < frontier. So all operations that could be GCd.
+            // Unfortunately [`BTreeMap::split_off`] returns the "high
+            // end" / not garbage, so we have to swap around pointers
+            // to get ownership right without cloning.
+            let tmp = non_garbage_changes.split_off(frontier);
```
whoahbot, line 62: Maybe [retain](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.retain) here instead?

whoahbot (approved):
Thanks for putting this together. I think this helps clarify a good deal of our recovery code.

Left you some non-blocking questions.

pytests/test_operators.py
```diff
@@ -248,6 +248,7 @@ def test_stateful_map(recovery_config):
         "BOOM",
         "b",
         "c",
+        "a",
```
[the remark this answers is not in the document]
davidselassie: Oh, this was from some debugging when things weren't working and I wanted to activate an older key.

I think we're going to eventually need a more nuanced test suite to get more corner cases of stateful operators. I'm not totally sure what's a nice systemic way to do that yet.

src/recovery/store/in_mem.rs
whoahbot: Funny, `retain` is actually implemented in terms of `drain_filter`, but yeah, unstable.

src/recovery/model/change.rs
Psykopear: Nope, I was pointing out a typo `coallece`->`coalesce`

davidselassie: Hah. Awkward for me. Well, more documentation doesn't hurt anyway.

PR #162, opened 2022-10-28
Traits refactoring

Psykopear:
This is a refactoring to make the pattern we use to build concrete objects from python config objects more explicit.
I also moved the `clocks` stuff in a submodule, but nothing should really change.

The first trait introduced is `ParentClass`. This trait should be implemented by any class we expose in Python that is a parent class of config objects, and it requires a method to convert from parent to one of the children:

```rust
pub(crate) trait ParentClass {
    type Children;
    fn get_subclass(&self, py: Python) -> StringResult<Self::Children>;
}
```

Then we have one trait for each module, that specific configurations have to implement. This trait requires a method to build a concrete object from a config one, so in the `inputs` modules it builds an `InputReader` from a child of the `InputConfig` class.

For example in the input module we have:
```rust
pub(crate) trait InputBuilder {
    fn build(
        &self,
        py: Python,
        worker_index: WorkerIndex,
        worker_count: usize,
        resume_snapshot: Option<StateBytes>,
    ) -> StringResult<Box<dyn InputReader<TdPyAny>>>;
}
```

And inputs have to implement this trait, example in ManualInputconfig:
```rust
impl InputBuilder for ManualInputConfig {
    fn build(
        &self,
        py: Python,
        worker_index: WorkerIndex,
        worker_count: usize,
        resume_snapshot: Option<StateBytes>,
    ) -> StringResult<Box<dyn InputReader<TdPyAny>>> {
        Ok(Box::new(ManualInput::new(
            py,
            self.input_builder.clone(),
            worker_index,
            worker_count,
            resume_snapshot,
        )))
    }
}
```
The nice thing is that this allows us to implement `InputBuilder` for the generic `InputConfig` if it implements the `ParentClass` trait:

```rust
impl InputBuilder for Py<InputConfig> {
    fn build(
        &self,
        py: Python,
        worker_index: WorkerIndex,
        worker_count: usize,
        resume_snapshot: Option<StateBytes>,
    ) -> StringResult<Box<dyn InputReader<TdPyAny>>> {
        self.get_subclass(py)?
            .build(py, worker_index, worker_count, resume_snapshot)
    }
}
```

So that we can do this:
```rust
let input_reader = input_config.build(py, worker_index, worker_count, step_resume_state)?;
```
instead of what we used to do:
```rust
let input_reader = build_input_reader(
    py,
    input_config,
    worker_index,
    worker_count,
    step_resume_state,
)?;
```

*NOTE*: I stopped here, but we could think of a more generic `PyBuilder` trait so that we don't need a new trait for each module, but after trying it I think it makes the implementation a bit harder to read:

```rust
pub(crate) trait PyBuilder {
    type Input;
    type Output;
    fn build(&self, py: Python, params: Self::Input) -> StringResult<Self::Output>;
}
```
Which turn the impl for ManualInputConfig into this
```rust
impl PyBuilder for ManualInputConfig {
    type Input = (WorkerIndex, usize, Option<StateBytes>);
    type Output = Box<dyn InputReader<TdPyAny>>;

    fn build(
        &self,
        py: Python,
        (worker_index, worker_count, resume_snapshot): Self::Input,
    ) -> StringResult<Self::Output> {
        Ok(Box::new(ManualInput::new(
            py,
            self.input_builder.clone(),
            worker_index,
            worker_count,
            resume_snapshot,
        )))
    }
}
```

And the ParentClass impl to something like:
```rust
impl ParentClass for Py<InputConfig> {
    type Children = Box<
        dyn PyBuilder<
            Input = (WorkerIndex, usize, Option<StateBytes>),
            Output = Box<dyn InputReader<TdPyAny>>,
        >,
    >;

    fn get_subclass(&self, py: Python) -> StringResult<Self::Children> {
        if let Ok(conf) = self.extract::<ManualInputConfig>(py) {
            Ok(Box::new(conf))
        } else if let Ok(conf) = self.extract::<KafkaInputConfig>(py) {
            Ok(Box::new(conf))
        } else {
            let pytype = self.as_ref(py).get_type();
            Err(format!("Unknown input_config type: {pytype}"))
        }
    }
}

```

*Final edit*: The end goal of this was to write a blanket implementation of `PyBuilder` for any struct that implement `ParentClass` with `Children = Box<dyn PyBuilder>`, but it doesn't work as it is. I'm going to write it here as a reference, maybe I was close enough:

```rust
impl<T, B> PyBuilder for T
where
    B: PyBuilder,
    T: ParentClass<Children = Box<B>>,
{
    type Input = B::Input;
    type Output = B::Output;

    fn build(&self, py: Python, params: Self::Input) -> StringResult<Self::Output> {
        self.get_subclass(py)?.build(py, params)
    }
}
```


src/common.rs
```diff
@@ -0,0 +1,11 @@
+use pyo3::Python;
+
+/// Result type used in the crate that holds a String as the Err type.
+pub(crate) type StringResult<T> = Result<T, String>;
+
+// This is a trait that can be implemented by any parent class.
+// The function returns one of the possible subclasses instances.
+pub(crate) trait ParentClass {
```
davidselassie, line 8: Could we maybe call this `PyConfigClass` to have the name hint at where it's supposed to be used?

src/inputs/mod.rs
```diff
@@ -78,6 +77,49 @@ impl InputConfig {
     }
 }
 
+// This trait has to be implemented by each InputConfig subclass.
+// It is used to create an InputReader from an InputConfig.
+pub(crate) trait InputBuilder {
```
davidselassie, line 82: FWIW, I don't know that it's worth it to try to extract this builder pattern further to be used across config types. Like not all concrete Rust-side instances will need things like `resume_snapshot` to be built or `worker_count`, so it might end up feeling messy. I guess we can return to this later.

src/common.rs
```diff
@@ -0,0 +1,11 @@
...
+    type Children;
+    fn get_subclass(&self, py: Python) -> StringResult<Self::Children>;
```
davidselassie, line 10: `downcast`? It's effectively doing that and matches the sort of the same name used by PyO3.

src/common.rs
davidselassie, line 8: Maybe this should be in `pyo3_extensions`? It's pretty specifically an extension to the way we're using PyO3 Pyclasses.

davidselassie (approved):
This looks good. Thanks for doing this. I have some random thoughts you can take or leave.

There's going to be a bit more complexity introduced into the input, output, and recovery configs because of rescaling, so we'll see how that meshes with this when that happens.

src/inputs/mod.rs
Psykopear: The idea was to make the pattern generic over the arguments passed to the `build` function, with the `type Input;` type param.
This would be different for each module. In InputConfig it's `(WorkerIndex, usize, Option<StateBytes>)`, in OutputConfig it's `(WorkerIndex, usize)` and in ClockConfig it's `()`.

Each subclass of each module would have to share the params set, but that's already happening, so we should be able to have the same behavior.

Anyway I didn't include it in this PR because I too am not sure it's worth it, since it makes the code a bit weirder to read.

Psykopear:
I noticed that I didn't use this pattern for the epoch builders. To do that I had to make a little change, since the builders are generic over the Scope parameter.
I replaced the use of the associated type with a generic on the trait, so `PyConfigClass` is now:
```rust
pub(crate) trait PyConfigClass<S> {
    fn downcast(&self, py: Python) -> StringResult<S>;
}
```

This keeps working on the previous cases (with a small fix), but also allows me to do:

```rust
impl<S> PyConfigClass<Box<dyn EpochBuilder<S>>> for Py<EpochConfig>
where
    S: Scope<Timestamp = u64>,
{
    fn downcast(&self, py: Python) -> StringResult<Box<dyn EpochBuilder<S>>> {
        if let Ok(config) = self.extract::<TestingEpochConfig>(py) {
            Ok(Box::new(config))
        } else if let Ok(config) = self.extract::<PeriodicEpochConfig>(py) {
            Ok(Box::new(config))
        } else {
            let pytype = self.as_ref(py).get_type();
            Err(format!("Unknown epoch_config type: {pytype}"))
        }
    }
}
```

which (I think) wouldn't have been possible without GAT with the associated type, right?

In the process I also moved the `epoch config` stuff inside it's own module, so that execution/mod.rs is smaller and narrower in scope. As I did for the clocks in windows, I left the submodule inside `execution`, but maybe it makes sense to make everything top level?

whoahbot (approved):

davidselassie (approved):

PR #163, opened 2022-10-28
Fix dockerfile

Psykopear:
This PR fixes the current Docerfile, and also adds a Dockerfile in `apidocs/` to help rebuild the docs on a windows environment, where compiling bytewax is complicated


Psykopear:
@konradsienkowski it should be the same dockerfile we tried on slack, but try this and let me know if it works.
You should only have to run `./pdocs.sh` from the bytewax repository.

miccioest (approved):
Looks good.

PR #169, opened 2022-11-14
Adding manylinux_2_27 wheel building to CI

miccioest:
This PR automates the building of compatible wheels with Google Colab environments which have the `2.27` version of `GLIBC`.

### DIFF
--- a/Dockerfile.Colab
+++ b/Dockerfile.Colab
@@ -1,0 +1,47 @@
+FROM ubuntu:18.04
+
+# Install Python 3.7
+RUN apt update
+RUN apt install software-properties-common -y
+RUN add-apt-repository ppa:deadsnakes/ppa
+RUN apt install python3.7 -y
+RUN python3.7 --version
+
+# Install pip and venv
+RUN apt install python3-pip -y
+RUN apt-get install python3.7-venv -y
+
+# Install cargo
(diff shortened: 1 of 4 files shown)


Dockerfile.Colab
```diff
@@ -0,0 +1,47 @@
+FROM ubuntu:18.04
+
+# Install Python 3.7
+RUN apt update
+RUN apt install software-properties-common -y
+RUN add-apt-repository ppa:deadsnakes/ppa
+RUN apt install python3.7 -y
+RUN python3.7 --version
```
davidselassie, line 8: Does Collab support multiple Python versions? Will we have to bump this when they update Python?

davidselassie (approved):
I was vaguely following along in Zulip, and this makes sense to me.

Dockerfile.Colab
miccioest: Thanks, David

As far as I know, they support 2.7, 3.6, and, 3.7. So, if someone wants to use Bytewax the only option is 3.7.

It's true that if they add a python version we will need to change our custom image as well. But only if they don't upgrade the operating system. Because if they upgrade also the operating system we won't need this special `2_27` wheel anymore and we should remove this new job.

davidselassie (approved):

PR #175, opened 2022-11-19
Adds webserver

whoahbot:
- Changes pickling of steps from Tuple to PyDict.

### DIFF
--- a/pysrc/bytewax/_encoder.py
+++ b/pysrc/bytewax/_encoder.py
@@ -24,7 +24,7 @@ class DataflowEncoder(json.JSONEncoder):
     """
 
     def default(self, obj):
-        if "__getstate__" in dir(obj):
+        if hasattr(obj, "__getstate__"):
             return obj.__getstate__()
         if isinstance(obj, types.BuiltinFunctionType):
             return obj.__name__
--- a/pysrc/bytewax/_encoder.py
+++ b/pysrc/bytewax/_encoder.py
@@ -12,33 +12,21 @@ from .outputs import (
     KafkaOutputConfig,
 )
 
-# A list of all types that should use vars(obj)
-# which uses the Python `__getstate__` method
-# to return a JSON encodable representation.
-config_types = (
-    InputConfig,
-    ClockConfig,
-    WindowConfig,
-    ManualOutputConfig,
-    ManualInputConfig,
-    ManualEpochOutputConfig,
-    StdOutputConfig,
-    KafkaOutputConfig,
-    SystemClockConfig,
-)
-
 
 class DataflowEncoder(json.JSONEncoder):
     """Custom JSON encoder for a Dataflow
 
     This class is used in conjunction with the Rust `webserver` module to
     produce a JSON representation of a bytewax Dataflow.
+
+    __getstate__() is a method defined on all of the Python classes we
+    create in Rust to return a PyDict representation of that class.
     """
 
     def default(self, obj):
         if isinstance(obj, Dataflow):
             return {"Dataflow": {"steps": obj.steps}}
-        if isinstance(obj, config_types):
+        if "__getstate__" in dir(obj):
             return obj.__getstate__()
         if isinstance(obj, types.BuiltinFunctionType):
             return obj.__name__
--- a/pysrc/bytewax/_encoder.py
+++ b/pysrc/bytewax/_encoder.py
@@ -24,8 +24,6 @@ class DataflowEncoder(json.JSONEncoder):
     """
 
     def default(self, obj):
-        if isinstance(obj, Dataflow):
-            return {"Dataflow": {"steps": obj.steps}}
         if "__getstate__" in dir(obj):
             return obj.__getstate__()
         if isinstance(obj, types.BuiltinFunctionType):
--- a/pysrc/bytewax/encoder.py
+++ b/pysrc/bytewax/encoder.py
@@ -42,7 +42,11 @@ class DataflowEncoder(json.JSONEncoder):
             return obj.__getstate__()
         if isinstance(obj, types.BuiltinFunctionType):
             return obj.__name__
+        if isinstance(obj, types.MethodDescriptorType):
+            return obj.__name__
         if isinstance(obj, types.FunctionType):
+            return obj.__name__
+        if isinstance(obj, types.BuiltinMethodType):
             return obj.__name__
         if isinstance(obj, (datetime.date, datetime.datetime)):
             return obj.isoformat()
@@ -50,7 +54,10 @@ class DataflowEncoder(json.JSONEncoder):
             return str(obj)
 
         # Call the default encoder method for any other instance types.
-        return json.JSONEncoder.default(self, obj)
+        try:
+            return json.JSONEncoder.default(self, obj)
+        except TypeError as err:
+            raise TypeError(f"{obj} can not be JSON encoded: {err}")
 
 
 def encode_dataflow(dataflow: Dataflow):
--- a/pysrc/bytewax/encoder.py
+++ b/pysrc/bytewax/encoder.py
@@ -1,0 +1,60 @@
+import datetime
+import json
+import types
+
+from .bytewax import Dataflow
+from .inputs import ManualInputConfig, InputConfig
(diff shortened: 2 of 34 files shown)


src/execution/mod.rs
```diff
@@ -427,7 +428,18 @@ where
     SW: StateWriter<u64> + 'static,
 {
     let span = tracing::trace_span!("Building dataflow").entered();
+
+    let rt = tokio::runtime::Builder::new_multi_thread()
+        .enable_all()
+        .build()
+        .unwrap();
```
Psykopear, line 435: We are starting to have a certain number of tokio runtimes running depending on the config (recovery, tracing, and this).
This is not a problem per se, but it could impact performances.
We'd need to profile to understand if it's a problem at all, but we could think of sharing the runtime for lower priority tasks (tracing and the webserver)

src/macros.rs
```diff
@@ -87,3 +87,17 @@ macro_rules! log_func {
...
+        $self.$attr_name = $pydict
+            .get_item(stringify!($attr_name))
+            .ok_or_else(|| {
+                PyValueError::new_err(format!(
+                    "bad pickle contents for {}: {}",
+                    stringify!($attr_name), $pydict
+                ))
+            })?.extract()?;
+    };
+}
```
Psykopear, line 103: We could do this with a function, which I'd prefer over a macro:

```rust
fn pickle_extract<'a, D>(dict: &'a PyDict, key: &str) -> PyResult<D>
where
    D: FromPyObject<'a>,
{
    dict.get_item(key)
        .ok_or_else(|| PyValueError::new_err(format!("bad pickle contents for {}: {}", key, dict)))?
        .extract()
}
```

And used like this:
```rust
        self.service_name = pickle_extract(dict, "service_name")?;

```

src/dataflow/mod.rs
```diff
@@ -684,64 +691,134 @@ pub(crate) enum Step {
...
-            Ok(Self::Capture { output_config })
-        } else {
-            Err(PyValueError::new_err(format!(
-                "bad python repr when unpickling Step: {obj:?}"
-            )))
+        let dict: &PyDict = obj.downcast()?;
+        let t: &str = dict
+            .get_item("type")
+            .expect("Unable to extract step, missing `type` field.")
+            .extract()?;
```
Psykopear, line 703: Since this is used a lot here, maybe wrap it in a function:
```rust
fn step_extract<'a, D>(dict: &'a PyDict, key: &str) -> PyResult<D>
where
    D: FromPyObject<'a>,
{
    dict.get_item(key)
        .expect("Unable to extract step, missing `{key}` field.")
        .extract()
}
```

Psykopear: Same for all the steps below

Psykopear:
```rust
let t: &str = step_extract(dict, "type")?;
```

```rust
        match t {
            "Input" => Ok(Self::Input {
                step_id: step_extract(dict, "step_id")?,
                input_config: step_extract(dict, "input_config")?,
            }),
        ...
```

src/webserver/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+        let encoder_module =
+            PyModule::import(py, "bytewax.encoder").expect("Unable to load Bytewax encoder module");
+        // For convenience, we are using a helper function supplied in the
+        // bytewax.encoder module.
+        let encode = encoder_module
+            .getattr("encode_dataflow")
+            .expect("Unable to load encode_dataflow function");
+
+        encode.call1((dataflow,)).unwrap().to_string()
+    });
```
Psykopear, line 46: Do we need to encode the dataflow at each request?
Unless we need to support a dataflow that can change while running (or other usecases I'm missing here), I'd do this only once

src/execution/mod.rs
whoahbot: That's a good point. For now, I'll limit the number of threads that this one manages.

pytests/test_encoder.py
```diff
@@ -0,0 +1,120 @@
...
+                "builder": "<lambda>",
+                "mapper": "update",
+                "step_id": "order_book",
+                "type": "StatefulMap",
+            },
+        ),
+        ("method_descriptor", {"type": "Map", "mapper": "split"}),
+    ],
+)
+def test_dataflow_encoding(step, output):
```
davidselassie, line 85: What's going on here? Are you asserting that the JSON encoding of each step type works? I think having 8 separate tests would be a little more clear.

davidselassie (commented):
Looks generally good! I have a few questions / suggestions.

src/execution/mod.rs
```diff
@@ -427,7 +428,20 @@ where
...
+    let rt = tokio::runtime::Builder::new_multi_thread()
+        .worker_threads(2)
+        .thread_name("webserver-threads")
+        .enable_all()
+        .build()
+        .unwrap();
+
     let probe = Python::with_gil(|py| {
+        let df = flow.extract(py).unwrap();
+        if worker.index() == 0 {
```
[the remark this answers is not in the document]
whoahbot: Good question. The initial idea was to have just one of the workers on a host spawn the webserver. I think there will be a case that the webserver will expose dynamic information for a process, but I don't think that we'll be reporting data on a worker thread basis.

pysrc/bytewax/encoder.py
```diff
@@ -0,0 +1,67 @@
...
+class DataflowEncoder(json.JSONEncoder):
+    """Custom JSON encoder for a Dataflow
+
+    This class is used in conjunction with the Rust `webserver` module to
+    produce a JSON representation of a bytewax Dataflow.
+    """
+
+    def default(self, obj):
+        if isinstance(obj, Dataflow):
+            return {"Dataflow": {"steps": obj.steps}}
```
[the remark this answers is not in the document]
whoahbot: I was trying to match the ergonomics we proposed in Slack, but I'm happy to remove this reformatting if @konradsienkowski and @Psykopear are :+1: with the shape of the JSON without it.

src/common.rs
```diff
@@ -1,2 +1,24 @@
...
+        .extract()
+}
+
+// Extract a field of a Dataflow `Step` from a PyDict
+pub(crate) fn step_extract<'a, D>(dict: &'a PyDict, key: &str) -> PyResult<D>
+where
+    D: FromPyObject<'a>,
+{
+    dict.get_item(key)
+        .expect("Unable to extract step, missing `{key}` field.")
```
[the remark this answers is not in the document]
whoahbot: Good call. I'll refactor them to use the same function.

pysrc/bytewax/encoder.py
```diff
@@ -0,0 +1,67 @@
...
+        if isinstance(obj, config_types):
```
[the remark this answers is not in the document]
whoahbot: I'll see if `__getstate__` is defined on most types. Maintaining the list of classes that should use it was one of my concerns with this approach.

pytests/test_encoder.py
whoahbot: I can break them out into multiple tests if that's clearest.

pysrc/bytewax/_encoder.py
```diff
@@ -0,0 +1,53 @@
...
+
+    This class is used in conjunction with the Rust `webserver` module to
+    produce a JSON representation of a bytewax Dataflow.
+
+    __getstate__() is a method defined on all of the Python classes we
+    create in Rust to return a PyDict representation of that class.
+    """
+
+    def default(self, obj):
+        if "__getstate__" in dir(obj):
```
davidselassie, line 27: I think a slightly more robust way to do this is [`hasattr`](https://docs.python.org/3/library/functions.html#hasattr) (and the way I've seen in the past), given all the hesitation we've had surrounding `dir`.

davidselassie (approved):

pysrc/bytewax/encoder.py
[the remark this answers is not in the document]
whoahbot: I made the change, it feels better to have them be uniform. I'll let @konradsienkowski and @Psykopear weigh in if they want it changed.

davidselassie (approved):

PR #176, opened 2022-11-25
Added EXPOSE command in Dockerfiles

Psykopear:
Added an `EXPOSE` line in Dockerfiles to document which ports we need to expose from inside the container.

I also made a small fix in how we get the port from the env var, since previously it accepted an entire url.
I'm not sure we need to use anything else than "0.0.0.0:PORT", but if we do I can change it back to accept the entire url.


miccioest (approved):
Looks good. Thanks for improving Dockerfiles

PR #180, opened 2022-12-05
Use cbfmt with pre-commit

whoahbot:
Requires the installation of https://github.com/lukas-reineke/cbfmt when using the pre-commit hooks.

- Uses https://github.com/bytewax/pre-commit-cbfmt to invoke cbfmt.
- Check in formatting changes made with cbfmt.


davidselassie (approved):
Thanks!

PR #183, opened 2022-12-22
Adds collect window operator

davidselassie:
Adds a new windowing operator that returns a time-sorted list of all
the values within a window.

Tests.

Had to thread through the `item_time` into the
`WindowLogic::with_next` to allow consistent sorting. We might want to
expose that to the custom `fold_window` and `reduce_window` operators
as of now they don't guarantee any order of items being aggregated.

### DIFF
--- a/src/operators/collect_window.rs
+++ b/src/operators/collect_window.rs
@@ -1,0 +1,58 @@
+use chrono::{DateTime, Utc};
+use pyo3::prelude::*;
+use tracing::field::debug;
+
+use crate::{pyo3_extensions::TdPyAny, window::*};
+
+/// Implements the collect window operator.
+///
+/// Append values in a window onto a list. Emit the list when the
+/// window closes.
+pub(crate) struct CollectWindowLogic {
+    acc: Vec<(TdPyAny, DateTime<Utc>)>,
+}
+
+impl CollectWindowLogic {
+    pub(crate) fn builder() -> impl Fn(Option<StateBytes>) -> Self {
+        move |resume_snapshot| {
+            let acc = resume_snapshot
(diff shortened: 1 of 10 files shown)


src/operators/collect_window.rs
```diff
@@ -0,0 +1,58 @@
...
+            }
+            // Emit in item time order at end of window.
+            None => {
+                self.acc
+                    .sort_by_key(|(_value, item_time)| item_time.clone());
+                let new_acc = Vec::new();
+                let out_values: Vec<TdPyAny> = std::mem::replace(&mut self.acc, new_acc)
+                    .into_iter()
+                    .map(|(value, _item_time)| value)
+                    .collect();
```
Psykopear, line 48: So, what you do here is:
- Sort `self.acc`
- Create an empty vector (`new_acc`)
- Replace `self.acc` with the empty vector
- Iterate over the previous contents of `self.acc` to extract the values into `out_values`

If I got this correctly, I think using `drain` rather than `std::mem::replace` might be more readable:

```rust
                self.acc
                    .sort_by_key(|(_value, item_time)| item_time.clone());
                let out_values: Vec<TdPyAny> = self.acc.drain(..).map(|(value, _)| value).collect();
                Python::with_gil(|py| Some(out_values.into_py(py).into()))


````

I'm not good at reading assembly from godbolt, but doing a quick check both generate an almost identical output:

https://rust.godbolt.org/z/rKqPMP66M

Psykopear (approved):

src/operators/collect_window.rs
davidselassie: That's a good call, that's more straightforward. Thanks

whoahbot (approved):
LGTM

PR #191, opened 2023-01-05
Python 3.11 package

### LINKED ISSUE #190: [FEATURE] Publish package for Python 3.11
**Is your feature request related to a problem? Please describe.**
We currently publish packages for python `3.7`-`3.10`

**Describe the solution you'd like**
Add Python `3.11` to the CI matrix

Psykopear:
Fixes #190

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -126,7 +126,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10']
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
     steps:
     - uses: actions/checkout@v2
     - name: Set up Docker Buildx
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -45,7 +45,7 @@ jobs:
       - name: Check the Version and Quantity of Wheel files
         env:
           NEEDED_VERSION: ${{steps.get-info.outputs.bytewax-version}}
-          WHEELS: 16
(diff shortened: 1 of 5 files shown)


Psykopear:
I'm going to work on some of the failing tests, but for all the tests that fails with:
```
 TypeError: fork_exec() takes exactly 23 arguments (21 given)
```

We'll need to wait for `multiprocess`  0.70.15 release (or work on removing the need for it):  https://github.com/uqfoundation/multiprocess/issues/122

I'm leaving the PR in draft untill that's released.
We could build the main branch for the dependency, but I'd prefer to wait for a proper release here.

Psykopear:
Another issue was caused by python 3.11 adding a default `__getstate__()` implementation for the `object` type, which returns `None`. Our encoder only checked for the existence of the method, and use that result, so I added a check for `is not None` on the result of the `__getstate__` call.

Psykopear:
Since we removed the dependency from multiprocessing in a recently merged PR, I rebased this on top of the current main, let's see if everything else work.

.github/workflows/CD.yml
```diff
@@ -126,7 +126,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10']
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
```
whoahbot, line 129: I think we also need to change the magic number in this file:

https://github.com/bytewax/bytewax/blob/84f8e1324a7d2557819215b4ed6eac7a55e7ec85/.github/workflows/CD.yml#L48

whoahbot (approved):

PR #192, opened 2023-01-11
Replace generic epoch with explicit u64

davidselassie:
We had a ton of generic `<T>` which I originally thought we might want
to use in case we wanted to swithc the type of our epochs. But it
turns out for recovery we have to do some epoch arithmetic so we were
effetively pinning it to `u64` anyway. This gets rid of that
complexity and hard codes that type.


Psykopear (approved):
I really like this :+1: Especially for the cases where we had to constrain the `T` with other traits that `u64` already implements (like `where T: Ord + Clone` etc.)

whoahbot (approved):

Issue #181, opened 2022-12-14
Recovery state will be incorrect if any worker crashes during the initial epoch


davidselassie:
### Bug Description

During dataflow execution, we write out whenever a worker completes an epoch. We use these written epochs to re-derive the progress of the entire cluster. Unfortunately, though, on resume we do not know beforehand how many worker's progress messages we're waiting for. Thus if the cluster crashes after worker 1 has written progress but worker 2 has not, then upon resume, it'll look like you're resuming from just a cluster of size 1 and will silently skip data.

We need to have some way of synchronously waiting for all workers to confirm they've written a marker of their existence to the recovery store before each execution (resume or not).

### Python version (`python -V`)

Python 3.10.6

### Bytewax version (pip list | grep bytewax)

0.13.1

### Steps to Reproduce

I don't have some example code yet, but this should theoretically be possible.

PR #193, opened 2023-01-25
Store execution IDs and worker count for recovery

davidselassie:
The progress recovery data model used to have an issue: we never knew
how many workers were part of a cluster. This caused us to "lose"
workers if we crashed before any progress data was written by a
worker; during resume, it just looked like the cluster was smaller and
nothing was missing. Adds a new kind of progress recovery message that
is written as each cluster starts up with the total worker count so we
can know if we're missing a later progress message.

Fixes https://github.com/bytewax/bytewax/issues/181

Related to this, we also need to disambiguate between workers across
clusters. During recovery we need to know which "worker 3" a progress
message "worker 3 moved to epoch 30" applies to. To do this we add an
`Execution` ID which is sort of like "epochs but for the whole
cluster". This is an `u64` that increments on each resume and all
progress recovery data is labeled with it so we ignore data from older
clusters that might come out of order. Adds tables to SQLite recovery
DB to do that.

This gets us one more step closer to rescaling.

Rewrites the `Progress` operator to work on frontiers again... Turns
out if a worker never generates input, it'll never generate state
data, which never generates progress info because it was based off of
the `Tick`s flowing through that part of the dataflow graph. This
means it's unnessecary gymnastics to store the `BorderEpoch` so
removes that from the progress data model.

### DIFF
--- a/src/recovery/operators.rs
+++ b/src/recovery/operators.rs
@@ -204,57 +204,39 @@ where
                     tmp_incoming.clear();
                 });
 
-                cap = if let Some(cap) = cap.clone() {
+                cap = cap.take().and_then(|cap| {
                     let frontier = input_frontiers[0].frontier().iter().min().copied();
-
                     // Do not delay cap before the write; we will
                     // delay downstream progress messages longer than
                     // necessary if so. This would manifest as
                     // resuming from an epoch that seems "too
                     // early". Write out the progress at the end of
                     // each epoch and where the frontier has moved.
-                    let should_write = if let Some(frontier) = frontier {
-                        frontier > *cap.time()
-                    } else {
-                        true
-                    };
-
+                    let should_write = frontier.map_or(true, |f| f > *cap.time());
                     if should_write {
-                        let write_frontier = if let Some(frontier) = frontier {
-                            frontier
-                        } else {
-                            // There's no way to guarantee this is
-                            // actually the resume epoch on the next
-                            // execution, but mark that this worker is
-                            // ready to resume there on EOF. It's also
-                            // possible that this results in a "too
-                            // small" resume epoch: if for some reason
-                            // this operator isn't activated during
-                            // every epoch, we might miss the
-                            // "largest" epoch and so we'll mark down
-                            // the resume epoch as one too
-                            // small. That's fine, we just might
-                            // resume further back than is optimal.
-                            let resume_epoch = *cap.time() + 1;
-                            resume_epoch
-                        };
+                        // There's no way to guarantee this is
+                        // actually the resume epoch on the next
+                        // execution, but mark that this worker is
+                        // ready to resume there on EOF. It's also
+                        // possible that this results in a "too small"
+                        // resume epoch: if for some reason this
+                        // operator isn't activated during every
+                        // epoch, we might miss the "largest" epoch
+                        // and so we'll mark down the resume epoch as
+                        // one too small. That's fine, we just might
+                        // resume further back than is optimal.
+                        let write_frontier = frontier.unwrap_or(*cap.time() + 1);
                         let msg = ProgressMsg::Advance(WorkerFrontier(write_frontier));
                         let kchange = KChange(worker_key.clone(), Change::Upsert(msg));
                         output_wrapper.activate().session(&cap).give(kchange);
                     }
-
-                    if let Some(frontier) = frontier {
-                        // We should never delay to something like
-                        // frontier + 1, otherwise chained progress
-                        // operators will "drift" forward and GC will
-                        // happen too early.
-                        Some(cap.delayed(&frontier))
-                    } else {
-                        None
-                    }
-                } else {
-                    None
-                };
+                    // We should never delay to something like
+                    // frontier + 1, otherwise chained progress
+                    // operators will "drift" forward and GC will
+                    // happen too early. If the frontier is empty,
+                    // also drop the capability.
+                    frontier.map(|f| cap.delayed(&f))
+                });
 
                 // Wake up constantly, because we never know when
                 // frontier might have advanced.
--- a/src/recovery/operators.rs
+++ b/src/recovery/operators.rs
@@ -14,6 +14,7 @@ use timely::dataflow::channels::pact::Exchange;
 use timely::dataflow::channels::pact::Pipeline;
 use timely::dataflow::operators::flow_controlled::iterator_source;
 use timely::dataflow::operators::flow_controlled::IteratorSourceInput;
+use timely::dataflow::operators::generic::builder_rc::OperatorBuilder;
 use timely::dataflow::operators::*;
 use timely::dataflow::ProbeHandle;
 use timely::dataflow::Scope;
@@ -139,7 +140,7 @@ where
         self.unary(Pipeline, "backup", |_init_cap, _info| {
             move |input, output| {
                 input.for_each(|cap, incoming| {
-                    let epoch = cap.time();
+                    let epoch = SnapshotEpoch(*cap.time());
 
                     assert!(tmp_incoming.is_empty());
                     incoming.swap(&mut tmp_incoming);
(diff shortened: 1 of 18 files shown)


src/recovery/operators.rs
```diff
@@ -179,27 +179,92 @@ where
...
+                        // frontier + 1, otherwise chained progress
+                        // operators will "drift" forward and GC will
+                        // happen too early.
+                        Some(cap.delayed(&frontier))
+                    } else {
+                        None
+                    }
+                } else {
+                    None
+                };
```
Psykopear, line 257: I'd make use of the Option trait more here (`take`, `map`, `map_or`, `unwrap_or_else`, `flatten`), but maybe it's less readable this way?
Just leaving it here as a possible alternative:
```rust
cap = cap
    .take()
    .map(|cap| {
        let frontier = input_frontiers[0].frontier().iter().min().copied();

        let should_write = frontier.map_or(true, |f| f > *cap.time());

        if should_write {
            let write_frontier = frontier.unwrap_or_else(|| *cap.time() + 1);
            let msg = ProgressMsg::Advance(WorkerFrontier(write_frontier));
            let kchange = KChange(worker_key.clone(), Change::Upsert(msg));
            output_wrapper.activate().session(&cap).give(kchange);
        }
        frontier.map(|f| cap.delayed(&f))
    })
    .flatten();
```

src/recovery/store/in_mem.rs
```diff
@@ -331,58 +331,201 @@ fn write_discard_drops_key() {
...
 impl InMemProgress {
-    pub(crate) fn new() -> Self {
-        Self(HashMap::new())
+    /// Init to the "default" progress of a cluster of a given
+    /// size.
+    ///
+    /// You have to specify a worker count so the initial data is
+    /// well-formed.
+    pub(crate) fn new(count: WorkerCount) -> Self {
+        let ex = Into::<i128>::into(<u64 as Timestamp>::minimum()) - 1;
```
Psykopear, line 352: Am I missing something, or could this just be:
```rust
let ex = -1;
```
you are requesting the minimum value of `Timestamp<T>`   which internally calls `T::min_value()` , but since `T` is hardcoded as `u64`, this should always resolve to `0`

src/recovery/store/in_mem.rs
```diff
@@ -331,58 +331,201 @@ fn write_discard_drops_key() {
...
+        let mut frontiers = HashMap::new();
+        for worker in count.iter() {
+            frontiers.insert(worker, WorkerFrontier(<u64 as Timestamp>::minimum()));
+        }
```
Psykopear, line 356: This could be:

```rust
        let frontiers = count
            .iter()
            .map(|worker| (worker, WorkerFrontier(<u64 as Timestamp>::minimum())))
            .collect();
```

Also, since I see you do this more than once, it could be nice to have:

```rust
impl WorkerFrontier {
    pub fn minimum() -> Self {
        Self(<u64 as Timestamp>::minimum())
    }
}
```

So that you can write:
```rust
        let frontiers = count
            .iter()
            .map(|worker| (worker, WorkerFrontier::minimum()))
            .collect();
```

src/recovery/store/in_mem.rs
```diff
@@ -331,58 +331,201 @@ fn write_discard_drops_key() {
...
+        let in_ex: i128 = ex.0.into();
+        // Execution should never regress because the entire cluster
+        // is shut down before resuming.
+        assert!(!(in_ex < self.ex), "Execution regressed");
+        if in_ex > self.ex {
+            self.ex = in_ex;
+            self.frontiers.clear();
+            for worker in count.iter() {
+                self.frontiers.insert(worker, WorkerFrontier(epoch.0));
+            }
```
Psykopear, line 377: Similar to the above, you could write:
```rust
            self.frontiers = count
                .iter()
                .map(|worker| (worker, WorkerFrontier(epoch.0)))
                .collect();

```

Not sure it doing this way implies an additional allocation though

src/recovery/store/in_mem.rs
```diff
@@ -331,58 +331,201 @@ fn write_discard_drops_key() {
...
+        let resume_epoch = ResumeEpoch(
+            self.frontiers
                 .values()
                 .min()
-                .cloned()
-                .map(|border| border.0 + 1)
-                .unwrap_or_else(<u64 as Timestamp>::minimum),
-        )
+                .copied()
+                .unwrap_or(WorkerFrontier(<u64 as Timestamp>::minimum()))
```
Psykopear, line 426: `WorkerFrontier::minimum()`

src/recovery/store/sqlite.rs
```diff
@@ -298,46 +323,88 @@ impl SqliteProgressWriter {
...
-        let WorkerKey(worker_index) = worker_key;
+        let KChange(key, change) = kchange;
+        let WorkerKey(ex, index) = key;
 
         match change {
-            Change::Upsert(epoch) => {
-                let sql = format!("INSERT INTO {} (worker_index, border) VALUES (?1, ?2) ON CONFLICT (worker_index) DO UPDATE SET border = EXCLUDED.border", self.table_name);
+            Change::Upsert(msg) => match msg {
+                ProgressMsg::Init(count, epoch) => {
+                    let sql = format!("INSERT INTO {} (execution, worker_index, worker_count, resume_epoch) VALUES (?1, ?2, ?3, ?4) ON CONFLICT (execution, worker_index) DO UPDATE SET worker_count = EXCLUDED.worker_count, resume_epoch = EXCLUDED.resume_epoch", self.execution_table_name);
```
Psykopear, line 362: you could split the query strings into multiple lines:
```rust
                    let sql = format!(
                        "INSERT INTO {} (execution, worker_index, worker_count, resume_epoch) \
                        VALUES (?1, ?2, ?3, ?4) \
                        ON CONFLICT (execution, worker_index) \
                        DO UPDATE SET \
                        worker_count = EXCLUDED.worker_count, resume_epoch = EXCLUDED.resume_epoch",
                        self.execution_table_name
                    );
```

Psykopear (approved):
Looks good to me.
I added some non blocking comments about writing style, feel free to ignore them if you don't like them.

src/recovery/operators.rs
davidselassie: Oh great! Thanks. I forget about `take` and `map_or` and that makes it easier.

src/recovery/store/in_mem.rs
davidselassie: Not in a hot loop anyway, so easier to read.

src/recovery/store/in_mem.rs
davidselassie: I was hoping that explicitly using `Timestamp::minimum` would help explain why we want the number `-1` but perhaps the type noise here makes it confusing. I don't like the idea of having hard-coded magic numbers, but I guess that's basically what's happening here no matter what.

whoahbot (approved):

src/recovery/store/sqlite.rs
```diff
@@ -157,17 +180,17 @@ impl SqliteStateWriter {
...
         tracing::debug!("Opening Sqlite connection to {db_file:?}");
-        let mut conn = rt.block_on(future).unwrap();
+        let conn = rt.block_on(future).unwrap();
 
         // TODO: SQLite doesn't let you bind to table names. Can
         // we do this in a slightly safer way? I'm not as worried
         // because this value is not from items in the dataflow
         // stream, but from the config which should be under
         // developer control.
         let sql = format!("CREATE TABLE IF NOT EXISTS {table_name} (step_id TEXT, state_key TEXT, epoch INTEGER, snapshot BLOB, PRIMARY KEY (step_id, state_key, epoch));");
```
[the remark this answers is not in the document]
davidselassie: Oh good call. I'll eventually look into that.

PR #198, opened 2023-02-02
Add `filter_map` operator

Psykopear:
Since I had this in a local branch, I'm opening a PR with a new operator, `filter_map`, that behaves like a map, but filters out any item that is `None`.

### DIFF
--- a/pytests/test_operators.py
+++ b/pytests/test_operators.py
@@ -44,6 +44,27 @@ def test_map():
     assert sorted(out) == sorted([1, 2, 3])
 
 
+def test_filter_map():
+    flow = Dataflow()
+
+    inp = [0, 1, 2, 3, 4, 5]
+    flow.input("inp", TestingInputConfig(inp))
+
+    def make_odd(item):
+        if item % 2 != 0:
+            return None
+        return item + 1
+
+    flow.filter_map(make_odd)
+
+    out = []
+    flow.capture(TestingOutputConfig(out))
+
+    run_main(flow)
+
+    assert sorted(out) == sorted([1, 3, 5])
+
+
 def test_flat_map():
     flow = Dataflow()
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ __Add any extra change notes here and we'll put them in the release
 notes on GitHub when we make a new release.__
 
 - Added `HoppingWindowConfig` for windowing operators.
+- Add `filter_map` operator.
 
 ## 0.15.0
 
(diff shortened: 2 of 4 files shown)


davidselassie (approved):
Thanks for this!

pytests/test_operators.py
```diff
@@ -44,6 +44,27 @@ def add_one(item):
...
+
+    inp = [0, 1, 2, 3, 4, 5]
+    flow.input("inp", TestingInputConfig(inp))
+
+    def make_odd(item):
+        if item % 2 != 0:
+            return None
+        return item + 1
+
+    flow.filter_map(make_odd)
```
ariskk, line 58: `flat_map` doesn't do that?!

It should in theory. Example of how flatMap behaves in Scala's collections
```scala
scala> val l = List(0, 1, 2, 3, 4, 5)
l: List[Int] = List(0, 1, 2, 3, 4, 5)

scala> def makeOdd(in: Int) = if (in % 2 != 0) None else Some(in + 1)
makeOdd: (in: Int)Option[Int]

scala> l.flatMap(makeOdd)
res2: List[Int] = List(1, 3, 5)
```

Psykopear: Hi! Thanks for the review.

It looks like what `Scala` calls `flatMap` is called `filter_map` in [Rust iterators](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter_map).

In Rust, the concept of "flattening" usually means to remove one level of nesting.
The `.flatten()` method is implemented for both iterators ([docs](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.flatten)) and `Option` ([docs](https://doc.rust-lang.org/std/option/enum.Option.html#method.flatten)).
`flat_map` on iterators is equivalent to `.map(f).flatten()` ([docs](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.flat_map)).

So it seems there is a difference in the meaning of the terms.
Being based on `timely-dataflow`, our own `flat_map` behaves like the underlying implementation (and Rust's interpretation): it takes an iterator and yields the items inside the iterator.

If I got this right, the `filter_map` operator introduced here behaves like Scala's collections `flatMap` you describe.

I do think that it makes sense for `bytewax` to use the conventions of the underlying ecosystem, so I'm going to merge this PR as it is, as changing the names of other operators is out of scope for this PR.

But I understand that switching names between development environments is not ideal, so if you want to open a discussion about this please open a new issue on this repo.

ariskk: Thanks for the explanation! That's completely up to you, I just got tripped up (coming from Flink) and thought to flag it.

Psykopear: Maybe we could add a page in the documentation explaining differences with Scala/other environments? I think you won't be the only one tripping up about those differences, so it might make sense. I'll talk about this with the team.

PR #203, opened 2023-02-16
Session window

### LINKED ISSUE #174: [FEATURE] Support session window
**Is your feature request related to a problem? Please describe.**

https://learn.microsoft.com/en-us/azure/stream-analytics/stream-analytics-window-functions#session-window

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Psykopear:
This PR introduces `SessionWindow`.

I decided to do something similar to what flink does, which is creating a new session every time an event comes, then immediately merge all the sessions that can be merged, and only after that assign the item to the appropriate session.

I named it `SessionWindow` rather than `SessionWindowConfig` since we plan to remove the `Config` suffix from our structures, but maybe it's better to keep it consistent and call it `SessionWindowConfig` until we fix them all (breaking the python api)?

Closes #174

### DIFF
--- a/pytests/test_window.py
+++ b/pytests/test_window.py
@@ -2,9 +2,65 @@ from datetime import datetime, timedelta, timezone
 
 from bytewax.dataflow import Dataflow
 from bytewax.execution import run_main
-from bytewax.outputs import TestingOutputConfig
 from bytewax.testing import TestingInput
-from bytewax.window import EventClockConfig, HoppingWindowConfig, TumblingWindowConfig
+from bytewax.outputs import TestingOutputConfig
+from bytewax.window import (
+    EventClockConfig,
+    HoppingWindowConfig,
+    TumblingWindowConfig,
+    SessionWindow,
+)
+
+
+class Dt(datetime):
+    def after(self, s):
+        return self + timedelta(seconds=s)
+
+
+def test_session_window():
+    start_at = Dt(2022, 1, 1, tzinfo=timezone.utc)
+
+    flow = Dataflow()
+
+    inp = [
+        # Session 1
+        ("ALL", {"time": start_at.after(1), "val": "a"}),
+        ("ALL", {"time": start_at.after(5), "val": "b"}),
+        # Session 2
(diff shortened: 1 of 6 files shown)


src/window/session_window.rs
```diff
@@ -0,0 +1,359 @@
...
+        // If the item time is before the watermark we mark it late
+        // before creating a new session.
+        if *item_time < *watermark {
+            // We need to pass a window_key to the "Late" error, but in
+            // this case we can't really say what session the item should
+            // have belong to, because it's presence might have changed
+            // sessions that we already dropped.
+            // So I'm just going to assign `self.max_key` here as a placeholder.
+            // XXX: Is this reasoning right, or should we try to get a more
+            //      meaningful number here?
```
whoahbot, line 164: I think it might be interesting to revisit this topic once we have the concept of a non-linear dataflow. Late events could be relegated to a separate stream and handled independently.

whoahbot (approved):
Looks good to me.

Left you some thoughts about late item handling.

pytests/test_window.py
```diff
@@ -4,7 +4,60 @@
...
+    def after(self, s):
+        return self + timedelta(seconds=s)
+
+
+def test_session_window():
+    start_at = Dt(2022, 1, 1, tzinfo=timezone.utc)
+
+    flow = Dataflow()
+
+    def gen():
```
davidselassie, line 25: Could you re-phrase this?

```py
inp = [
    ("ALL", ...),
    ...
]

flow.input("inp", TestingInputConfig(inp))
```

because we gave up on trying to futz with a global mocked clock and use event time, we don't actually need the whole "generator builder" pattern. And I'll be getting rid of the `TestingBuilderInputConfig` in my other PR.

src/window/session_window.rs
davidselassie: FWIW, that was my plan for all window operators too, add a second "late" output. Dropping them now is fine.

davidselassie (approved):
Looks great! Thanks!

PR #204, opened 2023-02-16
Partitioned input

davidselassie:
Changes the input API to be partition based.

Subclass `bytewax.inputs.CustomPartInput` to make your own input.

On the Rust side, gets rid of the `InputReader` interface and instead
uses the Python objects directly. I think this will have negligible
performance impact. It does mean I re-implemented the Kafka connector
in Python. I also moved it into `bytewax.connectors.kafka` so we can
have individual submodules that have special pip deps.

Recovery-wise, we use the partition key as the state key, and then
hash route it back to workers in the new cluster. We use that hash
routing during partition input building so state is routed
correctly. This means that the epoch sources now have to have more
state managment code, but they're basically the same.

I've also added a `bytewax.connectors.files` with some simple
line-by-line file based inputs. This makes many of the examples way
more straightforward.

I've tried to update all the examples to work with this new API, but I
haven't updated any of the docs. I want to do this same transformation
for output first, then go through everything.

There's also a sprinkling of changing `StringResult` to `PyResult` to
make all the errors work out. I think as we tie Python more deeply
into the dataflow building, we'll need it more anyway.


examples/utils/topics_helper.py
```diff
@@ -1,28 +1,29 @@
...
 from time import sleep
 
-import fake_events
+# pip install pandas confluent-kafka fake-web-events
 import pandas as pd
-from kafka import KafkaAdminClient, KafkaProducer
-from kafka.admin import NewTopic
+from confluent_kafka import Producer
+from confluent_kafka.admin import AdminClient, NewTopic
+from fake_web_events import Simulation
```
Psykopear, line 12: This was imported but not used. See comment below

examples/wikistream.py
```diff
@@ -8,17 +8,27 @@
...
 from bytewax.recovery import SqliteRecoveryConfig
 from bytewax.window import SystemClockConfig, TumblingWindowConfig
 
 
-def input_builder(worker_index, worker_count, resume_state):
-    # Multiple SSE connections will duplicate the streams, so only
-    # have the first worker generate input.
-    if worker_index == 0:
+class WikiStreamInput(CustomPartInput):
+    def list_keys(self):
```
Psykopear, line 19: Shouldn't this be `list_parts`? It fails as it is

examples/wikistream.py
```diff
@@ -8,17 +8,27 @@
 
 from bytewax import parse
 from bytewax.dataflow import Dataflow
-from bytewax.execution import spawn_cluster
-from bytewax.inputs import ManualInputConfig
+from bytewax.execution import run_main, spawn_cluster
+from bytewax.inputs import CustomPartInput
 from bytewax.outputs import StdOutputConfig
 from bytewax.recovery import SqliteRecoveryConfig
```
Psykopear, line 14: Remove unused import

examples/wikistream.py
```diff
@@ -68,8 +75,8 @@ def keep_max(max_count, new_count):
 
 
 if __name__ == "__main__":
-    spawn_cluster(
+    run_main(
         flow,
-        recovery_config=SqliteRecoveryConfig("."),
-        **parse.cluster_args(),
+        # recovery_config=SqliteRecoveryConfig("."),
+        # **parse.cluster_args(),
```
Psykopear, line 81: Remove the commented lines

pysrc/bytewax/connectors/kafka.py
```diff
@@ -0,0 +1,106 @@
...
+
+from confluent_kafka import Consumer, KafkaError, OFFSET_BEGINNING, TopicPartition
+from confluent_kafka.admin import AdminClient
+
+from bytewax.inputs import CustomPartInput
+
+
+def _stateful_read_part(consumer):
+    while True:
+        msg = consumer.poll(0.001)  # seconds
```
Psykopear, line 14: This seems to greatly alleviate the spinning problem on the kafka input, but I think it's can still be present in some other kind of inputs, it would be good to do this at the timely level probably. But that's for another PR.

src/execution/mod.rs
```diff
@@ -147,6 +132,10 @@ where
     resume_progress.write(progress_init.clone());
     progress_writer.write(progress_init);
 
+    // Remember! Never build different numbers of Timely operators on
+    // different workers! Timely does not like that and you'll see a
+    // mysterious `failed to correctly cast channel` panic. You must
+    // build asymmetry within each operator.
```
Psykopear, line 138: Thanks for adding the full error and explanation here, it will be nice next time someone hits the error :)

src/execution/mod.rs
```diff
@@ -411,26 +425,31 @@ fn run_until_done<A: Allocate, T: Timestamp>(
...
-) {
+) -> PyResult<()> {
     let mut span = PeriodicSpan::new(Duration::from_secs(10));
     while !interrupt_flag.load(Ordering::Relaxed) && !probe.done() {
         worker.step();
         span.update();
+        Python::with_gil(|py| Python::check_signals(py)).map_err(|err| {
+            interrupt_flag.store(true, Ordering::Relaxed);
+            err
+        })?;
```
Psykopear, line 436: Is this the piece that makes the error on `ctrl+c` finally readable? Good catch :+1:

src/inputs.rs
```diff
@@ -0,0 +1,209 @@
...
+            tracing::info!("{index:?} building input {step_id:?} partition {key:?} with resume state {state:?}");
+            let iter: Py<PyIterator> = self
+                .0
+                .call_method1(py, "build_part", (key.clone(), state.clone_ref(py)))?
+                .as_ref(py)
+                .iter()?
+                .into();
+            let part = PartIter { iter, state };
+            parts.insert(key, part);
+        }
```
Psykopear, line 106: :D Are you completely sure about that first comment?
Not saying you should, but Result [implements FromIterator](http://doc.rust-lang.org/std/result/enum.Result.html#method.from_iter), so you can move the Result outside and iterators will take care of the rest (including stopping iteration if an error is found):

```rust
        let parts = keys
            .into_iter()
            .filter(|key| key.is_local(index, worker_count))
            .map(|key| {
                let state = resume_state
                    .remove(&key)
                    .map(StateBytes::de::<TdPyAny>)
                    .unwrap_or_else(|| py.None().into());
                tracing::info!("{index:?} building input {step_id:?} partition {key:?} with resume state {state:?}");
                let iter: Py<PyIterator> = self
                    .0
                    .call_method1(py, "build_part", (key.clone(), state.clone_ref(py)))?
                    .as_ref(py)
                    .iter()?
                    .into();
                let part = PartIter { iter, state };
                Ok((key, part))
            }).collect::<PyResult<HashMap<StateKey, PartIter>>>()?;

```

Psykopear (approved):
I really like this PR. Overall looks good to me, I added a number of comments on the examples, some of the issues were not introduces in this PR, but since you are already working on them I think it's good to fix everything at this point.

Also, the `bigquery_out.py` example is missing the fix for the input (still using ManualInputConfig)

examples/apriori.py
```diff
@@ -3,21 +3,13 @@
 from collections import Counter
 from datetime import timedelta
 
+from bytewax.connectors.files import FileInput
 from bytewax.dataflow import Dataflow
 from bytewax.execution import run_main, spawn_cluster
-from bytewax.inputs import ManualInputConfig
 from bytewax.outputs import ManualOutputConfig, StdOutputConfig, TestingOutputConfig
```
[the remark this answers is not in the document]
davidselassie: Hmm. I'm surprised pre-commit is not catching this... Let me look into why it's not running on the examples?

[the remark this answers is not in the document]
davidselassie: I'm going to fix all these and enable linting in these dirs in another PR.

src/inputs.rs
davidselassie: This is great.

pysrc/bytewax/connectors/kafka.py
davidselassie: Yes. Async IO will be a big project, but hopefully this will make it acceptable for the time being.

src/execution/mod.rs
davidselassie: My issue here was that the `run_main` ctrl-c test was failing and I wasn't sure why. Especially because I could ctrl-c a long running example dataflow manually. I think this fixed it? I think there's something deeper we don't understand about signal management here. But if it works for now...

PR #205, opened 2023-02-16
Fix "interleaved executions"

davidselassie:
Turns out they're ok. During resume, each worker's progress events are
in order. But! Because we're reading all the progress logs in parallel
and broadcasting them, across workers they might be out-of-order. Thus
we might do something like read all of worker 1's progress, then all
of worker 2's. Execution init messages are stored on all workers, so
the resume cluster might already have advanced to the final exeuction,
and then go "back in time" to see a slow reader's log. That's fine
because we already know the next exeuction started and thus can ignore
those progress messages.

Re-works some of the recovery tests to crash and restart twice to
start to probe this.

### DIFF
--- a/src/recovery/store/sqlite.rs
+++ b/src/recovery/store/sqlite.rs
@@ -415,44 +415,55 @@ impl SqliteProgressReader {
                             .expect("SQLite int can't fit into epoch; might be negative"),
                     );
                     let msg = ProgressMsg::Init(count, epoch);
-                    KChange(key, Change::Upsert(msg))
+                    (key, msg)
                 })
                 .fetch(&conn)
                 .map(|result| result.expect("Error selecting from SQLite"));
 
-            while let Some(kchange) = stream.next().await {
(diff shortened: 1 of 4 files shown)


whoahbot (approved):

Psykopear (approved):

src/recovery/store/sqlite.rs
```diff
@@ -425,7 +425,10 @@ impl SqliteProgressReader {
                 tx.send(kchange).await.unwrap();
             }
 
-            let sql = "SELECT execution, worker_index, frontier FROM progress";
+            // TODO: We really could only select progress from the
+            // latest execution.
+            let sql =
+                "SELECT execution, worker_index, frontier FROM progress ORDER BY execution ASC";
```
Psykopear, line 431: Why don't we?

davidselassie: Because I was tired yesterday. But I am not today. Done below.

PR #210, opened 2023-03-01
Allows KafkaInput to take multiple topics

davidselassie:
It will multiplex them all at that point in the dataflow.

Adds a unit test for this, although since we don't have a Kafka
cluster in CI, skips it for now.

Also allows `PartInput.list_parts` to return an iterable.

### DIFF
--- a/pysrc/bytewax/connectors/kafka.py
+++ b/pysrc/bytewax/connectors/kafka.py
@@ -25,8 +25,8 @@ def _stateful_read_part(consumer):
 
 
 class KafkaInput(PartInput):
-    """Use a single [Kafka](https://kafka.apache.org) topic as the
-    input source.
+    """Use [Kafka](https://kafka.apache.org) topics as an input
+    source.
 
     Kafka messages are emitted into the dataflow as two-tuples of
     `(key_bytes, payload_bytes)`.
@@ -37,7 +37,7 @@ class KafkaInput(PartInput):
 
         brokers: List of `host:port` strings of Kafka brokers.
 
-        topic: Topic to consume from.
+        topics: List of topics to consume from.
 
         tail: Whether to wait for new data on this topic when the end
             is initially reached.
@@ -57,7 +57,7 @@ class KafkaInput(PartInput):
     def __init__(
         self,
         brokers: Iterable[str],
-        topic: str,
+        topics: Iterable[str],
         tail: bool = True,
         starting_offset: int = OFFSET_BEGINNING,
         add_config: Dict[str, str] = None,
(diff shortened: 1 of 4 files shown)


pysrc/bytewax/connectors/kafka.py
```diff
@@ -76,18 +76,22 @@ def list_parts(self):
         }
         config.update(self.add_config)
         client = AdminClient(config)
-        cluster_metadata = client.list_topics(self.topic)
-        part_idxs = cluster_metadata.topics[self.topic].partitions.keys()
+        cluster_metadata = client.list_topics()
 
-        return [f"{i}-{self.topic}" for i in part_idxs]
+        for found_topic in self.topics:
```
whoahbot, line 81: Should we do some error checking for the list of topics in this connector? If the topic hasn't been created yet, you'll get a non-specific `ValueError`.

You also get some interesting errors if you supply a `String` instead of a `List` where you end up iterating the string character by character.

whoahbot (approved):

src/inputs.rs
```diff
@@ -71,8 +71,17 @@ impl PartInput {
...
+            let mut keys = self
+                .0
+                .call_method0(py, "list_parts")?
+                .as_ref(py)
+                .iter()?
+                .map(|i| i.and_then(PyAny::extract::<StateKey>))
+                .collect::<PyResult<Vec<StateKey>>>()?;
+            keys.sort();
+            keys
+        };
```
Psykopear, line 84: you could collect to a `BTreeSet` to avoid the sort, the rest of the code should work as it is (if we don't expect duplicate keys):
```rust
        let keys = self
            .0
            .call_method0(py, "list_parts")?
            .as_ref(py)
            .iter()?
            .map(|i| i.and_then(PyAny::extract::<StateKey>))
            .collect::<PyResult<BTreeSet<StateKey>>>()?;
```

Psykopear (approved):
Looks good to me.
I like the kafka test. It would be nice to have a separate step for "integration tests" with external services, mostly for connectors and possibly for recovery, and having a working test in the codebase is a good first step, even if we don't run it in CI yet

src/inputs.rs
Psykopear: Also, not related to this PR, but I noticed this while reviewing, at the end of this function we do:
```rust
let remaining_keys: Vec<StateKey> = resume_state.into_keys().into_iter().collect();
assert!(remaining_keys.is_empty(), "Partition keys in resume state that are not in partition list; changing partition counts? recovery state routing bug?");
```

Given that we drop `resume_state` anyway at the end of the function, we could proxy the `is_empty` method of the hashmap in `StepStateBytes`:
```rust
impl StepStateBytes {
    pub(crate) fn is_empty(&self) -> bool {
        self.0.is_empty()
    }
}
```
And then do:
```rust
assert!(resume_state.is_empty(), "Partition keys in resume state that are not in partition list; changing partition counts? recovery state routing bug?");
```

davidselassie: Great suggestions, thank you.

PR #211, opened 2023-03-02
Partitioned output

davidselassie:
Refactors output system to look like the partitioned input system.

Python-side, there's a `PartOutput` and `DynamicOutput` that are the
two base output types. All output types derive from them. There are
now two output operators, `part_output` and `dynamic_output` which you
use to add an output at that point in the dataflow. They are two
separate operators because they have different semantics and input
data shapes.

Partitioned output is stateful and has the cluster list all partitions
then will divvy them up to workers. Since consistent routing of data
is crucial, the input to the `part_output` operator needs to look like
`(key, value)` 2-tuples and the operator will route to a consistent
partition. Since each partition is independent, we can use the
standard pattern of having recovery state that is routed back to the
constructor of that partition on resume.

Dynamic output is stateless output that runs on each worker in the
current cluster. We explicitly say "we can't store state" and so make
it obvious that you can not reach the same level of delivery
guarantees as with the other output forms. Since the semantics of how
to key data happen somewhere downstream, the input to the
`dynamic_output` operator has no constraints, items are written
directly to the output on the current worker.

Re-writes all of the existing output config types to use one of these
two base types. Perhaps counter-intuitively `KafkaOutput` is a
`DynamicOutput` because the hashing to partitions happens via the
producer and there are no limit on how many producers you can have.

Rust-side, tries getting rid of layers of abstraction and ends up
taking the Python output definitions directly into Timely
operators. This feels right because we directly want different
dataflow semantics (aka Timely operators) for each type of output.


Psykopear:
Nice!

The only thing I don't like is the 2 separate operators for output in the python api.
The distinction is needed on our side, but I'm not sure we need to expose it this way to users.

Users will need to understand the implications of the different output configurations beforehand anyway, so forcing them to change the operator if they go from one type to the other doesn't make things easier.
Errors in the shape of the data will manifest the first time you run the dataflow, and wrong assumptions about the semantics of the output config won't be mitigated by using a differently named operator in my opinion.

Is it possible to have a single `flow.output` operator, and make the distinction between dynamic and part outputs later?
Something like using a common subclass and only downcast it when building the dataflow.

src/dataflow/mod.rs
```diff
@@ -91,34 +92,49 @@ impl Dataflow {
...
+        self.steps.push(Step::PartOutput { step_id, output });
+    }
+
+    /// Write data to a dynamic output.
+    ///
+    /// At least one output is required on every dataflow.
+    ///
+    /// Emits no items downstream.
+    ///
+    /// Each worker will  See `bytewax.outputs.DynamicOutput` for more information on
```
whoahbot, line 126: Remove `Each worker will` ?

davidselassie:
My original thinking was that the different output operators was to hint at someone writing a dataflow that they're going to need different upstream data shapes: `(key, value)` or non-keyed.

But thinking about it more after meeting, I agree with you. Each specific output type, irrespective of if it's partitioned or not, still further introduces arbitrary requirements on the upstream data (like the Dynamo connector being a special dictionary), so even though different operators might help remind you of the "class" of shape, you still need to be careful and think about the details anyway.

Let me work on collapsing it back down to one operator and switching internally when building the Timely dataflow.

src/execution/mod.rs
```diff
@@ -346,26 +345,44 @@ where
...
+                            step_id,
+                            output,
+                            worker_index,
+                            worker_count,
+                            step_resume_state,
+                        )?;
+
+                    outputs.push(clock.clone());
+                    step_changes.push(changes);
+                    // Reset the stream to an empty stream.
```
whoahbot, line 363: Interesting. Why do we reset the stream to an empty stream here?

whoahbot (approved):
Looks good! I left you a few non-blocking questions and some copy edits.

pysrc/bytewax/outputs.py
```diff
@@ -1,82 +1,169 @@
...
+        resume_state: Optional[Any],
+    ) -> Optional[StatefulSink]:
+        """Build an output partition, resuming writing at the position
+        encoded in the resume state.
 
-class TestingOutputConfig(ManualOutputConfig):
-    """
-    Append each output item to a list.
-    You only want to use this for unit testing.
+        Will be called once within each cluster for each partition
```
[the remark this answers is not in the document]
davidselassie: I mean all workers in an execution. I'm trying to describe that Bytewax will take care of distributing this across all workers. Let me try again.

src/execution/mod.rs
davidselassie: I wanted the output operator to "end the line" of data transformations and not emit anything downstream. Unlike previously where capture was more like a "tap" which still emitted items downstream. But perhaps that is best left until we tackle non-linear dataflows. I'm going to revert it back to that.

PR #214, opened 2023-03-09
Source and sink classes, epoch config removal

davidselassie:
Instead of using specially crafted closures or generators to represent
how to read / write the next item of input / output, subclass
`{Stateful,Stateless}{Source,Sink}` and follow the instructions on
which methods to override. This hopefully will be a little less
delicate as the steps you need to take of "next" and "snapshot" are
explicit and separated. Also I like the parallel construction between
input and output now.

Also refactors the Rust-side input module to match the output module:
a single file that unifies the Python ABCs, the Timely operators, and
the internal structures needed to run them. Each input type
(partitioned or dynamic) is represented by its own Timely operator
because they have inherently different shapes and behaviors.

This does mean, though, I ripped out the concept of `EpochConfig`s and
how they map to input operators, since that's at odds with the input
type determining the Timely operator shape. Instead adds a top level
`epoch_interval` argument to our entry points to control how often
epochs are emitted. The above input Timely operators all have
"periodic epoch" semantics. Inspired from a comment from Fede, an
interval of zero is effectively "a new epoch each item" so we use that
for the existing recovery tests.

I eventually want to re-think how we test the recovery system to be
less "emergent", which I think will involve having more precisely
targeted unit tests. One of the problems here is that to fully control
the epoch progression of some test input, we sort of have to not use
our normal input operators. That means you can't really test the full
stack which doesn't feel right. All of that is for another day,
though. I have added some unit tests that probe the exact recovery
state for the stateful inputs which makes me feel more confident about
them.


Psykopear (approved):
Looks good to me. I like the way input and outputs now work under the same model.
I added a single comment about a docstring that seems off.

Psykopear (changes requested):
On a closer inspection, something seems wrong.

I tried to build a `DynamicInput` for the wikistream example, and the build method on the Python side requires `worker_index` and `worker_count` args, but they are not passed when calling the function from the rust side.

If I'm getting this right, the fix should be on the rust side to add the missing arguments.
We should probably add a test for this.

Another question, which is related to the latest PRs but only came to mind now is, why don't we expose `DynamicInput`  (and all the other base classes) from rust as a subclassable python class?
Obviously you had reasons, but I'm curious about them.
Can't add an "abstractmethod" with pyo3? Or you just wanted this to be on the python side for users to see?

pysrc/bytewax/inputs.py
```diff
@@ -8,41 +8,104 @@
...
     def list_parts(self) -> Set[str]:
-        """List all partitions for this input by a string key.
+        """List all partitions by a string key.
 
-        This must consistently return the same keys when called by any
-        worker in a cluster.
+        This must consistently return the same keys when called by all
+        workers in all executions.
 
         Keys must be unique within this dataflow step.
```
whoahbot, line 110: It might be helpful to expand on this docstring. I'm not sure where else we reference executions outside of the internals of our recovery system. It's also a little unclear (at least to me) what we mean when we say `called by all workers`.

pysrc/bytewax/inputs.py
```diff
@@ -58,18 +121,17 @@ def build_part(
...
         for_part: str,
         resume_state: Optional[Any],
-    ) -> Optional[PartIter]:
-        """Build an input source for a given partition, resuming from
-        the position encoded in the resume state.
+    ) -> Optional[StatefulSource]:
+        """Build an input partition, resuming from the position
+        encoded in the resume state.
+
+        Will be called once within each cluster for each partition
```
whoahbot, line 128: Should we combine these two paragraphs?

pysrc/bytewax/inputs.py
davidselassie: We're going to need to do a pass over all the documentation and at some point reference "executions" publicly in a few places, as the concept can't fully be hidden.

Perhaps a better way of describing this is "It will be called once on each worker during each execution. It must consistently return the same keys every call."

I think we should add a Glossary page to our docs. But that's for a later pass.

pysrc/bytewax/inputs.py
davidselassie: Almost every other docstring for operators, etc, has a separate paragraph answering "how often is this called" so I was keeping consistent with that pattern.

davidselassie:
I fixed the issue with forgetting to pass `worker_index` and `worker_count` to `DynamicInput.build`.

davidselassie:
Re: 

> Another question, which is related to the latest PRs but only came to mind now is, why don't we expose `DynamicInput` (and all the other base classes) from rust as a subclassable python class? Obviously you had reasons, but I'm curious about them. Can't add an `@abstractmethod`, or use ABC stuff with pyo3? Or you just wanted this to be on the python side for users to see?

I sorta forgot that there was the other option, so I tried prototyping out today what it would look like to define `Input` and `Source` and friends as `pyclass`es. And I'm not sure I like it because of the dynamism we want.

Pros:

- Docstrings and definitions for the interfaces are right in the Rust code, so you can see them when you're working on the operators and the calling code.
- You get the PyO3 "eager type checking" for free.

Wishy-wash:

- You can't use `ABC`. You can do something sort of similar and raise `NotImplementedError` in every method. Although until we remove pickling, we aren't using `ABC` anyway.

Cons:

- You don't actually get any type checking on the any of the interface methods themselves! We need dynamic dispatch to a runtime-defined subclass. This means we have to use the `call_method("name", (args, ) ...)` functions; you _can't_ use something like `py_cell.borrow().build_part(key, resume_state)` with Rust-side type checking because the latter will statically dispatch to the parent class method compiled in Rust (which always returns `NotImplementedError`). So you're using `extract` and hoping the args and return value types match up with the Rust superclass definition anyway. This means you're repeating and can make mistakes matching the types.
- And since you are exporting the abstract functions via PyO3, you need to figure out `FromPyObject` for all the return types on the classes (which is tricky for our "co-operative async iterator return value").
- Although one day this might get fixed in PyO3, it also means today that type annotations and rich function signatures are missing for these classes.

To me, it kinda feels like you get a mixture of both the downsides of having to call dynamic methods on objects with the non-nativeness of having to define a complex Python parent class via PyO3. I think we should only use classes defined via PyO3 that we are either: never going to subclass, and have significant code we want to call in a method. Otherwise I don't think the overall feel of it is worth it.

I am going to fix up some of the type bugs that I found through this experiment.

Psykopear:
> To me, it kinda feels like you get a mixture of both the downsides of having to call dynamic methods on objects with the non-nativeness of having to define a complex Python parent class via PyO3. I think we should only use classes defined via PyO3 that we are either: never going to subclass, and have significant code we want to call in a method. Otherwise I don't think the overall feel of it is worth it.

Yep, given that cons you listed, I agree it's better as it is.
Looks good to me then :+1:

Psykopear (approved):

davidselassie:
Two more thoughts on this:

If we do end up going this route more often with other "config" types, I think we should make a macro like `FromPyABC` which does the "check this is a subclass of the ABC" to reduce that boilerplate.

For posterity https://gist.github.com/davidselassie/a4d9749445318965ec6790a49dac8020 has the `input.rs` defining the "ABC" classes so you can see what it feels like. And if we need to go that route in the future: use `PyCell<$PYCLASS>::borrow` is basically the trick.

whoahbot (approved):

PR #216, opened 2023-03-17
Error handling

Psykopear:
This PR makes error handling a bit better on Bytewax.

The main goal is to be able to pinpoint the cause of any possible error that happens during the build and execution of a Dataflow.
Previously if an error occurred in the Python code (whether user's code or our own), we would just see a traceback pointing to `run_main` or `spawn_cluster`, without further info.
Since during the lifetime of the dataflow we do have the info, this PR introduces a trait to easily create a chain of errors with location tracking (using  the [`#[track_caller]`](https://doc.rust-lang.org/reference/attributes/codegen.html#the-track_caller-attribute) attribute) that can be bubbled up directly to the Python interpreter.

I made a pass over the whole codebase to use the trait wherever it made sense. I might have missed some places, but from now on we know that if we don't get a proper trace, adding a `.reraise("custom message")` or `.raise::<PyCustomException>("custom message")` should be enough to make the error clearer.

I also added a new folder inside `examples`, named `errors`, where we can keep a number of examples that are supposed to break, to check if the error message is useful.
They are examples, not supposed to be run as tests for now, but once the removal of multiprocessing takes place, we could find a way to make those dataflows part of the CI too with some modifications.

Errors when using `spawn_cluster`  and more than one process are still not perfect, but I avoided double panicking. Things will improve when we stop using multiprocessing.

To test the validity of this PR run each broken example in the folder, and also don't hesitate to write all kind of buggy dataflows, or to insert random errors in the Rust codebase to see how it behaves. If you catch a case where the error is not helpful, I can see if it can be improved.


whoahbot (approved):
This is great! Thanks for all of the careful work that went into this.

I left you some non-blocking comments.

src/execution/mod.rs
```diff
@@ -575,22 +603,26 @@ fn worker_main<A: Allocate>(
...
     let (progress_writer, state_writer) = Python::with_gil(|py| {
         build_recovery_writers(py, worker_index, worker_count, recovery_config)
+            .reraise("error building recovery writers")
     })?;
 
     let span = tracing::trace_span!("Resume epoch").entered();
     let resume_progress =
-        build_and_run_progress_loading_dataflow(worker, interrupt_flag, progress_reader)?;
+        build_and_run_progress_loading_dataflow(worker, interrupt_flag, progress_reader)
+            .reraise("error while resuming state")?;
```
davidselassie, line 616:
```suggestion
            .reraise("error while loading recovery progress")?;
```

src/execution/mod.rs
```diff
@@ -575,22 +603,26 @@ fn worker_main<A: Allocate>(
...
     span.exit();
     let resume_from = resume_progress.resume_from();
     tracing::info!("Calculated {resume_from:?}");
 
     let span = tracing::trace_span!("State loading").entered();
     let ResumeFrom(_ex, resume_epoch) = resume_from;
     let (resume_state, store_summary) =
-        build_and_run_state_loading_dataflow(worker, interrupt_flag, resume_epoch, state_reader)?;
+        build_and_run_state_loading_dataflow(worker, interrupt_flag, resume_epoch, state_reader)
+            .reraise("error loading state")?;
```
davidselassie, line 625:
```suggestion
            .reraise("error loading recovery state")?;
```

src/inputs.rs
```diff
@@ -108,32 +109,50 @@ impl PartitionedInput {
...
+                    .and_then(|part| part.extract(py))
+                {
                     Err(err) => Some(Err(err)),
                     Ok(None) => None,
                     Ok(Some(part)) => Some(Ok((key, part))),
                 }
-            }).collect::<PyResult<HashMap<StateKey, StatefulSource>>>()?;
+            })
+            .collect::<PyResult<HashMap<StateKey, StatefulSource>>>()
+            .reraise("error creating parts for input")?;
```
davidselassie, line 146:
```suggestion
            .reraise("error creating input source partitions")?;
```

src/inputs.rs
```diff
@@ -194,7 +213,10 @@ impl PartitionedInput {
                         let mut output_handle = output_wrapper.activate();
                         let mut output_session = output_handle.session(&output_cap);
                         for (key, part) in parts.iter() {
-                            match unwrap_any!(Python::with_gil(|py| part.next(py))) {
+                            match unwrap_any!(Python::with_gil(|py| part
+                                .next(py)
+                                .reraise("error getting part input")))
```
davidselassie, line 218:
```suggestion
                                .reraise("error getting next input item from source partition")))
```

src/inputs.rs
```diff
@@ -53,7 +54,7 @@ impl<'source> FromPyObject<'source> for Input {
             .extract()?;
         if !ob.is_instance(abc)? {
             Err(PyTypeError::new_err(
-                "input must derive from `bytewax.inputs.Input`",
```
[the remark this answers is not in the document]
davidselassie: Yeah, I'd either say `derive from` or `subclass`.

src/inputs.rs
```diff
@@ -320,19 +346,26 @@ impl StatefulSource {
     fn next(&self, py: Python) -> PyResult<Poll<Option<TdPyAny>>> {
         match self.0.call_method0(py, "next") {
             Err(stop_ex) if stop_ex.is_instance_of::<PyStopIteration>(py) => Ok(Poll::Ready(None)),
-            Err(err) => Err(err),
+            Err(err) => Err(err).reraise("error getting next input in stateful source"),
```
davidselassie, line 349: Is it possible to get the traceback into without needing to add a redundant message? Like https://github.com/bytewax/bytewax/pull/216/files#diff-316cbdd361ed963c0aa5d42dc4b0cb2dcb72b4a586da2bc4312164c5379709e9R218 will always be used whenever this error happens, so this extra annotation isn't super helpful noise.

Not a big deal if it's easier for everything to use the same code, though. One redundant message isn't a big deal.

davidselassie (approved):
This looks great! Thank you for figuring this out and I'm excited for my errors to be understandable!

src/inputs.rs
Psykopear: You are right that the reduntant message isn't helping. In fact in this specific case even the location isn't really useful, and I think it's ok to just remove this call to `.reraise`.

So right now we have this error:

```
  File "/home/docler/src/bytewax/bytewax/examples/errors/stateful_input_error.py", line 46, in <module>
    run_main(flow)
ValueError: (src/inputs.rs:218:34) error getting next input item from partition source
Caused by => ValueError: (src/inputs.rs:349:34) error getting next input in stateful source
Caused by => ValueError: A vague error
  File "/home/docler/src/bytewax/bytewax/examples/errors/stateful_input_error.py", line 20, in next
    raise ValueError("A vague error")
```

If we simply avoid adding a new message but still add location info, we'd get something like this (bar a different formatting of the error):

```
  File "/home/docler/src/bytewax/bytewax/examples/errors/stateful_input_error.py", line 46, in <module>
    run_main(flow)
ValueError: (src/inputs.rs:218:34) error getting next input item from partition source
Caused by => ValueError: (src/inputs.rs:349:34)
Caused by => ValueError: A vague error
  File "/home/docler/src/bytewax/bytewax/examples/errors/stateful_input_error.py", line 20, in next
    raise ValueError("A vague error")
```

But looking at the code, this function can only fail if the python code raises an exception.
Removing the `reraise` here and keeping it only at the caller site, we get this:
```

  File "/home/docler/src/bytewax/bytewax/examples/errors/stateful_input_error.py", line 46, in <module>
    run_main(flow)
ValueError: (src/inputs.rs:218:34) error getting next input item from partition source
Caused by => ValueError: A vague error
  File "/home/docler/src/bytewax/bytewax/examples/errors/stateful_input_error.py", line 20, in next
    raise ValueError("A vague error")
```

And if you go to the location, you'll notice that we just unwrapped a fallible function, and looking into it you can immediately see where it failed.
If this function was using `?` to propagate errors, or returning Errs in more than one way, it would make sense to reraise the exception here with different messages, but in this specific case it seems just noise.

So I'd keep the trait as it is unless we find other places where adding location without further info might still be useful (adding a method won't be much work given that they are automatically implemented by the trait),

Psykopear:
Ok, I moved everything to `must subclass`, and applied all the suggestions.
I removed the reduntant `.reraise` call and explained why in the comment, let me know if you agree on that otherwise I'm still in time to make the change (add a `.reraise` equivalent without a custom message)

src/inputs.rs
davidselassie: Sounds good to keep it as is.

davidselassie (approved):

PR #217, opened 2023-03-17
More deterministic windowing

davidselassie:
As part of debugging what was going on with windowing in the
possibly-buggy recovery example, I found some issues with how clocks
and windowers were implemented. This was basically a week of
continually finding more and more obscure off-by-one errors.

First `start_at` is renamed to `align_to`. That former name
incorrectly implied that there were no windows before that time, but
that is not true; windows extend (almost) infinitely in both
directions in time from that alignment point. What determines if an
item is late for a window is not the windowing definition, but the
watermark of the clock.

Next, both `{Sliding,Tumbling}Window` are required to take
`align_to`. Previously, this was filled with the timestamp of the
start of the dataflow, but we did not persist it as the recovery
state, nor can we actually properly persist it because recovery state
is per-state-key but at that point we don't have any key. I didn't
want to work on figuring out a way to store "dataflow global" recovery
state, so to make it easier we require you to pick a fixed alignment
time. If you don't ensure `align_to` is consistent across executions,
you have window boundaries jumping around even when you are using
event time.

Then fixes up and adds additional tests for event time watermarking. I
don't belive we were properly advancing the watermark, preventing it
from going backwards, and handling the extreme no data and EOF cases.

Also does a fix for system time watermarking. Once the data has
completed, we should forever return the watermark as "end of time",
not just when there is an item being processed. I don't believe this
causes any difference in behavior with the way our operators are
currently written, but if we didn't process all available data on each
awakening, this is needed.

Then writes a ton more tests for sliding windowing. I believe we were
improperly mixing in the watermark time into the window
definitions. The definition of the window boundaries is a function of
the arguments to the config, not the current item time or
watermark. The watermark is just used to determine if an item is late.

### DIFF
--- a/src/window/sliding_window.rs
+++ b/src/window/sliding_window.rs
@@ -2,6 +2,7 @@ use std::collections::HashMap;
 
 use chrono::prelude::*;
 use chrono::Duration;
+use num::integer::Integer;
 use pyo3::prelude::*;
 
 use crate::{add_pymethods, window::WindowConfig};
@@ -99,18 +100,14 @@ impl SlidingWindower {
     /// given time. Close time is exclusive.
     fn intersects(&self, time: &DateTime<Utc>) -> impl Iterator<Item = (WindowKey, DateTime<Utc>)> {
         let since_close_of_first_window = *time - (self.align_to + self.length);
-        let first_window_idx = since_close_of_first_window
-            .num_milliseconds()
-            // We always want to round towards -inf. TODO:
-            // i64::div_floor will more obviously do this.
-            .div_euclid(self.offset.num_milliseconds())
-            + 1;
+        let first_window_idx = Integer::div_floor(
+            &since_close_of_first_window.num_milliseconds(),
+            &self.offset.num_milliseconds(),
+        ) + 1;
 
-        // Round up to 1 in case we have offset > length. Always try
-        // at least one window. We might filter it out, though.
-        let num_windows = std::cmp::max(
-            self.length.num_milliseconds() / self.offset.num_milliseconds(),
-            1,
+        let num_windows = Integer::div_ceil(
+            &self.length.num_milliseconds(),
+            &self.offset.num_milliseconds(),
         );
 
         // Clone to not retain ownership of self in the closure.
@@ -194,7 +191,7 @@ impl Windower for SlidingWindower {
 }
 
 #[test]
-fn test_intersect_overlap_bulk_offset_divisible_by_length() {
+fn test_intersect_overlap_offset_divisible_by_length_bulk() {
     let windower = SlidingWindower {
         length: Duration::seconds(10),
         offset: Duration::seconds(5),
@@ -219,7 +216,7 @@ fn test_intersect_overlap_bulk_offset_divisible_by_length() {
 }
 
 #[test]
-fn test_intersect_overlap_negative_offset_divisible_by_length() {
+fn test_intersect_overlap_offset_divisible_by_length_bulk_negative() {
     let windower = SlidingWindower {
         length: Duration::seconds(10),
         offset: Duration::seconds(5),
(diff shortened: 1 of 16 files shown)


davidselassie:
If this description is too confusing, look at the unit tests I added and see if the semantics match up with what you expect. We can discuss that here.

davidselassie:
@whoahbot Have any thoughts on the mysterious `process didn't exit successfully: `/Users/gha/actions-runner/_work/bytewax/bytewax/target/aarch64-apple-darwin/debug/deps/bytewax-99769e2d4f978c9f` (signal: 11, SIGSEGV: invalid memory reference)` failure on the Mac ARM build but nowhere else? From comparing the lists of test that do run, something is fishy with `window::clock::event_time_clock::tests::test_watermark` since that one is not in the CI output but is in my local test output. I have no idea what is malformed about it, though...

Psykopear:
> First start_at is renamed to align_to. That former name
incorrectly implied that there were no windows before that time, but
that is not true; windows extend (almost) infinitely in both
directions in time from that alignment point. *What determines if an
item is late for a window is not the windowing definition, but the
watermark of the clock*.

Yep, that makes perfect sense, looking at the previous implementation now I see the problem.

> Next, both {Sliding,Tumbling}Window are required to take
align_to. Previously, this was filled with the timestamp of the
start of the dataflow, but we did not persist it as the recovery
state, nor can we actually properly persist it because recovery state
is per-state-key but at that point we don't have any key. I didn't
want to work on figuring out a way to store "dataflow global" recovery
state, so to make it easier we require you to pick a fixed alignment
time. If you don't ensure align_to is consistent across executions,
you have window boundaries jumping around even when you are using
event time.

Sounds good, we'll probably want to add a line about this in the getting started guide, to make it clear that the `align_to` has to be consistent across executions.
We should clearly say that it's not ok to just align the window to the beginning of the hour using `now` as a reference if you want proper recovery.

> Then fixes up and adds additional tests for event time watermarking. I
don't belive we were properly advancing the watermark, preventing it
from going backwards, and handling the extreme no data and EOF cases.
Also does a fix for system time watermarking. Once the data has
completed, we should forever return the watermark as "end of time",
not just when there is an item being processed. I don't believe this
causes any difference in behavior with the way our operators are
currently written, but if we didn't process all available data on each
awakening, this is needed.

:+1:

> Then writes a ton more tests for sliding windowing. I believe we were
improperly mixing in the watermark time into the window
definitions. The definition of the window boundaries is a function of
the arguments to the config, not the current item time or
watermark. The watermark is just used to determine if an item is late.

Thanks for that, the tests made things clearer.

Overall looks good to me, thank you for the good work, this must not have been nice :smile: 
Only one note about the PR, unless the changes to the CI and pyo3 update are needed for this PR to work, I'd keep them separate so we can fix the issues arising from there without adding unrelated thing to this PR.

PS: If we bump rust version to 1.68, note that `Option::unzip` is stable since 1.66, so you can already use it

davidselassie:
> Only one note about the PR, unless the changes to the CI and pyo3 update are needed for this PR to work, I'd keep them separate so we can fix the issues arising from there without adding unrelated thing to this PR.

I don't want to include them but something about my changes to Rust tests was causing a segfault. Very mysterious. So I was trying to debug that here. I'm going to back out those changes and try to get this building by commenting out some of the tests. I do want to do a "bump all the versions" PR soon too, though.

src/window/sliding_window.rs
```diff
@@ -1,26 +1,38 @@
...
-///   offset (datetime.timedelta): Offset between windows.
+///   offset (datetime.timedelta): Duration between start times of
+///     adjacent windows.
 ///
-///   start_at (datetime.datetime): Instant of the first window. You
-///       can use this to align all windows to an hour,
-///       e.g. Defaults to system time of dataflow start.
+///   align_to (datetime.datetime): Align windows so this instant
+///     starts a window. This must be a constant. You can use this to
+///     align all windows to hour boundaries, e.g.
```
whoahbot, line 35: Missing example here

whoahbot (approved):
Thank you for writing an exhaustive set of tests for this!

PR #225, opened 2023-04-06
Adds --bytewax-log-level to pytest

davidselassie:
You can now run `pytest --bytewax-log-level=TRACE pytests/` to add the `setup_tracing` call for you, rather than having to shove that into a single test you want to investigate. If you want to show the captured output for a passed test (which is usually suppressed), also add `-rP`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@ repos:
     hooks:
     - id: black
   - repo: https://github.com/PyCQA/isort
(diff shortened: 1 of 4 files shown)


whoahbot (approved):
nice!

davidselassie:
Just as a note: I don't think this works with any `spawn_cluster` tests, nor will it work without modification with any of the changes in https://github.com/bytewax/bytewax/pull/222/

PR #228, opened 2023-04-10
Bugfix: adler32 accepts only a bytes-like object.

whoahbot:
Encodes the string to a bytes-like object before passing it to adler32. Adds a test for this case.


davidselassie (approved):
Whoops.

PR #231, opened 2023-04-17
Update docstrings, add upgrading guide

whoahbot:
- Added upgrading to v0.16 guide.
- Fixed docstring code snippets, and added them to `pytest` invocations.


docs/articles/reference/upgrading-to-v16.md
```diff
@@ -0,0 +1,258 @@
...
+
+### `PartitionedInput`
+
+`PartitionedInput` sources are input sources that manage a fixed number of independent partitions, like Kafka.
+
+`PartitionedInput` sources maintain the state of each source and re-build that state during recovery.
+
+## Capture -> Output
+
+Outputs in v0.16 are now stateful, and are now configured with a name for the step:
```
davidselassie, line 27:
```suggestion
The `capture` operator has been renamed to `output` in v0.16 and is now stateful, so requires a step ID:
```

docs/articles/reference/upgrading-to-v16.md
```diff
@@ -0,0 +1,258 @@
...
+In v0.16+:
+
+``` python
+from bytewax.dataflow import Dataflow
+from bytewax.connectors.stdio import StdOutput
+
+flow = Dataflow()
+flow.output("out", StdOutput())
+```
+
```
davidselassie, line 44:
```suggestion

`ManualOutputConfig` is now replaced by two base superclasses that can be subclassed to create custom output sinks. They are `DynamicOutput` and `PartitionedOutput`.

```

docs/articles/reference/upgrading-to-v16.md
```diff
@@ -0,0 +1,258 @@
...
+from bytewax.dataflow import Dataflow
+from bytewax.testing import TestingInput
+from bytewax.connectors.stdio import StdOutput
+
+flow = Dataflow()
+flow.input("in", TestingInput(range(3)))
+flow.output("out", StdOutput())
+```
+
+The Dataflow can be run with the following invocation:
```
davidselassie, line 61:
```suggestion
Since a Python file `dataflow.py` defines a module `dataflow`, the Dataflow can be run with the following invocation:
```

docs/articles/reference/upgrading-to-v16.md
```diff
@@ -0,0 +1,258 @@
...
+
+flow = Dataflow()
+flow.output("out", StdOutput())
+```
+
+## New entrypoint
+
+In Bytewax v0.16, the way that Dataflows are run has been simplified, and most execution functions have been removed.
+
+Similar to other Python frameworks like Flask and FastAPI, Dataflows can now be run using a Datfaflow import string in the <module_name>:<dataflow_variable_name_or_factory_function> format. As an example, given a file named `dataflow.py` with the following contents:
```
davidselassie, line 49: I forgot, does it default to looking for a `flow` variable? Can you leave off the `:flow`?

docs/articles/reference/upgrading-to-v16.md
```diff
@@ -0,0 +1,258 @@
...
+
+``` python
+from bytewax.dataflow import Dataflow
+from bytewax.connectors.stdio import StdOutput
+
+flow = Dataflow()
+flow.output("out", StdOutput())
+```
+
+## New entrypoint
```
davidselassie, line 45: Do we want to call out at the top level that Kafka recovery is being deprecated? Should we also add that to `CHANGELOG`?

docs/articles/reference/upgrading-to-v16.md
```diff
@@ -0,0 +1,258 @@
...
+
+Becomes:
+```python doctest:SKIP
+from bytewax.connectors.files import FileInput
+from bytewax.connectors.stdio import StdOutput
+```
+
+We removed `run_main` as it is now only used for unit testing dataflows. Bytewax now has a built-in FileInput connector, which uses the `PartitionedInput` connector superclass.
+
+Since we are using a built-in connector to read from this file, we can delete our `file_input` function above.
```
davidselassie, line 150: There is no `file_input` function above.

docs/articles/reference/upgrading-to-v16.md
```diff
@@ -0,0 +1,258 @@
...
+('of', 2)
+('opposing', 1)
+```
+
+We can even run our sample Dataflow on multiple workers to process the file in parallel:
+
+In one terminal, run:
+
+``` bash
+> python -m bytewax.run dataflow:flow -i0 -a "localhost:2101;localhost:2102"
```
davidselassie, line 251: Is this better demonstrated via the "spawn cluster" `-p` flag, rather than starting two workers by hand?

docs/articles/reference/upgrading-to-v16.md
```diff
@@ -0,0 +1,258 @@
...
+
+In v0.16, we have restructured input to be based around partitions in order to support rescaling stateful dataflows in the future. We have also dropped the `Config` suffix to our input classes. As an example, `KafkaInputConfig` has been renamed to `KafkaInput`.
+
+`ManualInputConfig` is now replaced by two base superclasses that can be subclassed to create custom input sources. They are `DynamicInput` and `PartitionedInput`.
+
+### `DynamicInput`
+
+`DynamicInput` sources are input sources that support reading distinct items from any number of workers concurrently.
+
+Dynamic inputs must be able to supply disjoint data for each worker that is started. If multiple workers read the same data, the Dataflow will process those as duplicate items.
```
davidselassie, line 15: This is true for all inputs, not just dynamic ones, and it something that is touched upon in the input docstrings. I'd just delete this line.

docs/articles/reference/upgrading-to-v16.md
```diff
@@ -0,0 +1,258 @@
...
+
+`DynamicInput` sources do not support storing resume state.
```
davidselassie, line 17:
```suggestion
`DynamicInput` sources do not support resume state and thus generally do not provide delivery guarantees better than at-most-once.
```

docs/articles/reference/upgrading-to-v16.md
```diff
@@ -0,0 +1,258 @@
...
+
+### `PartitionedInput`
+
+`PartitionedInput` sources are input sources that manage a fixed number of independent partitions, like Kafka.
+
+`PartitionedInput` sources maintain the state of each source and re-build that state during recovery.
```
davidselassie, line 23:
```suggestion
`PartitionedInput` sources can keep internal state on the current position of each partition and if they can rewind and resume reading from those previous on resume, can provide delivery guarantees of at-least-once or better.
```

docs/articles/reference/upgrading-to-v16.md
```diff
@@ -0,0 +1,258 @@
+## Upgrading to v0.16
+
+Bytewax v0.16 introduces major changes to the way that custom Bytewax inputs are created, as well as improvements to and windowing and execution. In this article, we'll cover some of the updates we've made to our API to make upgrading to v0.16 easier.
+
+## Input changes
```
davidselassie, line 5: Should we also call out that `bytewax.inputs.KafkaInputConfig` has been moved to `bytewax.connectors.kafka.KafkaInput`?

davidselassie (commented):
Thanks for doing all the boring stuff.

conftest.py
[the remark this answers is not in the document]
whoahbot: That was the issue, when running all the tests, the `--bytewax-log-level` stuff wasn't found until I moved this into the root of the project.

docs/articles/reference/upgrading-to-v16.md
whoahbot: When you invoke the dataflow with the `-p` flag, you don't see differentiated output for each worker, but perhaps that doesn't matter for this article.

docs/articles/reference/upgrading-to-v16.md
whoahbot: Yes, good catch!

docs/articles/reference/upgrading-to-v16.md
```diff
@@ -0,0 +1,264 @@
...
+
+### `DynamicInput`
+
+`DynamicInput` sources are input sources that support reading distinct items from any number of workers concurrently.
+
+`DynamicInput` sources do not support resume state and thus generally do not provide delivery guarantees better than at-most-once.
+
+### `PartitionedInput`
+
+`PartitionedInput` sources can keep internal state on the current position of each partition. If a partition can "rewind" and resume reading from a previous position, they can provide delivery guarantees of at-least-once or better.
```
Psykopear, line 19: by `or better`, do we mean exactly once with some caveats? Maybe we could be more specific here

PR #239, opened 2023-04-27
Fix readme

Psykopear:
Fixed the current readme, outdated links to the documentation and the example.
Reading the example, it looks quite complex as the first thing that users will see. What do you think of using a simpler example, and maybe link to a specific article/guide on the website for a longer/deeper introduction?


awmatheson:
Thanks for updating this @Psykopear. It might be interesting to simplify the readme, but the idea was to provide as briefly as possible all the components of a "real" dataflow instead of just explaining around a toy example, which is why it ended up a bit complex. I am open to simplifying with something that still provides room to explain the parts (input, output, stateful operators and stateless operators).

Psykopear:
Ok, I'd say let's keep it as it is for this PR, and if I come up with something "simpler" that still showcases an advanced use case, I'll open a new one

awmatheson (approved):

whoahbot (approved):

PR #254, opened 2023-06-09
Build exception messages only on error

davidselassie:
Adds `{re,}raise_with` (to parallel `{re,}raise`) and has it take a
closure that produces the error message. This causes the error
handling code to only attempt to format the exception message string
when an error has occurred. This was a somewhat significant part of
runtime for things like `extract_state_pair` and also probably
`stateful_map` and `reduce`.

My inspiration for the name comes from the Rust std lib functions like
`Option::get_or_insert_with`.

Converts call sites which were formatting a string to use the `_with`
versions. This looks like we'll get speedups in `filter`, `reduce`,
`stateful_map`, and all the other stateful operators via
`extract_state_pair` due to not superflously formatting errors on
every valid item.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -22,7 +22,7 @@ pub(crate) trait PythonException<T> {
     /// func().raise::<PyTypeError>("Raise TypeError adding this message")?;
     /// ```
     #[track_caller]
-    fn raise<PyErrType: PyTypeInfo>(self, msg: &str) -> PyResult<T>
+    fn raise<PyErrType: PyTypeInfo>(self, msg: &'static str) -> PyResult<T>
     where
         Self: Sized,
     {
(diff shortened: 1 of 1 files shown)


Psykopear:
Good catch, we were allocating strings everytime we defined an error, that was unnecessary. Going a bit further, what do you think of also forcing `raise` and `reraise` to only take a `&'static str` so that we don't accidentally use it when we need to allocate a string for the message? This PR already fixes all the current usages, but with the signature change we make it harder in the future to allocate at errors definition. Also, the `tracked_err` function could use the same separation, as I see we sometimes allocate there too.

An alternative could be to just change the original methods to only accept a closure. It makes using the trait with a simple `&str` a bit more verbose, but we keep error handling api simpler without adding methods

PR #263, opened 2023-06-27
Add _SQLSource and SQLInput to bytewax.connectors.sql

Carlos-Marques:
Hi everyone 👋 

In the context of this [Slack thread](https://bytewaxcommunity.slack.com/archives/C035LJ7URAQ/p1687863336846369) I opened this MR to share a new kind of connector I wrote for bytewax.

The jist of it is that it uses SQLAlchemy core operations to abstract access to the SQL compatible DB (you will also depend on the underlying engine, our case is Postgres).

It presents this interface:
```
class SQLInput(PartitionedInput):
    def __init__(
        self,
        connection_string: str,
        query: str,
        starting_cursors: dict,
        backoff_start_delay: float = 1.0,
        backoff_factor: float = 2.0,
        backoff_max_delay: float = 60.0,
    ):
```

1. connection_string: connection URI of DB
2. query: query to run that returns the results you want
3. starting_cursors: let's you define a multi column cursor that is used to keep the state/init the Input where you wanted it to start
4. backoff_*: let's you define the polling behaviour for when you get to the end of the results in your query (it uses SQLAlchemy batching and streaming of results while the query still has results but when you get to the real time data you need to poll the DB for extra results that are coming in)

simple example usage:
```
sql_input = SQLInput(
        connection_string=connection_uri,
        query="SELECT \"timestamp\", \"signal_1\" FROM \"table_1\" WHERE \"timestamp\" > :timestamp"
        starting_cursors={"timestamp": start_time},
    )
```

This returns a stream of events of this type:
`{"timestamp": datetime.datetime(*), "signal_1": <signal_value>}`

The query requirement is that it needs to select the cursors and the variables in the query need to have the same name of the selected cursors. (I still don't check this)
In the example I gave this happens because I select `"timestamp"` and in my WHERE I have a `:timestamp` variable.

If there is interest in this I can add tests and docs too!

### DIFF
--- a/pysrc/bytewax/connectors/sql.py
+++ b/pysrc/bytewax/connectors/sql.py
@@ -1,0 +1,108 @@
+"""Connectors for [SQL](https://en.wikipedia.org/wiki/SQL).
+
+Importing this module requires the
+[`SQLAlchemy`](https://github.com/sqlalchemy/sqlalchemy)
+package to be installed.
+
+It also requires any database-specific packages to be installed, such as
+[`psycopg2`](https://github.com/psycopg/psycopg2) for PostgreSQL.
+
+"""
+from time import sleep
+
+from bytewax.inputs import PartitionedInput, StatefulSource
+from sqlalchemy import create_engine, text
+
+__all__ = [
+    "SQLInput",
+]
+
+
+class _SQLSource(StatefulSource):
+    def __init__(
+        self,
+        connection,
+        query,
+        starting_cursors,
+        resume_state,
+        backoff_start_delay,
+        backoff_factor,
+        backoff_max_delay,
+    ):
+        self._cursor_dict = resume_state or starting_cursors
+        self._connection = connection
+        self._query = query
+        self._result = None
+        self._backoff_start_delay = backoff_start_delay
+        self._backoff_factor = backoff_factor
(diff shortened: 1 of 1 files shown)


pysrc/bytewax/connectors/sql.py
```diff
@@ -0,0 +1,108 @@
+"""Connectors for [SQL](https://en.wikipedia.org/wiki/SQL).
```
OlivierCavadenti, line 1: If this connector is ok to be in bytewax, I suggest to update docs:
https://github.com/bytewax/bytewax/blob/main/docs/articles/getting-started/ins_and_outs.md

pysrc/bytewax/connectors/sql.py
```diff
@@ -0,0 +1,108 @@
...
+        connection.execution_options(isolation_level="READ COMMITTED", readonly=True)
+        return _SQLSource(
+            connection,
+            self._query,
+            self._starting_cursors,
+            resume_state,
+            self._backoff_start_delay,
+            self._backoff_factor,
+            self._backoff_max_delay,
+        )
```
OlivierCavadenti, line 108: Need to add tests here: 
https://github.com/bytewax/bytewax/tree/main/pytests/connectors

PR #267, opened 2023-07-24
Properly handle when resume_state is falsy

davidselassie:
Turns out the code golf trick I was using in sources and sinks to save
the resume state if there is one or fallback to a default has a bug in
it.

```py
x = resume_state or 123
```

Does do the right thing and falls back to the default `123` when
`resume_state == None`, but also does so when `resume_state == False`
or `resume_state == 0` or `resume_state == []` or `resume_state ==
""`, which should be valid resume states.
    
Changes to an explicit ternary check for `None`.

Also fixes flake errors in files.


whoahbot (approved):
Good catch!

PR #271, opened 2023-08-01
SQLite-only recovery with dynamic partitions to support rescaling

davidselassie:
We did a lot of in-person discussion of this PR, but here's an overview for posterity:

Recovery now supports SQLite only. All the adapter traits and Kafka recovery code are torn out. You have to pre-create a fixed number of recovery partition SQLite DB files using `python -m bytewax.recovery` CLI, then specify a local directory to look for recovery partitions when using `python -m bytewax.run`. As long as some worker has access to every partition, the cluster will figure out how to route partition data correctly.

Recovery store schema changed and it now records more info and created-at timestamps.

Recovery snapshots now flow in the dataflow as plain Python objects and are serialized just before insertion into the DB with a replaceable `Serde` class. Defaults to using [`jsonpickle`](https://jsonpickle.github.io/) which has the universality of pickle, but is encoded in JSON an so sort of human-readable.

There is now a "partition assignment" set of Timely operators (`partition`, `assign_primary`, `route`, `partd_write`, `partd_load`) that let you do fixed-partition IO, but allow partition access to be non-uniform. This gives you a ton of operational freedom to write to local "partitions" whatever they may be, and Bytewax figures it all out. All workers broadcast in the 0th epoch what partitions they know about, then each worker deterministically generates a primary worker mapping for each partition, then later data is routed using that map. This is implemented in a stateless way in the `bytewax::timely` module, but is re-implemented in a stateful way in the `bytewax::{inputs, outputs}` modules.

This means that the production data in a dataflow now only starts flowing in epoch 1, so you will not see snapshots of epoch 0 anymore in the recovery store.

This also introduces batched IO for the entire dataflow and there are requisite `{Source,Sink}` API changes to support that.

### DIFF
--- a/pytests/test_execution.py
+++ b/pytests/test_execution.py
@@ -67,9 +67,9 @@ def test_cluster_can_be_ctrl_c():
             # With 2 workers per process
             "-w",
             "2",
-            # Set epoch interval to 0 so that the output
-            # is written to the file as soon as possible
-            "--epoch-interval",
+            # Set snapshot interval to 0 so that the output is written
+            # to the file as soon as possible
+            "-s",
             "0",
         ],
         # Use PIPE to check the content of stdout


CHANGELOG.md
```diff
@@ -5,6 +5,15 @@
 __Add any extra change notes here and we'll put them in the release
 notes on GitHub when we make a new release.__
 
+- *Breaking change* Recovery system worked. Kafka-based recovery
```
whoahbot, line 8:
```suggestion
- *Breaking change* Recovery system re-worked. Kafka-based recovery
```

pytests/test_execution.py
```diff
@@ -1,94 +1,94 @@
...
+    reason=(
+        "Sending os.kill(test_proc.pid, signal.CTRL_C_EVENT) sends event to all"
+        " processes on this console so interrupts pytest itself"
+    ),
+)
+def test_cluster_can_be_ctrl_c():
+    """Test that we can stop cluster execution by sending SIGINT (ctrl+c)."""
+    # Create a tmp file we can use to check the output
+    tmp_file = tempfile.NamedTemporaryFile()
+    print(tmp_file.name)
```
whoahbot, line 56: Do you want to keep this print?

whoahbot (commented):
There is one test that is failing for me locally in `inspect_epoch`, from the doctests in dataflow/mod.rs#L295, as epoch 0 is now reserved.

src/inputs.rs
```diff
@@ -3,39 +3,68 @@
...
 impl EpochInterval {
-    pub(crate) fn new(dur: Duration) -> Self {
-        Self(dur)
+    /// Determine how many epochs (rounded up) exist in some other
+    /// duration.
+    ///
+    /// Useful for calculating GC commit epochs.
+    pub(crate) fn epochs_per(&self, other: Duration) -> u64 {
+        (other.as_secs_f64() / self.0.as_secs_f64())
+            // Round up to we always have at least the backup interval
```
whoahbot, line 50: Round up *so* that we always have at least the backup interval?

Psykopear:
It fails here too when running `pytest`

src/inputs.rs
```diff
@@ -107,228 +135,291 @@ impl<'source> FromPyObject<'source> for PartitionedInput {
...
+        loads: &Stream<S, Snapshot>,
+    ) -> PyResult<(Stream<S, TdPyAny>, Stream<S, Snapshot>)>
     where
         S: Scope<Timestamp = u64>,
     {
-        let mut parts = self.build(py, step_id.clone(), index, count, resume_state)?;
+        let this_worker = scope.w_index();
 
-        let mut op_builder = OperatorBuilder::new(step_id.0.clone(), scope.clone());
+        let local_parts = self.list_parts(py)?;
```
Psykopear, line 184: I would add a `.reraise("...")` here, so that we get the line info where this failed. What I currently see is:
```
thread '<unnamed>' panicked at 'Box<dyn Any>', src/run.rs:117:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/docler/src/bytewax/bytewax/pysrc/bytewax/run.py", line 422, in <module>
    cli_main(**kwargs)
TypeError: (src/worker.rs:152:10) error building production dataflow
Caused by => TypeError: (src/worker.rs:292:30) error building PartitionedInput
Caused by => TypeError: 'set' object cannot be converted to 'Sequence'
```
And as you can see, the error only points us up to `worker.rs`, and we can't see where the error happened, while adding a `reraise` here gives us:
```
thread '<unnamed>' panicked at 'Box<dyn Any>', src/run.rs:117:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/docler/src/bytewax/bytewax/pysrc/bytewax/run.py", line 422, in <module>
    cli_main(**kwargs)
TypeError: (src/worker.rs:152:10) error building production dataflow
Caused by => TypeError: (src/worker.rs:292:30) error building PartitionedInput
Caused by => TypeError: (src/inputs.rs:187:47) error listing parts
Caused by => TypeError: 'set' object cannot be converted to 'Sequence'
```

src/inputs.rs
```diff
@@ -354,43 +445,39 @@ impl<'source> FromPyObject<'source> for StatefulSource {
...
     fn next_awake(&self, py: Python) -> PyResult<Option<DateTime<Utc>>> {
-        self.0
-            .call_method0(py, "next_awake")
-            .reraise("error calling `StatelessSource.next_awake`")?
-            .extract(py)
-            .reraise("error converting `StatefulSource.next_awake` return value to UTC datetime")
+        Ok(self
+            .0
+            .call_method0(py, intern!(py, "next_awake"))?
+            .extract(py)?)
```
Psykopear, line 462:
```rust
        self.0
            .call_method0(py, intern!(py, "next_awake"))?
            .extract(py)
```

src/timely.rs
```diff
@@ -66,7 +100,7 @@ where
 
     /// Is a given epoch closed based on this frontier?
     fn is_closed(&self, epoch: &T) -> bool {
-        self.simplify().iter().all(|frontier| *epoch < *frontier)
+        !self.simplify().is_some_and(|frontier| *epoch >= frontier)
```
Psykopear, line 103: This raises our minimum supported version to `1.70.0`, so we need to update CI and docker images. I think we should start keeping track of this in a `rust-toolchain.toml` file to make it explicit when we need to do that, but you'll need to do it in this PR too to make the CI pass

Psykopear (commented):
Ok, I read the PR and it looks ok. I love that we avoid most of the serialization work we were doing before, this will surely impact performances positively. It's hard to do a proper review due to the size, but things seem to work properly, and I trust you did a good job here :smile:

I'd ask you to do a pass of `clippy`, there are a number of easy fixes that improve the code at no cost.

There are 3 broken examples:
- anomaly_detector.py, periodic_input.py:stateful_flow
These fail because they return a set in `list_parts`, rather than a list. I think we should allow a set here (if PyO3 doesn't make this too hard). It made sense to restrict to a set when we needed that, but with the changes in this PR, both a list and a set make sense.

- periodic_input.py:stateless_flow
This fails because `next` needs to renamed to `next_batch`

src/inputs.rs
Psykopear: There are a number of fixes like this that you can spot with `clippy`, see main PR comment

pysrc/bytewax/recovery.py
```diff
@@ -1,15 +1,202 @@
...
+execution cluster synchronizes and snapshots its progress and
+state. You can adjust this duration via the `-e` parameter to
+`bytewax.run`. It defaults to every 10 seconds.
+
+The dataflow can only resume on epoch boundaries.
+
+In general, the longer this duration is, the less overhead there will
+be while the dataflow is executing due to recovery data, but the
+further back the dataflow might have to resume from.
+
```
awmatheson, line 139: Is it worth having a discussion about naming around Epochs here? I feel like epochs are largely opaque to the user and the most meaningful piece of information is around recovery.

recovery-interval
recovery-epoch

Psykopear (approved):
Looks good to me. CI is stuck at `test_cluster_can_be_ctrl_c`, we need to change `--epoch-interval` to `--snapshot-interval` in the test, or the dataflow will fail and the test will wait indefinitely that at least 10 lines are written to the output file.

pysrc/bytewax/recovery.py
awmatheson: Like it!

whoahbot (approved):
This is great work!

It's hard for me to give a thorough review of a PR of this size, but I feel that the approach you've taken here is solid.

PR #278, opened 2023-08-15
Uses ruff to lint

davidselassie:
- Uses ruff to lint
- Fixes all the ruff lints; mostly docstring formatting

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,7 @@ extras = [
 
 [project.optional-dependencies]
 kafka = [
-    "confluent-kafka>=2",
+    "confluent-kafka<=2.0.2",  # Update this to a broader bound once we fix CI.
 ]
 test = [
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -16,24 +16,37 @@ classifiers = [
...
+    "kafka",
+    "test",
+]
 
 [project.optional-dependencies]
-kafka = ["confluent-kafka<=2.0.2"]
-test = ["pytest==7.1.0", "myst-docutils==0.17.0"]
-docs = ["pdoc3==0.10.0"]
+kafka = [
+    "confluent-kafka>=2",
```
whoahbot, line 36: I think this is the source of the CI failures you're seeing here. Do you want to revert this change and we can bump this version in a subsequent PR?

whoahbot (approved):

PR #279, opened 2023-08-16
Update orderbook.py to use level2_batch

awmatheson:
Update orderbook example to use level2_batch. https://docs.cloud.coinbase.com/exchange/docs/websocket-channels#level2-batch-channel

Fix suggested in slack:
> Incase anyone is following the [Real-Time Financial Exchange Order Book](https://bytewax.io/guides/real-time-financial-exchange-order-book-application?utm_source=pocket_saves) guide it appears Coinbase made an update to their API
{'type': 'error', 'message': 'Failed to subscribe', 'reason': 'level2, level3, and full channels now require authentication. https://docs.cloud.coinbase.com/exchange/docs/websocket-auth'} if you update the channel to level2_batch everything still works. (edited)

### DIFF
--- a/examples/orderbook.py
+++ b/examples/orderbook.py
@@ -17,7 +17,7 @@ class CoinfbaseSource(StatefulSource):
                 {
                     "type": "subscribe",
                     "product_ids": [product_id],
-                    "channels": ["level2"],
+                    "channels": ["level2_batch"],
                 }
             )
         )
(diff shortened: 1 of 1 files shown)


examples/orderbook.py
```diff
@@ -17,7 +17,7 @@ def __init__(self, product_id):
                 {
                     "type": "subscribe",
                     "product_ids": [product_id],
-                    "channels": ["level2"],
+                    "channels": ["level2_batch"],
```
whoahbot, line 20: Since this is now returning a batch, you may want to consider updating your input adapter to use `next_batch` and return the list of messages you receive from the websocket.

whoahbot (approved):

PR #280, opened 2023-08-16
Removes now-unused pickling infra

davidselassie:
Since we don't use pickling to send over `Dataflow` objects and
instead each process imports from scratch the same Python file, we
don't need any of the pickling ceremony anymore.


whoahbot (approved):
I think you need to remove `mod common` from `lib.rs`, but other than that, this looks good to me.

whoahbot (approved):

PR #281, opened 2023-08-16
Calculate primaries only on worker 0

davidselassie:
Modifies `assign_primaries` to only run on worker 0: collect all
partition advertisements on worker 0, then run the operator there,
then broadcast back to everyone.

I think this will be less confusing to debug because only one worker
logs out the primary calculation, so we don't have to worry about
non-determinism in the algo.


whoahbot (approved):
LGTM

PR #294, opened 2023-08-28
Reference documentation on rescaling

davidselassie:
This is my deep dive on how rescaling works. I'd like to use this as a
blog post, but want to keep it for posterity in the reference
documentation.


whoahbot (approved):
This is great. My only thought is that it ends abruptly. Perhaps there is a short closing sentence or two that can book-end the article format we start with?

PR #304, opened 2023-10-04
Adds batching getter input helpers

davidselassie:
A few more batching helpers for when you have a client library that is
like what Igor was using via `redis-py`: `get_message()` that might
return a new message or `None` if no messages yet.

Adds a `batch_getter` which you can pass that function to
automatically create a batch iterator. Also adds `batch_getter_ex` if
the getter function raises an exception when there's no data yet.


whoahbot (approved):
:heavy_plus_sign:

PR #307, opened 2023-10-05
Rename IO classes

davidselassie:
Let me know if you think this API churn is ill-advised, but I think
the names we currently have surrounding IO are confusing.

Our input and output data model is centered around partitions, but
none of the classes are actually named anything like "partition". I
think this is confusing since we refer to partitions in comments,
docs, and when we talk about writing input sources. I think renaming
`*Source` to `*Partition` will help clarify this.

I also think its confusing that both the input objects and input
operator are named "input". In our communication and docs, we've
sometimes used "source" to indicate a type of location that data could
be read from. I think that is a better disambiguating word. I think
renaming `*Input` to `*Source` and `*Output` to `*Sink` will help
clarify this and make talking about these objects more natural because
they are nouns that only mean "location" and not "data" or "step".

I do not think we need to rename the `input` and `output` operators
once we do the above changes.

In summary, here's my suggested base class renames:

| Old | New |
|-----|-----|
| `Input` | `Source` |
| `PartitionedInput` | `FixedPartitionedSource` |
| `DynamicInput` | `DynamicSource` |
| `StatefulSource` | `StatefulSourcePartition` |
| `StatelessSource` | `StatelessSourcePartition` |
| `SimplePollingInput` | `SimplePollingSource` |
| `Output` | `Sink` |
| `PartitionedOutput` | `FixedPartitionedSink` |
| `DynamicOutput` | `DynamicSink` |
| `StatefulSink` | `StatefulSinkPartition` |
| `StatelessSink` | `StatelessSinkPartition` |

Thus that results in the following connector renames:

| Old | New |
|-----|-----|
| `FileInput` | `FileSource` |
| `FileOutput` | `FileSink` |
| `DirInput` | `DirSource` |
| `DirOutput` | `DirSink` |
| `CSVInput` | `CSVSource` |
| `StdOutput` | `StdOutSink` |
| `KafkaInput` | `KafkaSource` |
| `KafkaOutput` | `KafkaSink` |
| `TestingInput` | `TestingSource` |
| `TestingOutput` | `TestingSink` |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ notes on GitHub when we make a new release.__
 - *Breaking change* IO classes and connectors have been renamed to
   better reflect their semantics and match up with documentation.
 
+- Moves the ability to start multiple Python processes with the
+  `-p` or `--processes` to the testing namespace.
+
 ## v0.17.1
 
 - Adds the `batch` operator to Dataflows. Calling `Dataflow.batch`
--- a/docs/articles/getting-started/execution.md
+++ b/docs/articles/getting-started/execution.md
@@ -8,8 +8,8 @@ dataflow. Workers can be grouped into separate **processes**, but
 refer to the individual threads within.
 
 Bytewax's execution model uses identical workers. Workers execute all
-steps in a dataflow and automatically trade data to ensure the
-semantics of the operators. If a dataflow is run on multiple
+steps (including input and output) in a dataflow and automatically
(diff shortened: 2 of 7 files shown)


awmatheson:
This makes sense, although it is a lot of namespace to change. Better now than after there are several external connectors developed. 

I agree that it will make it a lot easier to talk about leveraging the new class names as they are more related to the actual mechanics.

whoahbot (approved):

PR #315, opened 2023-10-31
Deterministically awaken keys

davidselassie:
I found this issue as part of my other work.

Rust `HashMap`s do not guarantee a consistent iteration order. We're
storing the state for stateful operators in these maps and sometimes
to produce output, we iterate over the whole map. This means that
sometimes for an otherwise deterministic dataflow, you can see outputs
that change order.

So that we do get (more) deterministic outputs, converts all the state
`HashMap`s to `BTreeMap`s. This needs to happen in stateful unary, and
in the windowing logic (since it kept window state in a hash map too).

Yes, there are still other possible sources of non-determinism in the
multi-worker scenario, but this lets us get deterministic output in
tests in the 1-worker case. This means we don't need to add `sorted`
to all our unit tests. I'm doing that as part of overhauling all the
operator unit tests in my other branch, but wanted to bring up this
concept here.

### DIFF
--- a/src/operators/stateful_unary.rs
+++ b/src/operators/stateful_unary.rs
@@ -12,9 +12,8 @@
 //! store.
 
 use std::collections::hash_map::DefaultHasher;
+use std::collections::BTreeMap;
 use std::collections::BTreeSet;
-use std::collections::HashMap;
-use std::collections::HashSet;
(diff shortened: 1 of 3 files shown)


whoahbot (approved):
:heavy_plus_sign:

PR #317, opened 2023-11-01
Adds `TestingSource.{ABORT, EOF}`

davidselassie:
Previously, it was kind of annoying to test recovery semantics: You
had to construct a dataflow that threw an exception, but then somehow
stopped throwing that exception on the second pass. It's possible, but
it makes the tests feel very hack-y and there's a lot of very careful
setup to make that happen.

This change introduces two sentinel values you can add to the input
iterator of `TestingSource`.

`TestingSource.EOF()` causes the input to EOF, and sets up the state
so on the next resume, you don't have to do anything and the input
will continue from right after this sentinel. This is good for some
kinds of recovery state testing, but does cause the EOF to be
propagated through the dataflow and stateful operators might be
listening for it and change the state or emit things. So...

`TestingSource.ABORT()` raises a special new exception
`AbortExecution` to the input operator to set a flag telling the
worker "stop the main run loop". This causes the dataflow to
gracefully exit (we stop doing `worker.step()`), but does not require
any panics or exceptions. Since this doesn't propagate an EOF, all the
state snapshots of the operators are unchanged. This lets you test
resume from an abnormal situation. There is a trigger value inside of
the sentinel that is turned off so the resume does not trigger again;
this only works if the iterable is persistent, which is what will
happen in a standard unit test. See
`test_testing_source_{eof,abort}_run` for an example of how this
looks.

I haven't backported our existing tests to use this change, but I'm
starting to use it in my new Python operator tests and wanted to
introduce it separately.

Also updates the traceback messages to mention the step ID and
specific method names.

### DIFF
--- a/pysrc/bytewax/testing.py
+++ b/pysrc/bytewax/testing.py
@@ -2,7 +2,7 @@
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
 from itertools import islice
-from typing import Any, Iterable, Iterator, List, Optional
+from typing import Any, Iterable, Iterator, Optional
 
 from bytewax.inputs import (
     AbortExecution,


src/inputs.rs
```diff
@@ -630,59 +690,57 @@ impl DynamicSource {
...
+                                        .reraise_with(|| format!("error calling `StatelessSourcePartition.next_batch` in step {step_id}"))?
+                                );
+
+                                match res {
+                                    BatchResult::Batch(batch) => {
+                                        let batch_len = batch.len();
+
+                                        let mut downstream_session = downstream_handle.session(&part_state.output_cap);
+                                        for item in batch {
+                                            downstream_session.give(item.into());
```
whoahbot, line 714: Is there a reason that we want to iterate through the batch here, instead of using `give_vec` as we were?

I don't have a benchmark to point to directly, but my recollection was that when we were doing performance work, we found that `give_vec` had some benefits for inputs.

whoahbot (approved):

PR #321, opened 2023-11-07
Non-linear dataflows and Python operators

davidselassie:
We've discussed in person a good amount of this work, but a quick summary of highlights.

Operators are now created by defining a builder function and decorating with `@operator()`. The decorator turns the builder function into a dataclass containing the parameters for the operator and ~a builder method that will be monkey patched / loaded onto the relevant classes wrapping the builder code with the ceremony to build the dataclass and add it to the `Dataflow`~ the function that comes out of the other end of the operator wraps up scoping and instantiating that dataclass for you. You call the function directly.

`Stream` handles are now returned and contain a "scope" to ensure that this nesting occurs properly. Operators can return multiple streams and take multiple streams. Tons of new operators.

Adds in `Generic` metaclass and `TypeVar` to have proper typing for all operators and `Stream[X]`. This means if you turn on type checking and annotate all your logic functions, you can find errors. `Source` and `Sink` are also now generic so Bytewax can test that you are passing the correct shape items from / to them.

~`KeyedStream` subclass which contains operators that required keyed streams to work. This should help with finding these bugs early.~ `KeyedStream` is now an alias to the generic type `String[Tuple[str, X]]` so MyPy will show errors for any operator that doesn't use the right shape.

~The operators must be "loaded" as to prevent name clashing. Built-in operators are loaded automatically.~ Operators are just functions. Import them to use them. A nice convention is to do `import bytewax.operators as op` and then you can use `op.` to see all the operators.

The "functional form" of calling the operators as functions is the lowest level because it provides the easiest way to implement type checking and the full flexibility of varargs, etc. But we also have a chaining convenience method, `Stream.then` which you can pass the operator function as the first argument and it'll do argument munging to give you a fake fluent interface. It only works with simple shapes of operators, though. You'll have to fall back to the functional form for more complicated named joins, etc.

Calling the operator methods mutates and results in a fully constructed `Dataflow` in terms of "core operators" or the lowest-level operators that are implemented in Rust / Timely. The `build_production_dataflow` function now does a depth-first-walk through all the operators and whenever we encounter a core one, builds the appropriate Timely operator. This means there is zero overhead to nesting operators.

The `fold_window` operator is the root of all windowing operators, but I'd like to eventually rewrite it to be built on top of `unary` so we can allow Python extensions to clocks and windowers.

Main things you want to look at:

- `pysrc/bytewax/dataflow.py` Contains the dataflow data model definition and `@operator()` decorator. The module docstring here explains how to define custom operators.
- `pysrc/bytewax/operators/__init__.py` Contains all built-in operator definitions.
- `pysrc/bytewax/operators/window.py` Contains all built-in window operator definitions.
- `src/worker.rs` contains the new Timely dataflow building process.
- `src/dataflow.rs` contains the Rust shims over the Python dataflow data model.
- `src/operators/mod.rs` contains the core operator definitions.
- `pytests/operators/` contains one file per operator to test.
- `pysrc/bytewax/_encoder.py` contains the JSON data model.

Other stuff:

- Some tests consolidated: event time tests into window operator tests, next awake tests into input tests.
- I ended up writing most of the instructional documentation in the `bytewax` module docstring and this ended up duplicating what was in `docs/articles/getting-started`, so I ended up moving all that content into module docstrings. We don't have to permanently live with this decision, but I do think using the pdoc generated docs more heavily in the future is the way to go because we get automatic linking and can easily reference specific methods and things. In any case, when we're doing a pass over docs and guides for this release, we can bring back the rm'd stuff here and copy and paste move everything around.
- Metrics retrofit onto new operators.
- Fixing a few examples.
- `inspect` now counts as an "output" operator. You can have an `inspect` instead of an output and the dataflow will run. It no longer passes through the values since the non-fluent interface means that isn't very helpful.

I'd like to have this PR reviewed for structure and the ideas in the code, then try to do some deep dives / pairing / boot camps on what this new work enables. Then work separately and collaboratively on updating examples and then docs once everyone feels more comfortable with the new patterns we have.

### DIFF
--- a/pysrc/bytewax/operators/__init__.py
+++ b/pysrc/bytewax/operators/__init__.py
@@ -60,7 +60,7 @@ Using a defined function:
 ...     return sentence.split()
 >>> s = op.input("inp", flow, TestingSource(["hello world"]))
 >>> s = op.flat_map("split", s, split_sentence)
->>> op.inspect("out", s)
+>>> _ = op.inspect("out", s)
 >>> run_main(flow)
 use_def.out: 'hello'
 use_def.out: 'world'
@@ -70,7 +70,7 @@ Or a lambda:
 >>> flow = Dataflow("use_lambda")
 >>> s = op.input("inp", flow, TestingSource(["hello world"]))
 >>> s = op.flat_map("split", s, lambda s: s.split())
->>> op.inspect("out", s)
+>>> _ = op.inspect("out", s)
 >>> run_main(flow)
 use_lambda.out: 'hello'
 use_lambda.out: 'world'
@@ -80,7 +80,7 @@ Or an unbound method:
 >>> flow = Dataflow("use_method")
 >>> s = op.input("inp", flow, TestingSource(["hello world"]))
 >>> s = op.flat_map("split", s, str.split)
->>> op.inspect("out", s)
+>>> _ = op.inspect("out", s)
 >>> run_main(flow)
 use_method.out: 'hello'
 use_method.out: 'world'
@@ -173,8 +173,8 @@ def branch(
     >>> flow = Dataflow("branch_eg")
     >>> nums = op.input("nums", flow, TestingSource([1, 2, 3, 4, 5]))
     >>> evens, odds = op.branch("even_odd", nums, lambda x: x % 2 == 0)
-    >>> op.inspect("evens", evens)
-    >>> op.inspect("odds", odds)
+    >>> _ = op.inspect("evens", evens)
+    >>> _ = op.inspect("odds", odds)
     >>> run_main(flow)
     branch_eg.odds: 1
     branch_eg.evens: 2
@@ -229,7 +229,7 @@ def flat_map(
     >>> def split_into_words(sentence):
     ...     return sentence.split()
     >>> s = op.flat_map("split_words", s, split_into_words)
-    >>> op.inspect("out", s)
+    >>> _ = op.inspect("out", s)
     >>> run_main(flow)
     flat_map_eg.out: 'hello'
     flat_map_eg.out: 'world'
@@ -295,7 +295,7 @@ def inspect_debug(
     >>> from bytewax.dataflow import Dataflow
     >>> flow = Dataflow("inspect_debug_eg")
     >>> s = op.input("inp", flow, TestingSource(range(3)))
-    >>> op.inspect_debug("help", s)
+    >>> _ = op.inspect_debug("help", s)
     >>> run_main(flow)
     inspect_debug_eg.help W0 @1: 0
     inspect_debug_eg.help W0 @1: 1
@@ -750,7 +750,7 @@ def filter(  # noqa: A001
     >>> def is_odd(item):
     ...     return item % 2 != 0
     >>> s = op.filter("filter_odd", s, is_odd)
-    >>> op.inspect("out", s)
+    >>> _ = op.inspect("out", s)
     >>> run_main(flow)
     filter_eg.out: 1
     filter_eg.out: 3
@@ -847,7 +847,7 @@ def filter_map(
     ...     else:
     ...         return data["key"], data
     >>> s = op.filter_map("validate", s, validate)
-    >>> op.inspect("out", s)
+    >>> _ = op.inspect("out", s)
     >>> run_main(flow)
     filter_map_eg.out: ('a', {'key': 'a', 'val': 1})
 
@@ -955,7 +955,7 @@ def inspect(
     >>> from bytewax.dataflow import Dataflow
     >>> flow = Dataflow("my_flow")
     >>> s = op.input("inp", flow, TestingSource(range(3)))
-    >>> op.inspect("help", s)
+    >>> _ = op.inspect("help", s)
     >>> run_main(flow)
     my_flow.help: 0
     my_flow.help: 1
@@ -1245,7 +1245,7 @@ def map(  # noqa: A001
     >>> def add_one(item):
     ...     return item + 10
     >>> s = op.map("add_one", s, add_one)
-    >>> op.inspect("out", s)
+    >>> _ = op.inspect("out", s)
     >>> run_main(flow)
     map_eg.out: 10
     map_eg.out: 11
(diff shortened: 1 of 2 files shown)


ariskk:
YES!!! 🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥

examples/join.py
```diff
@@ -0,0 +1,19 @@
...
+    length=timedelta(seconds=2), align_to=datetime(2023, 1, 1, tzinfo=timezone.utc)
+)
+
+
+flow = Dataflow("join")
+inp1 = flow.input("inp1", TestingSource(["a"])).key_on("k1", lambda x: "KEY")
+inp2 = flow.input("inp2", TestingSource(["b"])).key_on("k2", lambda x: "KEY")
+inp3 = flow.input("inp3", TestingSource(["c"])).key_on("k3", lambda x: "KEY")
+
+inp1.join_inner("j1", inp2).join_inner("j2", inp3).output("out", StdOutSink())
```
ariskk, line 19: I am so loving this

examples/apriori.py
```diff
@@ -1,91 +1,31 @@
...
 
 
-product_counter: Counter = Counter()
-most_popular_products: dict[str, int] = {}
-
-cc = SystemClockConfig()
-wc = TumblingWindow(
-    length=timedelta(seconds=5), align_to=datetime(2023, 1, 1, tzinfo=timezone.utc)
+pairs = (
+    baskets.flat_map("pairs", lambda basket: itertools.combinations(basket, 2))
```
whoahbot, line 26:
```suggestion
    baskets.flat_map("pairs", lambda basket: list(itertools.combinations(basket, 2)))
```

examples/hn_scraper.py
```diff
@@ -1,45 +1,62 @@
...
+
+def mapper(old_max_id, new_max_id):
+    if old_max_id is None:
+        # Get the last 10 items on the first run.
+        old_max_id = new_max_id - 10
+    return (new_max_id, range(old_max_id, new_max_id))
+
+
+ids = (
+    max_id.stateful_flat_map("range", mapper)
```
whoahbot, line 53: Was this intended to include a custom operator example?

pysrc/bytewax/__init__.py
```diff
@@ -1,15 +1,339 @@
...
+If you've ever used Python's
+[`map`](https://docs.python.org/3/library/functions.html#map) or
+[`filter`](https://docs.python.org/3/library/functions.html#filter) or
+`functool`'s
+[`reduce`](https://docs.python.org/3/library/functools.html#functools.reduce)
+or equivalent in other languages, operators are the same concept. If
+not, no worries, there's an example provided in the documentation for
+each operator in `bytewax.operators`.
+
+The each operator method takes a **step ID** you should set to a
```
Psykopear, line 165:
```suggestion
Each operator method takes a **step name** you should set to a
```

pysrc/bytewax/__init__.py
```diff
@@ -1,15 +1,339 @@
...
+unique name that represents the purpose of this computational step.
+This gets attached to errors, logging, and metrics so you can link
+these bits of information for debugging.
+
+Here's an example that uses a few more operators.
+
+TODO Word count again?
```
Psykopear, line 172: I think word count is ok in this phase of the docs

pysrc/bytewax/__init__.py
```diff
@@ -1,15 +1,339 @@
...
+## Debugging
+
+### Inspecting
+
+Bytewax provides the `bytewax.operators.inspect` operator to help you
+visualize what is going on inside your dataflow. It prints out the
+repr of all items that pass through it when you run the dataflow. You
+can attach it on the end of any stream you'd like to see.
+
+TODO Example?
```
Psykopear, line 210: Maybe link to operator's doc if it already has one

pysrc/bytewax/__init__.py
```diff
@@ -1,15 +1,339 @@
...
+
+### Step IDs
```
Psykopear, line 212: Should we change all reference of `Step IDs` to `Step Names`?

pysrc/bytewax/tracing.py
```diff
@@ -1,33 +1,134 @@
...
+      - jaeger
+```
+
+Now run `docker compose up` and everything should be up and running.
+
+Now we need the dataflow. Download the example in this folder:
+
+```shell
+curl https://raw.githubusercontent.com/bytewax/bytewax/main/examples/wikistream.py \
+  -o dataflow.py
```
Psykopear, line 93: We should probably fix this link to a specific release

pysrc/bytewax/tracing.py
```diff
@@ -1,33 +1,134 @@
...
+
+```shell
+python3 -m venv .venv
+source .venv/bin/activate # Or activate.fish on fish shell
+pip install bytewax sseclient-py urllib3
+```
+
+Now you can run it with:
+```shell
+python dataflow.py
```
Psykopear, line 127:
```suggestion
python -m bytewax.run dataflow
```

Psykopear (commented):
This is great work, thank you :D 

I added some doc related comments, but the PR looks good, the new api unlocks a lot of new possibilities for Bytewax.

I've opened a separate PR where I fix all the examples, so I could get used to the new fluent api, and it all seems to work great.

We need to work on the docs too. I think it might make sense to treat this branch as a feature branch, and open the needed PRs using this as a base. Once everything is completed, we merge everything in main. What do you think?

pysrc/bytewax/dataflow.py
```diff
@@ -1,12 +1,880 @@
...
+                if isinstance(val, _HasScope):
+                    vals[fld.name] = val._with_scope(outer_scope)
+            out = dataclasses.replace(out, **vals)
+
+        return out
+
+    return method
+
+
+def operator(_detect_parens=None, *, _core: bool = False) -> Callable:
```
whoahbot, line 761: Not terribly important- We chatted about this, but here is a version that handles the use of the decorator with and without parens. I can't remember if you had a case in mind where this doesn't work, so I'm adding it here for posterity.

```python
def operator(fn=None, *, _core: bool = False) -> Callable:
    """Function decorator to define a new operator.

    See `bytewax.dataflow` module docstring for how to use this.

    """

    def inner_deco(builder: FunctionType) -> Type:
        sig = inspect.signature(builder)
        sig_types = typing.get_type_hints(builder)
        cls = _gen_op_cls(builder, sig, sig_types, _core)
        op_method = _gen_op_method(sig, sig_types, builder, cls, _core)
        setattr(cls, op_method.__name__, op_method)
        # Note that although this is a function definition, a class will
        # be bound to the function name.
        return cls

    if callable(fn):
        return inner_deco(fn)
    else:
        return inner_deco
```

whoahbot (approved):

Psykopear (approved):

PR #333, opened 2023-12-07
Performance work

davidselassie:
- Updates PyO3 to 0.20
- Use a persistent output buffer in flat_map
- Use GILPool to GC memory earlier
- Turn logic functions into refs before item loops
- Make `map` a core operator
- Comment on swapping
- `map` uses outbuff too; renames for consistency
- Don't convert next_awake into Python if the wakeup was immediate
- Fixes encoder tests

### DIFF
--- a/pysrc/bytewax/operators/__init__.py
+++ b/pysrc/bytewax/operators/__init__.py
@@ -1151,11 +1151,12 @@ def map_value(
 
     """
 
-    def shim_mapper(v: V) -> List[W]:
+    def shim_mapper(k_v: Tuple[str, V]) -> Tuple[str, W]:
+        k, v = k_v
         w = mapper(v)
-        return [w]
+        return (k, w)
 
-    return flat_map_value("flat_map_value", up, shim_mapper)
+    return map("map", up, shim_mapper)
 
 
 @overload


PROFILING.md
```diff
@@ -13,47 +13,68 @@ Install `perf` on Ubuntu with:
...
+* `-C force-frame-pointers` helps call stack recreation.
+
 ## 2. Profile Execution
 
 We can now use [`perf`](https://www.brendangregg.com/perf.html) to
 profile an execution of a dataflow:
 
 ```bash
-perf record -F 999 --sample-cpu --call-graph dwarf --aio -- python example_dataflow.py
+perf record -F max --sample-cpu --call-graph=dwarf --aio -- python -X perf -m bytewax example_dataflow
```
Psykopear, line 72:
```suggestion
perf record -F max --sample-cpu --call-graph=dwarf --aio -- python -X perf -m bytewax.run example_dataflow
```

Psykopear (commented):
Looks good to me.

I'm not sure about the manual `pool` handling though. Simply removing all the:

```rust
let pool = unsafe { py.new_pool() };
let py = pool.python();
```

Improves performances on my machine with the pathological test case we started with, and analyzing the `perf` output, I still don't see the `gc` calls dominating the cpu time like previsouly, so maybe this is not needed? Or did you see cases where doing this actually improved performances/perf output analysis?

Psykopear (approved):

Psykopear:
Just another datapoint to prove this is useful, I ran my usual `kafka-to-kafka` throughput benchmark for my `schema_registry_pyoperators` branch both with and without this PR merged in, and without it the dataflow struggles to handle even 5k msgs/s, while with this PR merged its back to ~100k msgs/s that we had in 0.17

whoahbot (approved):

PR #335, opened 2023-12-09
Changes snapshot and backup interval to saner defaults

davidselassie:
Previously the defaults for this were backing up every 10 seconds and
keeping around snapshots for a day. This means your recovery DB would
max out at a size theoretically thousands of times bigger than your
in-memory state. This is "correct" but surprising.

Let's change the defaults to have only a 2x size increase. This is
probably a "somewhat resonable" default value for production but
should definitely be tuned for each deployment's backup configuration.

### DIFF
--- a/pysrc/bytewax/run.py
+++ b/pysrc/bytewax/run.py
@@ -230,9 +230,9 @@ def _create_arg_parser():
         "-s",
         "--snapshot-interval",
         type=_parse_timedelta,
-        default=timedelta(minutes=30),
         help="""System time duration in seconds to snapshot state for recovery;
-        defaults to 30 minutes""",
+        on resume, dataflow might need to rewind and replay all the data processed
+        in one of these intervals""",
         action=_EnvDefault,
         envvar="BYTEWAX_SNAPSHOT_INTERVAL",
     )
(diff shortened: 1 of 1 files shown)


pysrc/bytewax/run.py
```diff
@@ -230,20 +230,20 @@ def _create_arg_parser():
         "-s",
         "--snapshot-interval",
         type=_parse_timedelta,
-        default=timedelta(seconds=10),
+        default=timedelta(minutes=30),
```
whoahbot, line 233: I'm not sure what the correct value here should be.

I was thinking of jobs that don't have much state but use Kafka as an input source. Since Kafka offsets are also stored, 30 minutes can potentially be a large backlog of Kafka data to re-process.

I agree that 10 seconds is too short, but is 30 minutes possibly too long?

whoahbot: I completely agree with your point that there are no "sane" defaults for every use case. Bearing that in mind, asking users to specify those defaults when configuring recovery seems like an attractive option. Should we consider asking users to specify those as non-optional parameters when configuring recovery?

whoahbot (approved):

PR #336, opened 2023-12-12
Update getting started

whoahbot:


docs/articles/getting-started/overview.md
```diff
@@ -1,32 +1,38 @@
...
+- Recovery
+- Windowing
+- Joins
+- And more!
 
-- The Bytewax runtime provides a recovery system which automatically
-  backs up the state in your dataflow and allows you to recover from
-  failure without slowly re-processing all data.
+Bytewax also offers a [Platform](https://bytewax.io/platform) that simplifies
+the management of stateful workloads on Kubernetes, and adds many additional features.
```
Psykopear, line 26: Maybe cite `aws` and `gcp` support too here?

docs/articles/getting-started/batch-example.md
```diff
@@ -1,68 +1,171 @@
...
 import bytewax.operators as op
+import bytewax.operators.window as win
+from bytewax.dataflow import Dataflow
+from bytewax.connectors.stdio import StdOutSink
 from bytewax.testing import TestingSource
-import bytewax.testing
+
+flow = Dataflow("batching")
+s = op.input("input", flow, TestingSource(list(range(10))))
+# The following line creates a keyed stream where all values have the key "ALL"
```
Psykopear, line 40: I'd comment on why we are doing this rather than what we are doing, something like:
```python
# Here we want to batch all the items together, so we use the same fixed key for all the items
```

docs/articles/getting-started/batch-example.md
```diff
@@ -1,68 +1,171 @@
...
+keyed_stream = op.key_on("key", s, lambda _x: "ALL")
+batched_stream = op.batch("batch", keyed_stream, timedelta(seconds=10), 3)
```
Psykopear, line 42: Being one of the first things our users will read about bytewax, I'd specify the name of the arguments explicitly here:
```python
batched_stream = op.batch("batch", keyed_stream, timeout=timedelta(seconds=10), batch_size=3)

File: docs/articles/getting-started/batch-example.md [hunk 4]
```diff
@@ -1,68 +1,171 @@
...
+inp = [
+    {"time": align_to, "user": "a", "val": 1},
+    {"time": align_to + timedelta(seconds=4), "user": "a", "val": 1},
+    {"time": align_to + timedelta(seconds=5), "user": "b", "val": 1},
+    {"time": align_to + timedelta(seconds=8), "user": "a", "val": 1},
+    {"time": align_to + timedelta(seconds=12), "user": "a", "val": 1},
+    {"time": align_to + timedelta(seconds=13), "user": "a", "val": 1},
+    {"time": align_to + timedelta(seconds=14), "user": "b", "val": 1},
+]
+s = op.input("input", flow, TestingSource(inp))
```
Psykopear, line 88: Same as above, `stream` or `inp` instead of `s`

Psykopear (commented):
Looks good to me, I added some nitpicky comments, but the structure looks fine.

Since we are doing a lot of work to make typing work, I think we should showcase this too. We could either add type hints to all the examples, or add a chapter about it. I'm not sure which way would be better though. Always using type hints might be too much info all at once in the first examples, but only having a specific chapter of the guide talk about it doesn't seem optimal either.

cra:
> 
> We could either add type hints to all the examples, or add a chapter about it. [...] 

We can also consider a toggle that turns them on or off for first examples 🤔

docs/articles/getting-started/batch-example.md
```diff
@@ -1,68 +1,171 @@
...
 import bytewax.operators as op
+import bytewax.operators.window as win
+from bytewax.dataflow import Dataflow
+from bytewax.connectors.stdio import StdOutSink
 from bytewax.testing import TestingSource
-import bytewax.testing
+
+flow = Dataflow("batching")
+stream = op.input("input", flow, TestingSource(list(range(10))))
+# Here we want to batch all the items together, so we use the same fixed key for all the items
```
Psykopear, line 38: Rereading this, I must have changed the phrasing a couple of times and ended up with the "all the items" repetition. Maybe better:
```
# We want to batch all the items together, so we assign the same fixed key to each of them
```

docs/articles/getting-started/simple-example.md
```diff
@@ -1,85 +1,117 @@
-# Getting Started
+## A Simple Example
 
-## Baby Steps
+Let's write our first Bytewax Dataflow. Be sure that you've followed the instructions
+for installing bytewax in [Installing bytewax](/docs/articles/getting-started/installation.md).
```
konradsienkowski, line 4: This link format is correct when navigating through our GH repository, but it'd result in an error on our website (where the path is: `/docs/getting-started/installation/`) if we merge it now. 

If we want to keep it that way in the source files, I need to check if there's an option to override link hrefs while parsing markdown in the website app

konradsienkowski: FYI: I prepared a branch in our website repo with a fix that replaces the "internal GH repo" link format (e.g. /docs/articles/[category]/[filename].md) to the format we're using in the app (e.g. /docs/[category]/[filename]) during the markdown parsing step - so keeping that format in the source shouldn't result in any errors on front-end. 

Let me know if we want to use it, then I'll deploy the branch on staging for testing

docs/articles/getting-started/overview.md
```diff
@@ -1,32 +1,40 @@
-# Overview
+Bytewax is a Python framework for building Dataflow programs over bounded or unbounded data.
```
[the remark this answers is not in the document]
whoahbot: Makes sense to me. I'll change it.

PR #338, opened 2023-12-12
Handle `Optional` and more complex types in operator signatures

davidselassie:
Fede found that the `@operator` decorator barfed on having an argument
that was `x: Optional[Dict[str, str]]`. This was because we weren't
properly delineating between what was an "annotation" (which is the
thing after the `x: ` and could be anything) and a "type" (like a
Python type). Remember this unfortunate fact:

```
In [54]: type(Stream)
Out[54]: type

In [55]: type(Stream[int])
Out[55]: typing._GenericAlias
```

Python plops whatever value is the annotation. It might be an actual
`type`, but it also might be the value that represents a generic type
or a "generic alias". The latter is not actually part of the class
hierarchy, so we have to handle it via special casing:

```
In [56]: issubclass(Stream, Stream)
Out[56]: True

In [57]: issubclass(Stream[int], Stream)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[57], line 1
----> 1 issubclass(Stream[int], Stream)

TypeError: issubclass() arg 1 must be a class
```

Instead of trying to "normalize" the annotations into types right
away, instead I keep all the annotations as `Any` and add a
`_anno_to_typ` function which tries to determine the type and handles
the special case of the generic aliases.

Then adds the extra branches when we're building up the operator
dataclass definition to handle when that function can't determine the
type. Handling that last case is what actually fixes the bug Fede
found.

In general it is more rigorously correct to build up the operator
dataclass definition using annotations, not the "origin types" since
those lose the generics information.

Two other tiny things:

- Uses `TypeAlias` for `KeyedStream` which is recommended
- Removes unused `_rec_subclasses`

### DIFF
--- a/pysrc/bytewax/dataflow.py
+++ b/pysrc/bytewax/dataflow.py
@@ -19,6 +19,7 @@ from typing import (
     Generic,
     Iterable,
     List,
+    Optional,
     Protocol,
     Type,
     TypeVar,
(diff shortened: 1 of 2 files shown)


Psykopear (approved):
Looks good to me! As soon as this is merged, I'll rebase the kafka connector PR on top of this

PR #341, opened 2023-12-14
`flat_map_batch` operator

davidselassie:
After hearing about the use cases for being able to access the Timely
runtime batches from Bytewax operators, I thought I'd prototype up a
little `flat_map_batch` operator.

It takes a list of the current batch of items on this activation, and
let's you return a list of as many items to emit downstream as you
want.

I then defined `flat_map` in terms of `flat_map_batch`. Easy, works.

Then I was curious if defining `map` in terms of `flat_map_batch`
would let us have some of the performance benefit without the code
duplication of having another core operator. So I benchmarked our
pathological dataflow again.

```python
import bytewax.operators as op
from bytewax.outputs import DynamicSink, StatelessSinkPartition
from typing import List, Any, TypeVar


class NullSinkPartition(StatelessSinkPartition[Any]):
    def write_batch(self, items: List[Any]) -> None:
        return None


class NullSink(DynamicSink):
    def build(self, worker_index: int, worker_count: int) -> NullSinkPartition:
        return NullSinkPartition()

X = TypeVar("X")

def identity(x: X) -> X:
    return x

def true(x: Any) -> bool:
    return True

flow = Dataflow("patho_eg")
inp = op.input("input", flow, TestingSource([list(range(1_000_000))]))

inp.then(
    op.flat_map, "1", identity
).then(op.map, "2", identity).then(op.map, "3", identity).then(
    op.map, "4", identity
).then(op.map, "5", identity).then(op.output, "output", NullSink())
```

Turns out! It's even faster to have `map` in terms of
`flat_map_batch`. I guess you get some big performance wins when you
can write stuff out to buffers in a single go.

```
(.venv312) dsel@lemur ~/b/bytewax (flat-map-batch-map) [0|SIGINT]> hyperfine --parameter-list branch_name main,flat-map-batch,flat-map-batch-map \
                                                                       --setup "git co {branch_name}; maturin develop --release" \
                                                                       "python -m bytewax.run examples.patho"
Benchmark 1: python -m bytewax.run examples.patho (branch_name = main)
  Time (mean ± σ):     750.4 ms ±  63.9 ms    [User: 705.9 ms, System: 42.6 ms]
  Range (min … max):   682.1 ms … 894.4 ms    10 runs

Benchmark 2: python -m bytewax.run examples.patho (branch_name = flat-map-batch)
  Time (mean ± σ):     742.1 ms ±  59.9 ms    [User: 699.8 ms, System: 41.6 ms]
  Range (min … max):   687.8 ms … 883.8 ms    10 runs

Benchmark 3: python -m bytewax.run examples.patho (branch_name = flat-map-batch-map)
  Time (mean ± σ):     511.3 ms ±  40.4 ms    [User: 476.0 ms, System: 34.4 ms]
  Range (min … max):   480.7 ms … 623.1 ms    10 runs

Summary
  python -m bytewax.run examples.patho (branch_name = flat-map-batch-map) ran
    1.45 ± 0.16 times faster than python -m bytewax.run examples.patho (branch_name = flat-map-batch)
    1.47 ± 0.17 times faster than python -m bytewax.run examples.patho (branch_name = main)
```

### DIFF
--- a/pytests/operators/test_flat_map_batch.py
+++ b/pytests/operators/test_flat_map_batch.py
@@ -1,0 +1,22 @@
+from typing import Iterable, List
+
+import bytewax.operators as op
+from bytewax.dataflow import Dataflow
+from bytewax.testing import TestingSink, TestingSource, run_main
+
+
+def test_flat_map_batch():
+    inp = ["split this", "and this"]
+    out = []
+
+    def split_into_words(sentences: List[str]) -> Iterable[str]:
+        for sentence in sentences:
+            yield from sentence.split()
+
+    flow = Dataflow("test_df")
+    s = op.input("inp", flow, TestingSource(inp, batch_size=10))
+    s = op.flat_map_batch("split_into_words", s, split_into_words)
+    op.output("out", s, TestingSink(out))
+
+    run_main(flow)
+    assert out == ["split", "this", "and", "this"]


Psykopear (approved):
Nice catch! It seems that batching whenever we can is the way to go.
Looks good to me

whoahbot (approved):
:rocket:

PR #344, opened 2023-12-15
Joins concepts documentation and `join_window` fixes

davidselassie:
- Fixes bugs in windowing operators: return types were missing `WindowMetadata`, `join_window` event clock needs to be shimmed
- First pass on join concepts docs
- Adds `product: bool` arg to `join_window`
- Finishes joins doc

### DIFF
--- a/.cbfmt.toml
+++ b/.cbfmt.toml
@@ -1,3 +1,3 @@
 [languages]
 rust = ["rustfmt"]
-python = ["black --fast -"]
+python = ["ruff format -"]
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -23,8 +23,6 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
       - uses: actions/setup-python@v3
-      - name: Install black for cbfmt
-        run: pip install .[fmt]
       - uses: mkroening/rust-toolchain-toml@main
       - name: Install cbfmt
         run: |
@@ -32,6 +30,8 @@ jobs:
             curl -fsSL -o cbfmt.tar.gz "https://github.com/lukas-reineke/cbfmt/releases/download/v0.2.0/cbfmt_linux-x86_64_v0.2.0.tar.gz"
             tar --strip-components 1 -xvf cbfmt.tar.gz
             mv cbfmt /usr/local/bin/
+      - name: Install cbfmt sub-formatting commands
+        run: pip install .[cbfmt]
       - uses: pre-commit/action@v3.0.0
 
   linux:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,24 +7,24 @@ repos:
     - id: mixed-line-ending
       args: ["--fix=lf"]
     - id: trailing-whitespace
-  - repo: https://github.com/psf/black
-    rev: 22.3.0
-    hooks:
-    - id: black
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.0.284
+    rev: v0.1.8
     hooks:
       - id: ruff
         args: ["--fix", "--exit-non-zero-on-fix"]
+      - id: ruff-format
   - repo: https://github.com/doublify/pre-commit-rust
     rev: v1.0
     hooks:
       - id: clippy
+        name: "cargo clippy"
       - id: fmt
+        name: "cargo fmt"
   - repo: https://github.com/bytewax/pre-commit-cbfmt
     rev: v0.1
     hooks:
       - id: fmt
+        name: cbfmt
   - repo: https://github.com/netromdk/vermin
     rev: v1.5.2
     hooks:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,9 @@ notes on GitHub when we make a new release.__
   `fold_final` and `reduce_final`. They now only emit on EOF and are
   only for use in batch contexts.
 
+- *Breaking change* `batch` operator renamed to `collect`, so as to
+  not be confused with runtime batching. Behavior is unchanged.
+
 - *Breaking change* `output` operator does not forward downstream its
   items. Add operators on the upstream handle instead.
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -725,8 +725,9 @@ checksum = "b4668fb0ea861c1df094127ac5f1da3409a82116a4ba74fca2e58ef927159bb3"
 
 [[package]]
 name = "libsqlite3-sys"
-version = "0.26.0"
-source = "git+https://github.com/rusqlite/rusqlite.git?rev=7b0393210be4bce17156a08fde732083d6342a75#7b0393210be4bce17156a08fde732083d6342a75"
+version = "0.27.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cf4e226dcd58b4be396f7bd3c20da8fdee2911400705297ba7d2d7cc2c30f716"
 dependencies = [
  "cc",
  "pkg-config",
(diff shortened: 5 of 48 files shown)


docs/articles/concepts/joins.md
```diff
@@ -1,3 +1,774 @@
-# Joins
+## Joins
 
-Streaming joins are not the same as your typical RDBMS joins!
+A **join** is a way to combine items in two or more streams into a
+single stream, matching them by a **join key**.
+
+This is a very similar concept to the joins available in SQL, but
+there are some new edge cases and behaviors you have to be aware of
+because streaming data is possible infinite.
```
whoahbot, line 8:
```suggestion
when handling unbounded streams of data.
```

docs/articles/concepts/joins.md
```diff
@@ -1,3 +1,774 @@
...
+[`join_window`](/apidocs/bytewax.operators/window/index#bytewax.operators.window.join_window),
+and
+[`join_window_named`](/apidocs/bytewax.operators/window/index#bytewax.operators.window.join_window)
+operators to provide this functionality.
+
+### Join Keys
+
+All the operators above are stateful operators and so require all of
+the upstreams to contain 2-tuples with the first element being a
+string. All items that have the same key will be joined together.
```
whoahbot, line 66:
```suggestion
string. The first element of the tuple is the key. All items that have
the same key will be joined together.
```

We mention key here but didn't define it previously.

docs/articles/concepts/joins.md
```diff
@@ -1,3 +1,774 @@
...
+
+Make this key unique enough so you don't bring together too much.
```
whoahbot, line 68: This is good advice, but perhaps it belongs at the end of the document in something like important considerations? In the example we're working through at this point, we don't give the reader a choice of key space.

docs/articles/concepts/joins.md
```diff
@@ -1,3 +1,774 @@
...
+
+This key must be a string, so if you have a different data type,
```
whoahbot, line 70: Perhaps this paragraph can be used above where I tried to inject a definition.

docs/articles/concepts/joins.md
```diff
@@ -1,3 +1,774 @@
...
+in the dict, we'll map the value to just include the relevant field.
+
+```python
+keyed_names = op.map("key_names", names, lambda x: (str(x["user_id"]), x["name"]))
+keyed_emails = op.map("key_emails", emails, lambda x: (str(x["user_id"]), x["email"]))
+```
+
+Let's
+[`inspect`](/apidocs/bytewax.operators/index#bytewax.operators.inspect)
+our streams to double check we know what they look like:
```
whoahbot, line 90: I struggled with the inclusion of `run_main` in the examples I've wrote, but I also like the idea that these docs are all tested.

I wonder if it would be possible to add a DOCTEST annotation that adds the `run_main` invocation to the end of the snippet so that it doesn't have to be included.

docs/articles/concepts/joins.md
```diff
@@ -1,3 +1,774 @@
...
+memory _forever_ if you keep adding keys. This might be the behavior
+you want, but realize that it does not come for free. A non-running
+streaming join is better if you know you'll only get one value for
+each side for each key, since it knows it can discard the state after
+sending the values downstream.
+
+## Windowed Joins
+
+The streaming join assumes that a key could come anywhere in the
+entire lifetime of a stream. This means it could possible wait forever
```
whoahbot, line 428:
```suggestion
entire lifetime of a stream. This means it could possibly wait forever
```

docs/articles/concepts/joins.md
```diff
@@ -1,3 +1,774 @@
...
+for a value to come on a side that will never come. Another option is
```
whoahbot, line 429:
```suggestion
for a value that will never arrive. Another option is
```

whoahbot (approved):
Looks good! I loved the diagrams. 

Left you some suggestions you can take as you see fit.

docs/articles/concepts/joins.md
```diff
@@ -1,3 +1,775 @@
...
+[`join`](/apidocs/bytewax.operators/index#bytewax.operators.join), has
+[`join_named`](/apidocs/bytewax.operators/index#bytewax.operators.join_named)
+and
+[`join_window`](/apidocs/bytewax.operators/window/index#bytewax.operators.window.join_window),
+has
+[`join_window_named`](/apidocs/bytewax.operators/window/index#bytewax.operators.window.join_window).
+The named versions have identical parameters and join semantics, but
+in stead of emitting `tuple`s down stream, they emit `dict`s in which
+the keys are the names of the keyword arguments you use to specify the
+upstream sides. This helps out a lot when you many-sided joins and it
```
cra, line 699: (couldn't help but notice that this is similar to the infamous meme ["I accidently X. The whole thing"](https://knowyourmeme.com/memes/i-accidentally).
Probably you want to change the last sentence of this paragraph

docs/articles/concepts/joins.md
```diff
@@ -1,3 +1,775 @@
...
+            "at": datetime(2023, 12, 14, 0, 30, tzinfo=timezone.utc),
+            "email": "queen@bytewax.io",
+        },
+    ]
+)
+
+bytewax.testing.run_main(flow)
+```
+
+Notice how now we only have the latest email for the bee:
```
cra, line 651: to clarify, this example is great and fits the same user-id reasons, and let's keep it. But everywhere else we probably want to change that

docs/articles/concepts/joins.md
```diff
@@ -1,3 +1,775 @@
...
+
+Bytewax provides the operators
+[`join_window`](/apidocs/bytewax.operators/window/index#bytewax.operators.window.join_window),
+and
+[`join_window_named`](/apidocs/bytewax.operators/window/index#bytewax.operators.window.join_window)
+to implement this.
+
+For the details of all the types of windows you can define and
+explanation of the parameters, see our [windowing
+documentation](windowing.md). I'm going to use a simple 1 hour
```
cra, line 444: It's better to stick to the same-person style for consistency and in this sort of docs it's better to use "**We** are going to use" instead of "**I'm** going to use", since it's technically not a blog post

docs/articles/concepts/joins.md
```diff
@@ -1,3 +1,775 @@
...
+('123', ('Bee', 'queen@bytewax.io'))
+```
+
+So the running join is cool in that you can track updates to changes
+in values over time. _But this comes with a downside!_ Because we
+never throw away the state for a key, this state keeps growing in
+memory _forever_ if you keep adding keys. This might be the behavior
+you want, but realize that it does not come for free. A non-running
+streaming join is better if you know you'll only get one value for
+each side for each key, since it knows it can discard the state after
```
cra, line 423: "it knows it can discard" sounds just a bit too black-magical-ish, let's change it to something like "it is set up to discard"

docs/articles/concepts/joins.md
```diff
@@ -1,3 +1,775 @@
...
+```
+
+First the table starts empty.
+
+| Key | Name Value | Email Value |
+| --- | ---------- | ----------- |
+
+Then we see the first name value. Since the table was updated in any
+way, it emits the current values for that key and keeps them in the
+table! If it doesn't know a value yet, it fills it in as `None`.
```
cra, line 360: Let's include right here a short suggestion on what the reader can do with such `None` values, even if we think it's obvious that it should be a separate step downstream, we should explicitly encourage the user to do it this way.

cra (changes requested):
Great job! Noticed a couple of typos and leaved a few suggestions

docs/articles/concepts/joins.md
```diff
@@ -1,3 +1,775 @@
...
+    {"user_id": 12345, "name": "Bee"},
+    {"user_id": 12345, "name": "Hive"},
+]
+names = op.input("names", flow, TestingSource(names_l))
+
+emails_l = [
+    {"user_id": 12345, "email": "bee@bytewax.io"},
+    {"user_id": 12345, "email": "hive@bytewax.io"},
+]
+emails = op.input("emails", flow, TestingSource(emails_l))
```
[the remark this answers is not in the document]
cra: hmm good point. But then let's link the getting-started article here somewhere or above this? Like "if you just want to run some joins, check out "getting started / joins" or similar

PR #349, opened 2023-12-19
Update container guide

Psykopear:
Update the chapter about containers in the guide.

As suggested by @cra the article starts by building a Docker container to run a bytewax dataflow from scratch, then present a more complex scenario with `docker compose` to run a redpanda cluster and the dataflow.
Finally, it introduces the existing bytewax images, mostly following the previous article here with some fixes.

@cra @miccioest the docker compose setup seems to be working on my machine, but please try it to be sure it works.


miccioest (approved):
Looks good to me.

The docker-compose example works from the point of view that it starts the redpanda cluster and runs the dataflow. 

Then, the Python script crashes because the Bytewax version installed in the container image is the current one (`0.17.2`), and the code is for `0.18.x`.

I left some suggestions.

docs/articles/deployment/container.md
```diff
@@ -1,25 +1,164 @@
...
+You'll need an image with python support, and can just install bytewax and run the dataflow.
+
+Create a `Dockerfile` with the following content:
+```Dockerfile
+# Start from a debian slim with python support
+FROM python:3.11-slim-bullseye
+# Setup a workdir where we can put our dataflow
+WORKDIR /bytewax
+# Install bytewax and the dependencies you need here
+RUN pip install bytewax
```
cra, line 40: Pin the version plz

docs/articles/deployment/container.md
```diff
@@ -1,25 +1,164 @@
...
+# Copy the dataflow in the workdir
+COPY dataflow.py dataflow.py
+# And run it
+# Note: use `-u` too to let python run unbuffered,
```
cra, line 44: Alternatively, you can use full cli arg and then we can skip this note

docs/articles/deployment/container.md
```diff
@@ -1,25 +1,164 @@
...
+from bytewax.connectors.kafka import operators as kop
+from bytewax.connectors.stdio import StdOutSink
+from bytewax.dataflow import Dataflow
+
+flow = Dataflow("test")
+inp = kop.input("in", flow, brokers=["redpanda:9092"], topics=["in_topic"])
+op.output("out", inp, StdOutSink())
+```
+
+Now we can create a docker-compose file that runs both a `Redpanda` instance and our dataflow.
```
cra, line 86: Let's add a note here that we are basically using the same standard redpanda docker compose almost exactly

docs/articles/deployment/container.md
Psykopear: yep, my bad, fixed

docs/articles/deployment/container.md
Psykopear: I'm going to explicitely set the env var, so the setting doesn't depend on the parameters passed to `python`. What to you mean with "use the full cli args"?

docs/articles/deployment/container.md
```diff
@@ -1,25 +1,164 @@
...
+```bash
+docker run --rm bytewax-custom
+```
+
+## Example: docker compose, kafka connector and redpanda
+
+Modify the Dockerfile to install optional `kafka` dependencies in bytewax:
+```diff
+- RUN pip install bytewax
++ RUN pip install bytewax[kafka]
```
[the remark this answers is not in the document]
Psykopear: oh right, I left it like this because I was using a `0.17` dataflow to test things out, pinning it now.

docs/articles/deployment/container.md
Psykopear: ok, should I also link redpanda's docs here, or just mention them?

docs/articles/deployment/container.md
```diff
@@ -106,125 +306,48 @@ We are going to use the `BYTEWAX_KEEP_CONTAINER_ALIVE` environment variable to k
 
 ```bash
 docker run --rm --name=my-dataflow \
-    -v /var/bytewax/examples:/bytewax/examples \
-    -e BYTEWAX_PYTHON_FILE_PATH=examples.pagerank:flow \
+    -v ./dataflows:/bytewax/dataflows \
```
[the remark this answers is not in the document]
Psykopear: Now that you mention it, I remember having to use absolute paths in the past, but I copy/pasted all the commands from the guide several times before pushing, so it did work. Maybe this is fixed with buildkit? I can  put `$(pwd)` rather than `./` there just to be sure.

docs/articles/deployment/container.md
```diff
@@ -106,125 +306,48 @@ We are going to use the `BYTEWAX_KEEP_CONTAINER_ALIVE` environment variable to k
...
+0
+1
+2
+3
+4
 ```
-
-In summary, to create your own docker image, you will need to use the Bytewax images available in Docker Hub as your base image and then install the needed modules in the virtual environment that the image already has.
-
-## Bytewax Container Images and Security
```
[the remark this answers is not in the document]
Psykopear: Putting it back

docs/articles/deployment/container.md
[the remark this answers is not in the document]
miccioest: Yeah, they now allow us to use relative paths in the new version. I agree with you. It sounds good to put `$(pwd)`.

docs/articles/deployment/container.md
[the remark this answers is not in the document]
miccioest: thanks

docs/articles/deployment/container.md
cra: Sorry, my bad, I was 100% sure that there's a full version of that flag like `--unbuffered-stdin-and-stdiout`
+1 for env var

PR #352, opened 2023-12-19
Properly handle resume from EOF

davidselassie:
We need to snapshot a partition on EOF since it will have no more
state changes and won't be around when / if the epoch interval
actually occurs later.

Fixes the continuation test to actually test non-epoch aligned
continuation.


whoahbot (approved):
nice catch!

PR #365, opened 2024-01-09
Fixed count_window so it passes the whole item to the clock

### LINKED ISSUE #364: `count_window` only sends the count to the event clock's `dt_getter` function
### Bug Description

The `bytewax.operators.window.count_window` operator, used with an `EventClockConfig` only sends the count to the `dt_getter` function rather than the whole item from where the time can be extracted. This makes it impossible to use `EventClockConfig` with `count_window`.

Currently `count_window` is implemented like this:
```python
@operator
def count_window(
    step_id: str,
    up: Stream[X],
    clock: ClockConfig,
    windower: WindowConfig,
    key: Callable[[X], str],
) -> KeyedStream[Tuple[WindowMetadata, int]]:
    init_count: KeyedStream[int] = op.map("init_count", up, lambda x: (key(x), 1))
    return reduce_window("sum", init_count, clock, windower, lambda s, x: s + x)
```

So we first `map` the stream to a keyed stream of counts, and pass that stream to `reduce_window`, which in turn only passes the count to the reducer.

We can fix this problem by using `op.key_on("key", up, key)` instead of `map`, and implement `count_window` in terms of `fold_window` rather than `reduce_window`:

```python
@operator
def count_window(
    step_id: str,
    up: Stream[X],
    clock: ClockConfig,
    windower: WindowConfig,
    key: Callable[[X], str],
) -> KeyedStream[Tuple[WindowMetadata, int]]:

    def _shim_builder() -> int:
        return 0

    def _shim_folder(count: int, _item: X) -> int:
        return count + 1

    keyed: KeyedStream[int] = op.key_on("extract_key", up, key)
    return fold_window("sum", keyed, clock, windower, _shim_builder, _shim_folder)
```

Python 3.11.6

bytewax==0.18.0

### Operating System version (`uname -morp`)

6.6.10-arch1-1 x86_64 unknown GNU/Linux

### Relevant log output

_No response_

### Steps to Reproduce

Try this dataflow:

```python
import json
from datetime import datetime, timedelta, timezone
from operator import itemgetter

from bytewax import operators as op
from bytewax.operators import window as win

flow = Dataflow("state_counting")

data = [
    '{"time":1672531218000,"state":"running"}',
    '{"time":1672531204000,"state":"running"}',
    '{"time":1672531220000,"state":"aborted"}',
]
stream = op.input("input", flow, TestingSource(data))
stream = op.map("parse", stream, json.loads)


def extract_ts(m):
    return datetime.fromtimestamp(m["time"] / 1000, tz=timezone.utc)


clock = win.EventClockConfig(extract_ts, wait_for_system_duration=timedelta(seconds=0))
windower = win.TumblingWindow(length=timedelta(minutes=1), align_to=align_to)
counted_stream = win.count_window("count", stream, clock, windower, itemgetter("state"))
op.inspect("output", counted_stream)
```

`extract_ts` should receive the full item, instead it only receives the counter.

Psykopear:
Fixes #364
See the issue for details.

### DIFF
--- a/pysrc/bytewax/operators/window.py
+++ b/pysrc/bytewax/operators/window.py
@@ -194,8 +194,15 @@ def count_window(
         A stream of `(key, count)` per window at the end of each window.
 
     """
-    init_count: KeyedStream[int] = op.map("init_count", up, lambda x: (key(x), 1))
-    return reduce_window("sum", init_count, clock, windower, lambda s, x: s + x)
+
+    def _shim_builder() -> int:
+        return 0
+
+    def _shim_folder(count: int, _item: X) -> int:
+        return count + 1
+
+    keyed: KeyedStream[int] = op.key_on("extract_key", up, key)
+    return fold_window("sum", keyed, clock, windower, _shim_builder, _shim_folder)
 
 
 @operator(_core=True)
--- a/pytests/operators/window/test_count_window.py
+++ b/pytests/operators/window/test_count_window.py
@@ -1,0 +1,40 @@
+from datetime import datetime, timedelta, timezone
+
+import bytewax.operators as op
+import bytewax.operators.window as win
+from bytewax.dataflow import Dataflow
+from bytewax.operators.window import EventClockConfig, TumblingWindow, WindowMetadata
+from bytewax.testing import TestingSink, TestingSource, run_main
+
+ZERO_TD = timedelta(seconds=0)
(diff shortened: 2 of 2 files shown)


cra (approved):

whoahbot (approved):
LGTM, do you want to add a test that shows the bug?

Psykopear:
Added a test that fails without the fix, thanks for the heads up @whoahbot . Merging this

PR #366, opened 2024-01-09
Add Kafka concepts section

whoahbot:
This PR adds a Kafka concepts section to our documentation, but also fixes some issues I found in our API docs when I was writing the guide.

### DIFF
--- a/docs/articles/concepts/kafka.md
+++ b/docs/articles/concepts/kafka.md
@@ -74,7 +74,14 @@ extra information fields like `.headers`.
 
 Messages that are published to a `KafkaSink` must be of type
 [`KafkaSinkMessage`](/apidocs/bytewax.connectors/kafka/message#bytewax.connectors.kafka.message.KafkaSinkMessage).
-With the `.key` and `.value` fields set, or of type `KafkaSourceMessage`, if the message is unmodified.
+
+You can create a `KafkaSinkMessage` with the data you want:
+```python
+msg = KafkaSinkMessage(key=None, value="some_value")
+```
+And you can optionally set `topic`, `headers`, `partition` and `timestamp`.
+
+The `output` operator also accepts messages of type [`KafkaSourceMessage`](/apidocs/bytewax.connectors/kafka/message#bytewax.connectors.kafka.message.KafkaSourceMessage). They are automatically converted to `KafkaSinkMessage` keeping only `.key` and `.value` from `KafkaSourceMessage`.
 
 ## Kafka and Recovery
 
--- a/docs/articles/concepts/kafka.md
+++ b/docs/articles/concepts/kafka.md
@@ -1,6 +1,10 @@
 Bytewax ships with connectors for [Kafka](https://www.confluent.io/) and
 Kafka compatible systems, like [Redpanda](https://redpanda.com/).
 
+In this section, we'll discuss these connectors in detail, as well as
+provide some important operational information about using them as sources
+and sinks.
+
 ## Connecting to Kafka
 
 Bytewax provides two basic ways to connect to Kafka.
@@ -39,8 +43,8 @@ kop.output("kafka-out", processed, brokers=brokers, topic="out-topic")
 By default, Bytewax will consume a single message at a time from Kafka. The default
 setting is often sufficient for lower latency, but negatively affects throughput.
 
-If your workload would from higher throughput, you can set the `batch_size` parameter
-to a higher value (eg: 1000). The `batch_size` parameter sets the maximum number of messages
+If your dataflow would benefit from higher throughput, you can set the `batch_size` parameter
+to a higher value (eg: 1000). The `batch_size` parameter configures the maximum number of messages
 that will be fetched at a time from each worker.
 
 ## Message Types
@@ -51,7 +55,7 @@ This dataclass includes basic Kafka fields like `.key` and `.value`, as well as
 extra information fields like `.headers`.
 
 Messages that are published to a `KafkaSink` must be of type [`KafkaSinkMessage`](
-https://bytewax.io/apidocs/bytewax.connectors/kafka/message#bytewax.connectors.kafka.message.KafkaSinkMessage).
+/apidocs/bytewax.connectors/kafka/message#bytewax.connectors.kafka.message.KafkaSinkMessage).
 With the `.key` and `.value` fields set.
 
 ## Kafka and Recovery
@@ -146,15 +150,10 @@ will need to be restarted in order to fetch new versions of a schema.
 
 ## Error handling
 
-Kafka input sources in Bytewax return a dataclass that contains two output streams of
-[`KafkaMessage`s](/apidocs/bytewax.connectors/kafka/message#bytewax.connectors.kafka.message.KafkaSourceMessage).
-
-The `.oks` field contains a stream of messages that were successfully processed. The `.errs` field
-contains a stream of messages where an error was encountered. Items that encountered an error
-have their `.error` field set with more details about the error.
-
-In the previous section, we used the [`raises`](/apidocs/bytewax.operators/index#bytewax.operators.raises) operator
-to crash the dataflow when encountering messages in the `.errs` stream.
+Kafka input sources in Bytewax return a dataclass that contains two output streams.
+The `.oks` field contains a stream of [`KafkaMessage`](/apidocs/bytewax.connectors/kafka/message#bytewax.connectors.kafka.message.KafkaSourceMessage)
+that were successfully processed. The `.errs` field contains a stream of [`KafkaError`](/apidocs/bytewax.connectors/kafka/error#bytewax.connectors.kafka.message.KafkaError)
+where an error was encountered. Items that encountered an error have their `.err` field set with more details about the error.
 
 It is important to note that if no processing is attached to the `.errs` stream of messages, they will be silently
 dropped, and processing will continue.
--- a/docs/articles/concepts/kafka.md
+++ b/docs/articles/concepts/kafka.md
@@ -1,0 +1,133 @@
+Bytewax ships with connectors for [Kafka](https://www.confluent.io/) and
+Kafka compatible systems, like [Redpanda](https://redpanda.com/).
+
+## Connecting to Kafka
+
+Bytewax provides two basic ways to connect to Kafka.
+
+You can use [`KafkaSource`](/apidocs/bytewax.connectors/kafka#bytewax.connectors.kafka.KafkaSource)
+and [`KafkaSink`](/apidocs/bytewax.connectors/kafka#bytewax.connectors.kafka.KafkaSink) directly:
+
+```python
+from bytewax.connectors.kafka import KafkaSource, KafkaSink
+from bytewax import operators as op
+from bytewax.dataflow import Dataflow
+
+brokers = ["localhost:1909"]
+flow = Dataflow("example")
+kinp = op.input("kafka-in", flow, KafkaSource(["localhost:1909"], ["in-topic"]))
+processed = op.map("map", kinp, lambda x: (x.key, x.value))
+op.output("kafka-out", processed, KafkaSink(brokers, "out-topic"))
+```
+
+Or use the custom Kafka operator:
+
+```python
+from bytewax.connectors.kafka import operators as kop
+from bytewax import operators as op
+from bytewax.dataflow import Dataflow
+
+brokers = ["localhost:1909"]
+flow = Dataflow("example")
+kinp = kop.input("kafka-in", flow, brokers=brokers, topics=["in-topic"])
+op.inspect("errors", kinp.errs).then(op.raises, "crash-on-err")
+processed = op.map("map", kinp.oks, lambda x: (x.key, x.value))
+kop.output("kafka-out", processed, brokers=brokers, topic="out-topic")
+```
+
+## Error handling
+
+In the previous section, you may have noted that we added a chain of operators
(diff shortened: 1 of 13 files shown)


pysrc/bytewax/connectors/kafka/operators.py
```diff
@@ -91,27 +91,26 @@ def input(  # noqa A001
...
-            Whether to wait for new data on this topic when the
+        step_id: Unique name for this step
+
+        flow: A Dataflow
+
+        brokers: List of `host:port` strings of Kafka brokers.
+
+        topics: List of topics to consume from.
+
+        tail: Whether to wait for new data on this topic when the
```
Psykopear, line 102: Is this a different formatter config, or just a style choice? I like it, but if it's an automated thing I might need to update something in my local environment

Psykopear (commented):
Good work, thanks for this. I left some comments but it looks good.

One thing you mentioned in the call that's missing is a chapter about customization, like the ability to create your own schema registry, or (de)serializer classes, which is basically hidden right now as you have to look at the code to see what you can do.

docs/articles/concepts/kafka.md
```diff
@@ -0,0 +1,167 @@
...
+flow = Dataflow("example")
+kinp = kop.input("kafka-in", flow, brokers=brokers, topics=["in-topic"])
+processed = op.map("map", kinp.oks, lambda x: KafkaSinkMessage(x.key, x.value))
+kop.output("kafka-out", processed, brokers=brokers, topic="out-topic")
+```
+
+## Error handling
+
+The [`KafkaSource`](/apidocs/bytewax.connectors/kafka/index#bytewax.connectors.kafka.KafkaSource) input
+source will raise an exception whenever an error is encountered when consuming from Kafka.
```
cra, line 44: This subsection lacks some sample code examples on handling exceptions in the user's code. It could be very high-level pseudo-code but a starting point is nice to have

docs/articles/concepts/kafka.md
```diff
@@ -0,0 +1,167 @@
...
+returns a dataclass containing two output streams. The `.oks` field is a stream of
+[`KafkaMessage`](/apidocs/bytewax.connectors/kafka/message#bytewax.connectors.kafka.message.KafkaSourceMessage)
+that were successfully processed. The `.errs` field is a stream of [`KafkaError`](/apidocs/bytewax.connectors/kafka/error#bytewax.connectors.kafka.message.KafkaError)
+messages where an error was encountered. Items that encountered an error have their `.err` field set with more
+details about the error.
+
+Note that if no processing is attached to the `.errs` stream of messages, they will be silently
+dropped, and processing will continue.
+
+Alternatively, error messages in the `.errs` stream can be published to a "dead letter queue" where they
```
cra, line 56: short clarification of what we mean by that (separate Kafka topic for bad messages) or, ideally, a very short code snippet how one does that would be amazing to have here

docs/articles/concepts/kafka.md
```diff
@@ -0,0 +1,167 @@
...
+can be inspected and reprocessed later, while allowing the dataflow to continue processing data.
+
+## Batch sizes
+
+By default, Bytewax will consume a single message at a time from Kafka. The default
+setting is often sufficient for lower latency, but negatively affects throughput.
```
cra, line 62: Any numbers that we can share here? _"Up to 10k messages per second can be processed by a single worker in default mode, blablabla"_
If we have this numbers of course and if they are not too infrastructure-dependent

docs/articles/concepts/kafka.md
```diff
@@ -0,0 +1,167 @@
...
+## Message Types
+
+Messages received from `KafkaSource` are of type [`KafkaSourceMessage`](
+/apidocs/bytewax.connectors/kafka/message#bytewax.connectors.kafka.message.KafkaSourceMessage).
+This dataclass includes basic Kafka fields like `.key` and `.value`, as well as
+extra information fields like `.headers`.
+
+Messages that are published to a `KafkaSink` must be of type
+[`KafkaSinkMessage`](/apidocs/bytewax.connectors/kafka/message#bytewax.connectors.kafka.message.KafkaSinkMessage).
+With the `.key` and `.value` fields set, or of type `KafkaSourceMessage`, if the message is unmodified.
```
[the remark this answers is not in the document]
cra: what happens if the topic is not specified and there were multiple topics in the in-op?

docs/articles/concepts/kafka.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Kafka and Recovery
+
+Typical deployments of Kafka utilize [consumer groups](https://developer.confluent.io/courses/architecture/consumer-group-protocol/)
+in order to manage partition assignment and the storing of Kafka offsets.
+
+Bytewax does **not** use consumer groups to store offsets or asssign partitions to consumers. In order
+to correctly support [recovery](/docs/concepts/recovery), Bytewax must manage and store the
+consumer offsets in recovery partitions.
```
cra, line 86: semantic collision detected. We need to be careful to specify when we talk about Kafka partitions here and bytewax partitions

docs/articles/concepts/kafka.md
```diff
@@ -0,0 +1,167 @@
...
+extra information fields like `.headers`.
+
+Messages that are published to a `KafkaSink` must be of type
+[`KafkaSinkMessage`](/apidocs/bytewax.connectors/kafka/message#bytewax.connectors.kafka.message.KafkaSinkMessage).
+With the `.key` and `.value` fields set, or of type `KafkaSourceMessage`, if the message is unmodified.
+
+## Kafka and Recovery
+
+Typical deployments of Kafka utilize [consumer groups](https://developer.confluent.io/courses/architecture/consumer-group-protocol/)
+in order to manage partition assignment and the storing of Kafka offsets.
```
cra, line 82: Up until reading this, I was 99% sure that even when there's a single consumer it gets assigned to its own consumer group. Skimming through the docs I don't see any explicit confirmation for that, but I also don't see anything that hints otherwise. I think we should soften this particular statement, with something like

> Typically, in order to manage partition assignment and the storing of Kafka offsets, every consumer is assigned to a consumer group (for single consumer case it gets assigned to one automatically)

Also, I think we can link this https://developer.confluent.io/learn/apache-kafka-on-the-go/consumer-groups/ instead, or have both links (the one you provided is more technical, so we can have a separate mention about it like "If you are interested in learning about how Kafka uses consumer groups, check this link")

docs/articles/concepts/kafka.md
```diff
@@ -0,0 +1,167 @@
...
+flow = Dataflow("kafka_in_out")
+kinp = kop.input(
+    "inp",
+    flow,
+    starting_offset=OFFSET_STORED,
+    add_config=add_config,
+    brokers=BROKERS,
+    topics=IN_TOPICS,
+)
+```
```
cra, line 113: A bit of hand-holding but this empty space after the code snippet begs for a line like "The code above creates a dataflow input using consumer with consumer group name `consumer_group` that will automatically commit offsets to Kafka every once in a while depending on server configuration" (not sure about the last bit though, pretty sure it's 5 sec by default but maybe it is handled by client by another setting like "commit.interval.ms" or something)

docs/articles/concepts/kafka.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Partitions
+
+Partitions are a fundamental concept for Kafka and Redpanda, and are the unit of parallelism for
+producing and consuming messages.
+
+When multiple workers are started, Bytewax will internally assign individual partitions to the
```
cra, line 120: I would suggest to clarify that we mean Kafka partitions (or topic partitions for that matter)

docs/articles/concepts/kafka.md
```diff
@@ -0,0 +1,167 @@
...
+available number of workers. If there are fewer partitions than workers, some workers will not
+be assigned a partition to read from.
+
+If the number of partitions changes, Dataflows will need to be restarted in order to rebalance
+new partition assignments to workers.
+
+## Schema registry
+
+Bytewax supports integrating with the [Redpanda Schema Registry](https://docs.redpanda.com/current/manage/schema-reg/)
+as well as the [Confluent Schema Registry](https://docs.confluent.io/platform/current/schema-registry/index.html).
```
cra, line 130: This lacks one more sentence about why would you want to do this at all, either

> Sometimes you wake up and wonder what if you want to be sure about the format of my messages, now you're in luck because schema registry does just that!

just kidding, but just some reasoning as to why that thing might be useful would be good to have here

docs/articles/concepts/kafka.md
```diff
@@ -0,0 +1,167 @@
...
+import bytewax.operators as op
+
+from bytewax.dataflow import Dataflow
+from bytewax.connectors.kafka import KafkaSinkMessage, KafkaSourceMessage
+from bytewax.connectors.kafka import operators as kop
+from bytewax.connectors.kafka.registry import RedpandaSchemaRegistry, SchemaRef
+
+BROKERS = ["localhost:19092"]
+IN_TOPICS = ["in_topic"]
+OUT_TOPIC = "out_topic"
```
cra, line 145: the `OUT_TOPIC` is no used in the code below

docs/articles/concepts/kafka.md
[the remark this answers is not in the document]
Psykopear: The in-op topics have no use in the sink, when converting only `key` and `value` are ported. So you either specify a topic in the out operator, or you specify the topic in each `KafkaSinkMessage` manually. If you send a message without a topic in that case, the dataflow crashes.

docs/articles/concepts/kafka.md
```diff
@@ -0,0 +1,167 @@
...
+can be inspected and reprocessed later, while allowing the dataflow to continue processing data.
+
+## Batch sizes
+
+By default, Bytewax will consume a single message at a time from Kafka. The default
+setting is often sufficient for lower latency, but negatively affects throughput.
+
+If your dataflow would benefit from higher throughput, you can set the `batch_size` parameter
+to a higher value (eg: 1000). The `batch_size` parameter configures the maximum number of messages
+that will be fetched at a time from each worker.
```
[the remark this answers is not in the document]
whoahbot: :heavy_plus_sign: I will add a note about the time waiting.

I would also be in favor of having a higher default value for `batch_size`. I suspect that more dataflows would benefit from larger batches, but that is just a guess.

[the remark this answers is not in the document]
whoahbot: Looking at this again, we differentiate between `KafkaSource` and the kafka input operator. The operator defaults to 1000, while the `KafkaSource` defaults to 1.

I would like to set them both to 1000 and update the docstring.

docs/articles/concepts/kafka.md
whoahbot: We have some benchmarks that Fede wrote to help us assess some performance issues we encountered, but they do not include any processing steps, and are infrastructure-dependent as you mentioned.

docs/articles/concepts/kafka.md
```diff
@@ -0,0 +1,167 @@
...
+from bytewax.connectors.kafka import operators as kop, KafkaSinkMessage
+from bytewax import operators as op
+from bytewax.dataflow import Dataflow
+
+brokers = ["localhost:19092"]
+flow = Dataflow("example")
+kinp = kop.input("kafka-in", flow, brokers=brokers, topics=["in-topic"])
+processed = op.map("map", kinp.oks, lambda x: KafkaSinkMessage(x.key, x.value))
+kop.output("kafka-out", processed, brokers=brokers, topic="out-topic")
+```
```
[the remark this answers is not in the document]
Psykopear: In my head, the custom operator can offer some niceties that the source alone doesn't, but you can use the source if you don't want any of them. For example the `kop.input` operator returns the result splitted in 2 streams, and you have to manually raise on kafka error by inspecting the errors stream if you want your dataflow to crash on kafka errors. If you use the source with `op.input`, you can make it raise directly if an error occurs.

The `kop.output` offers the automatic convertion between `KafkaSourceMessage` and `KafkaSinkMessage`, because the sink implementation shouldn't bother to handle both, but the operator can add a map step before each message and do this. Here too, you might want to avoid this, and can use the sink directly

docs/articles/concepts/kafka.md
Psykopear: Well, you can't handle the exceptions inside the `KafkaSource`, but there's a `raise_on_error` option there, that is automatically handled in the operator by splitting the stream in `oks` and `errs`.
So you either let the dataflow crash on any relevant error, use `raise_on_error=False` in the source and handle the `KafkaError` type messages in a later step, or use the operator and do what you want with the error stream.

Maybe we need to clarify the options here, the `raise_on_error` parameter is never mentioned

docs/articles/concepts/kafka.md
Psykopear: something like this?

```python
errs = op.map(
    "kafka-errors",
    kinp.errs,
    lambda res: KafkaSinkMessage(
        key=None, value=json.dumps({"error": f"{res.err}", "msg": res.msg.value})
    ),
)
kop.output("dlq", errs, topic="errors", brokers=BROKERS)

```

That also makes me think that we should make the `KafkaError` (and so also `KafkaSourceMessage`) easily serializable, so that all the metadata can be sent

whoahbot: I think making `KafkaSourceMessage`  serializable is a great idea. Ideally you would want all of the source fields from the original message, as you would need them when they are re-processed.

PR #369, opened 2024-01-12
Adds a basic stub file generator

davidselassie:
- Adds stubgen
- Moves PyO3 inner module to `_bytewax`
- stubgen no longer emits trailing whitespace in docstrings
- Fixes all Python docstrings in Rust code to have proper formatting
- Topo sorts stubgen output so subclasses always work
- Makes stubgen deps depend on Python version; we don't need them in >= 3.9
- Adds `bytewax._bytewax` stub file; will need to be re-generated on re-build

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -13,7 +13,18 @@ Make a PR which commits the following:
    +version = "1.2.3"
    ```
 
-2. Commits updated API docs
+2. Commits updated stubs
+
+   ```sh
+   (.venv) bytewax/ $ maturin develop -E dev
+   (.venv) bytewax/ $ python stubgen.py bytewax._bytewax -o pysrc/bytewax/_bytewax.pyi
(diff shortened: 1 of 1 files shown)


stubgen.py
```diff
@@ -180,7 +221,12 @@ def _stub_cls(ctx: _Ctx, cls: type) -> Tuple[_Meta, ast.ClassDef]:
         body += [ast.Expr(ast.Constant(docstring, col_offset=ctx.col_offset))]
     body += [ast.Expr(ast.Constant(...))]
 
-    children = [
+    children = []
+    if "__new__" in cls.__dict__:
+        new = _stub_new(ctx.new_scope("__new__"), cls)
```
whoahbot, line 228: I'm still not seeing the correct signature for any of the rust-based classes after this change. Adding the stub for `__init__` with the same signature as the newly-added `__new__` stub fixed it for me.

whoahbot (approved):

PR #374, opened 2024-01-16
Prevent dataflow hang if next_awake is far in future

davidselassie:
Does what Fede suggested. Always advances the epoch if the epoch
interval has passed, even if the partition should not be awoken.

Tests this by making some fancier metronome inputs.


Psykopear (approved):
Looks good to me, it does solve the issue. Thanks for fixing this

PR #375, opened 2024-01-18
Resample operator

Psykopear:
Following the discussion during the office hours, this PR introduces the `resample` operator. It's a stateful operator that accumulate items for a fixed period of time, and passes the previous sample (if any) and the current batch (which might be empty) to a user provided `sampler` function.

Given a stream of floats, it can be used like this:

```python
def sampler(prev_sample: Optional[float], batch: List[float]) -> Optional[float]:
    if batch:
        return sum(batch) / len(batch)
    # prev_sample could be `None`, you might want to
    # set a default value here
    return prev_sample

resampled = op.resample("resample", stream, timedelta(seconds=0.5), sampler)
```

This will emit either the average of received values or the previously recorded average (which might be `None`) for each key at 2Hz, regardless of the input's frequency.

_edit_: A warning I should add here is that if the key space grows indefinitely, so does the state and the number of items emitted, since the keys are only discarded on EOF.

### DIFF
--- a/examples/resampler.py
+++ b/examples/resampler.py
@@ -16,12 +16,10 @@ merged = op.merge("sources", s1, s2, s3)
 def sampler(prev_sample: Optional[float], batch: List[float]) -> Optional[float]:
     if batch:
         return sum(batch) / len(batch)
-    elif prev_sample:
-        # If no items in this batch, return the previous
-        # sampled value, if any
-        return prev_sample
     else:
-        return None
+        # If no items in this batch, return the previous
+        # sampled value, which might still be `None`
+        return prev_sample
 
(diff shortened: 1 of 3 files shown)


Psykopear:
Yep, now that you make me think about it, this is just a `collect_window` and `map` with less flexibility:

```python
clock = wop.SystemClockConfig()
win = wop.TumblingWindow(timedelta(seconds=0.5), align_to)
collected = wop.collect_window("resample", merged, clock, win)
resampled = op.map("map", collected, lambda x: (x[0], len(x[1][1])))
op.inspect("resmapled", resampled)
```

PR #382, opened 2024-01-22
Shuffles around Kafka objects and updates docstrings

davidselassie:
The new docstring generator I'm using does not like re-importing
things from relative imports.

In general I'd suggest always using absolute imports and keeping
re-arranging objects via `__all__` to a minimum. Also `__all__` just
includes symbols local to that module, not submodules. So if a
submodule is not to be documented it must be prefixed with `_`, like
`_types` was previously, but e.g. `source` should have been.

I also add some docstrings.

None of the programmatic API should change; everything was moved to
where it was finally public in the previous version.

### DIFF
--- a/docs/articles/reference/migration.md
+++ b/docs/articles/reference/migration.md
@@ -49,7 +49,7 @@ add_one_stream = op.map("add_one", stream, lambda x: x + 1)
 op.output("out", add_one_stream, StdOutSink())
 ```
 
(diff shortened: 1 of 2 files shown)


Psykopear (approved):
:+1: Looks good to me, `KafkaOpOut` is definitely more readable than having the four generics on the types

PR #383, opened 2024-01-24
Using MyST + Sphinx for API docs

davidselassie:
Converts to using Sphinx + MyST + autodoc2 to build API documentation.

Part 1:

Converts all the docstrings to the new format. The biggest change is
how arguments are described using `:arg x: description`. This is all
in 1 commit.

There's also some renaming of some things to prefix them with `_` to
make them private. And removing `__all__` where it isn't necessary.

I wanted to get this merged first because it touches a lot of code and
don't want to conflict with other ongoing work.

Part 2:

Deletes all the old API doc stuff and adds in the Sphinx machinery in
`/apidocs`. There's a lot of configuration in `/apidocs/conf.py` to
setup Sphinx correctly. Also a lot of hacks around things that didn't
work right. `/apidocs/renderer.py` is a plugin to format the API docs
in a way that is closer to what we had and less boilerplate.

Part 3:

Setup `pipenv`. I wanted a way to recreate a deterministic doc
building environment. And adding extra optional dependencies to the
`bytewax` package isn't really the way to do that.

`/apidocs/Pipenv` contains all the high level dependencies (and a
fixed Python version), and `/apidocs/Pipenv.lock` contains all
dependencies.

Part 4:

I modified the `/apidocs/build.sh` script and added an
`/apidocs/autobuild.sh` script that automatically sets up the `pipenv`
venv and then runs the right Sphinx commands to build the docs / start
a little web server and rebuild the docs on changes.

All the new docs are committed into this repo in the same place as the
old `pdoc3` docs: `/apidocs/html`.

I also added some "meta documentation" in `/apidocs/writing-docs.md`
which has a cheat sheet and info about the tooling.

TODO:

I did some tweaking of the Sphinx theme to make things look vaguely
like the old API docs, but I'm sure much more could be done. Konrad
said he'd look into some of this.

Once we can get this sort of working with the current version of the
website, I'll look into ways to do a more permanent deployment.


whoahbot (approved):
This is still marked as a draft, but I think the approach looks good to me.

PR #386, opened 2024-01-29
All docs to Sphinx and RTD

davidselassie:
Unifies all our docs under `/docs` and compiles under Sphinx. This PR is mostly doing the formatting translations of the articles.

Adds the [Read the Docs](https://docs.readthedocs.io/en/stable/index.html) config and this branch is deployed at https://bytewax.readthedocs.io/all-the-doks/ . I removed the `/docs/build.sh` script since RTD will do this for us. The `/docs/autobuild.sh` script still exists for prototyping.

My thought is to try to get the basic things working here and we can iterate on Sphinx themes and other stuff in followup PRs.

### DIFF
--- a/docs/articles/advanced-concepts/custom-connectors.md
+++ b/docs/articles/advanced-concepts/custom-connectors.md
@@ -1,36 +1,14 @@
 (custom-connectors)=
 # Custom IO Connectors
 
-* why
-* how
-* tips
-* common issues and performance pitfalls to avoid
-* when to use python and when to go lower to rust
-* how to debug and make sure it's working
-* example code snippets and repos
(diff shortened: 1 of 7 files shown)


whoahbot (approved):
Thanks for putting all of this together.

I think there are a few placeholder pages that weren't linked from the previous `metadata.json` we should delete, like https://bytewax.readthedocs.io/all-the-doks/articles/concepts/streaming-and-batch.html

Structurally, this looks good to me.

konradsienkowski (approved):
Looks great 👍 `autobuild.sh` works well for prototyping, I'll use it when working on a custom theme

PR #387, opened 2024-01-30
Removes `stateful_map` `builder` function and adds `stateful_flat_map`

davidselassie:
- Removes `builder` function. Instead, the `last_state` value is
  `None` on the first time a key is encountered. You can easily call
  the `builder` manually. I think this is an easier to understand API
  because it has symmetry between the return `updated_state` value and
  the `last_state`. If you return `None` one time, you get `None` the
  next time. It also helps with not needing to construct a "zero
  value".

- Adds a `stateful_flat_map` operator. Exactly like `stateful_map` but
  you return an iterable of downstream. It seemed like many of our
  `stateful_map` functions had a followup step to filter out `None` to
  handle the "no value" case, and this allows that more gracefully.

### DIFF
--- a/docs/articles/getting-started/recovery-example.md
+++ b/docs/articles/getting-started/recovery-example.md
@@ -65,8 +65,8 @@ keyed_stream = op.key_on("key_stream", input_stream, lambda _: "ALL")
 
 
 def update_sum(current_sum, new_item):
-    if current_some is None:
-        current_some = 0
+    if current_sum is None:
+        current_sum = 0
 
     updated_sum = current_sum + new_item
     return updated_sum, updated_sum


docs/articles/getting-started/recovering-snapshot.md
```diff
@@ -57,11 +57,14 @@ keyed_stream = op.key_on("key_stream", input_stream, lambda _: "ALL")
 
 
 def update_sum(current_sum, new_item):
+    if current_some is None:
```
whoahbot, line 60: This should be `current_sum`

whoahbot (approved):

PR #398, opened 2024-02-08
Add option to include custom headers in RedpandaSchemaRegistry's requests

Psykopear:
Adds an option to include headers in each request to redpanda's schema registry.


Psykopear:
Fixed with #399

PR #400, opened 2024-02-09
Customize Sphinx docs theme

konradsienkowski:
A few customizations for pydata-sphinx-theme to make the docs look more consistent with the website & platform docs


konradsienkowski:
@davidselassie Thanks for the review. Before the merge we'll need to replace "Lorem ipsum..." part in https://github.com/bytewax/bytewax/blob/22bb5c353fcc3c2461da2610eb9085b3cc7c763b/docs/_templates/slack-footer.html

konradsienkowski:
@whoahbot it can be the same, but doesn't have to. The text you mentioned is editable via CMS

PR #404, opened 2024-02-22
Start working on custom metrics for Kafka

whoahbot:
## Overview

The premise of this PR is to provide the ability for our users to add custom instrumentation to dataflows that will be reported from our `/metrics` endpoint.

I've tried a few different approaches, and I didn't feel as though any of the ideas I've tried was a clear winner, so I'm opening this PR as a strawman to solicit some ideas on how to proceed.

I'll start with a list of general issues I encountered:

### Use of the Python Prometheus library, or OpenTelemetry client

I looked for a bit into using the Python OpenTelemetry library to collect metrics on the Python side, and register them as a [Collector](https://docs.rs/prometheus/latest/prometheus/struct.Registry.html#method.register) to the Rust-side registry. I didn't end up prototyping this approach as it felt overly complicated.

### Which Rust crates to use for metrics?

Our use of the `opentelemetry` crate's `metrics` feature was initially a little problematic as that crate does not offer a synchronous  Gauge type [yet](https://github.com/open-telemetry/opentelemetry-rust/pull/1410). The callback pattern of metrics collection that is used by the `confluent_kafka` Python library doesn't seem amenable to using an async Gauge.

The OpenTelemetry rust crate also requires the use of _both_ the [prometheus](https://docs.rs/prometheus/latest/prometheus/) crate, _and_ the [opentelemetry_prometheus](https://docs.rs/opentelemetry-prometheus/latest/opentelemetry_prometheus/) crate. The `opentelemetry_prometheus` crate depends on the `prometheus` crate to provide the registry, and the encoder that is used to generate the text format for metrics that are scraped by Prometheus.

You can see in this strawman PR that since we are already depending on the `prometheus` crate, I used the `Gauge`, `Counter`, and `Histogram` implementations that are provided there.

I also included the `prometheus` crate's `process` feature, which includes some useful metrics around process memory usage and thread counts.

### A smaller issue- Where should custom metrics get the `step_id` to use as a label for metrics?

In the `KafkaSource` example in this PR, there wasn't an obvious way to reference the `step_id` to use as a label for metrics without changing the signature for `FixedPartitionSource`. Since this approach will probably change, I didn't make any changes to the API of input sources.


Psykopear:
Sorry, I somehow missed this PR.

I agree with David, that the way it's setup right now means we'll need to update and maintain the bindings just to offer a small subset of what the underlying library offers, depending on what we think might be useful to us or the users.

But, being able to expose custom metrics in our own connectors is surely useful, and to do that we need to access the metrics backend from python.

Doing a second, separate setup of the metrics system in python did introduce some problems when parallelizing, as most of the libraries I've worked with tend to rely on a global object, but maybe that's less of a problem now that the multiprocess sheningans are relegated to the testing runner, so it could be something to explore again.

Or maybe we could try to expose a less specific api, something like a `def stats(self, **possibly_some_params) -> dict[str, list[float]]` method that can be defined in sources and sinks and is called after every `next_batch`? It's less flexible, but this way we only have to deal with the actual data for the metrics, rather than the entire metrics system.

Psykopear:
> @davidselassie @Psykopear I reworked this PR to use the Python side exporter to generate metrics from the prometheus-client library, and combine that with the output from the Rust side exporter.

I like it, it keeps the nice user facing api without us having to map it manually.
There's a source of problems though: we need to make sure to use unique names for time series in each worker. If you try to run the `custom_metrics` example as it is with multiple workers, it crashes due to a duplicate time series (each worker uses the same name). It works if you add the worker_index to each Gauge's name:

```python
class PeriodicPartition(StatelessSourcePartition):
    def __init__(self, step_id: str, worker_index: int, frequency: timedelta):
        self.frequency = frequency
        self._next_awake = datetime.now(timezone.utc)
        self._gauge = Gauge(
            f"next_batch_delay_{worker_index}",
            "Calculated delay of when next batch was called in seconds",
            ["step_id", "partition"],
            unit="seconds",
        )
        self._counter = 0
        self._metric_labels = {"step_id": step_id, "partition": "0"}
```

But it could be enough to warn users of this caveat.

pysrc/bytewax/connectors/demo.py
```diff
@@ -92,7 +92,9 @@ def list_parts(self) -> List[str]:
         return [self._metric_name]
 
     @override
-    def build_part(self, for_part: str, resume_state: Optional[_RandomMetricState]):
+    def build_part(
+        self, _step_id: str, for_part: str, resume_state: Optional[_RandomMetricState]
```
Psykopear, line 96: When overriding methods we should probably keep the original name of the kwarg, so `step_id` instead of `_step_id`, even if it's not used, so that if we pass explicitly named arguments to the subclass it still works. Same for all the other overrides in this PR

pysrc/bytewax/connectors/kafka/__init__.py
```diff
@@ -191,22 +204,53 @@ class _KafkaSourcePartition(
...
+        config: dict,
+        topic: str,
+        part_idx: int,
+        starting_offset: int,
+        resume_state: Optional[int],
+        batch_size: int,
+        raise_on_errors: bool,
     ):
         self._offset = starting_offset if resume_state is None else resume_state
+        print(f"starting offset: {starting_offset}")
```
Psykopear, line 217: Maybe log rather than print here?

Psykopear (approved):
Looks good to me :+1: I added a couple of comments, but nothing major.

pysrc/bytewax/connectors/kafka/__init__.py
whoahbot: Forgot to remove this. Thanks for pointing it out!

PR #414, opened 2024-03-01
Don't call `time_for` twice

whoahbot:
This PR attempts to remove duplicate calls to `Clock.time_for`, and is related to, but not the ultimate fix for the performance issue demonstrated in this dataflow:

https://gist.github.com/damiondoesthings/b5d16d22d18f37675a8a76e318c1bc8c

The following is a benchmark between this branch and `main` using the dataflow in that gist as a benchmark.

```shell
❯ hyperfine \
                 --parameter-list branch_name main,deduplicate_time_for \
                 --setup "git checkout {branch_name}; maturin develop --release" \
                 "python -m bytewax.run examples.slow_wax"
Benchmark 1: python -m bytewax.run examples.slow_wax (branch_name = main)
  Time (mean ± σ):     13.964 s ±  1.383 s    [User: 5.072 s, System: 12.765 s]
  Range (min … max):   12.492 s … 16.433 s    10 runs

Benchmark 2: python -m bytewax.run examples.slow_wax (branch_name = deduplicate_time_for)
  Time (mean ± σ):      9.895 s ±  0.724 s    [User: 3.945 s, System: 8.302 s]
  Range (min … max):    8.485 s … 11.191 s    10 runs

Summary
  python -m bytewax.run examples.slow_wax (branch_name = deduplicate_time_for) ran
    1.41 ± 0.17 times faster than python -m bytewax.run examples.slow_wax (branch_name = main)
```

### DIFF
--- a/src/window/clock/event_time_clock.rs
+++ b/src/window/clock/event_time_clock.rs
@@ -164,23 +164,16 @@ where
         watermark.unwrap_or(DateTime::<Utc>::MIN_UTC)
     }
 
-    fn time_for(&mut self, event: &Poll<Option<TdPyAny>>) -> Poll<Option<DateTime<Utc>>> {
-        match event {
-            Poll::Ready(Some(event)) => {
-                let event_time = Python::with_gil(|py| {
-                    self.dt_getter
-                        // Call the event time getter function with the event as parameter
-                        .call1(py, (event.clone_ref(py),))
-                        .unwrap()
-                        // Convert to DateTime<Utc>
-                        .extract(py)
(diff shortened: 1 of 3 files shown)


Psykopear (approved):
That's great. I think it's the second time we run into performance issues with the python<->rust conversion of datetimes.
We should avoid doing that when non necessary, but we can't avoid doing it in a lot of instances, so it might make sense to look if it's possible to improve performances of that. That's part of the PR on PyO3 I worked on, so next time we need to do some performance work, I can have a look at that too.

Issue #413, opened 2024-03-01
Data missing from windows when `align_to` is very long ago


davidselassie:
### Bug Description

Seemingly arbitrary changes in sliding windowing definition will cause some data points to be marked as "late" and dropped.

In the reported dataflow, this worked

```python
SlidingWindow(
    length=timedelta(milliseconds=1600),
    align_to=ALIGN_DATE,
    offset=timedelta(milliseconds=800),
)
```

but this resulted in no output

```python
SlidingWindow(
    length=timedelta(milliseconds=1600),
    align_to=ALIGN_DATE,
    offset=timedelta(milliseconds=700),
)
```

For the following dataflow

```python
from datetime import timedelta, datetime, timezone

from bytewax.dataflow import Dataflow
import bytewax.operators as op
from bytewax.connectors.stdio import StdOutSink
import bytewax.operators.window as win
from bytewax.operators.window import EventClockConfig, SlidingWindow
from bytewax.testing import TestingSource
from bytewax.tracing import setup_tracing

# from nats_connector import NATSSource


# setup_tracing(log_level="TRACE")

tsrc = TestingSource(
    [
        {
            "ts": 1709315512577,
            "sensor": {"mac": "gateway1", "type": "gw-x76"},
            "data": [
                {
                    "timestamp": 1709315512577,
                    "timezone": -10,
                    "type_code": 0,
                    "type": "badge",
                    "rssi": -81,
                    "connectable": 1,
                    "mac": "dev-id1",
                },
                {
                    "timestamp": 1709315512578,
                    "timezone": -10,
                    "type_code": 0,
                    "type": "badge",
                    "rssi": -45,
                    "connectable": 1,
                    "mac": "dev-id2",
                },
                {
                    "timestamp": 1709315512578,
                    "timezone": -10,
                    "type_code": 0,
                    "type": "badge",
                    "rssi": -58,
                    "connectable": 1,
                    "mac": "dev-id3",
                },
                {
                    "timestamp": 1709315512578,
                    "timezone": -10,
                    "type_code": 0,
                    "type": "badge",
                    "rssi": -74,
                    "connectable": 1,
                    "mac": "dev-id4",
                },
                {
                    "timestamp": 1709315512578,
                    "timezone": -10,
                    "type_code": 0,
                    "type": "badge",
                    "rssi": -63,
                    "connectable": 1,
                    "mac": "dev-id1",
                },
            ],
        }
    ]
)

flow = Dataflow("location_flow")
[...]

PR #415, opened 2024-03-05
Refactors window boundary calculation to avoid overflow

davidselassie:
Fixes https://github.com/bytewax/bytewax/issues/413

Use some slightly different number twiddling to calculate window boundaries so that there isn't an overflow when using item times that are very far from the `align_to` time. Instead of finding the window index and then multiplying by the offset to get the start time of each window, use the remainder from the index calculation to find the start of a nearby window and then adjust.


Psykopear (approved):
Nice solution, looks good to me :+1:

PR #422, opened 2024-03-18
Refactor operators

whoahbot:
This PR refactors the `branch` and `extract_key` operators to use Timely's `OperatorBuilder` and avoid taking the GIL in a tight loop.


Psykopear (approved):
Nice, another significant performance bump here, I'm loving the automated benchmarks. The build errors seem to be just about variable names, but it should be an easy fix. Looks good to me :+1:

PR #425, opened 2024-03-21
Add builds for Python 3.12

whoahbot:
- Upgrade benchmarking runner to 3.12


Psykopear (approved):
The CI builds, and the regression in performances is expected since with python 3.12 we are now collecting perf info as you noticed, so looks good to me :+1:

PR #429, opened 2024-03-26
refactor orderbook example to remove ordereddict

lfunderburk:
Last week I modified the orderbook guide to match the 0.19 version changes

Within the discussion the topic of refactoring code to remove ordered dictionaries came up (ordered by default Python >3.7). Proposed a change, change was approved. Bringing changes under the examples/orderbook.py file

[Discussion](https://github.com/bytewax/crypto-orderbook-app/pull/2#discussion_r1532760038)

### DIFF
--- a/examples/orderbook.py
+++ b/examples/orderbook.py
@@ -3,7 +3,6 @@ from dataclasses import dataclass, field
 from datetime import timedelta
 from typing import Dict, List, Optional
 
-
 import websockets
 from bytewax import operators as op
 from bytewax.connectors.stdio import StdOutSink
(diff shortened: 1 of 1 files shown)


examples/orderbook.py
```diff
@@ -1,10 +1,8 @@
 import json
-from collections import OrderedDict
 from dataclasses import dataclass, field
 from datetime import timedelta
-from typing import List, Optional
+from typing import Dict, List, Optional
 
-# pip install websockets
```
Psykopear, line 7: This comment should problably stay.

To explain why it's here, we don't have an explicit `requirements.txt` file for examples, so we write the exact dependencies as a comment. In this case it's trivial, python module and package on pypi have the same name, but sometimes they don't, and we used to have errors every time we installed the wrong package with a similar name, so we add a comment for any external package. We could remove this, but we should have a `requirements.txt` to keep track of which packages we use in all the examples.
*edit*: I see that some other examples don't have all the packages specified, but I'd rather add it there than remove it here

Psykopear (approved):
Added a small comment, but looks good to me :+1:

lfunderburk:
Thank you Federico! I'm not sure what etiquette you follow when merging PRs
Do we wait for more approvals/comments?

PR #433, opened 2024-04-01
Move windowing operators into Python

davidselassie:
In narrative order:

- Transforms `unary` into a `stateful_batch` operator that takes an entire list of values for a key. This let's us do "bulk setup" to help mitigate performance issues. Instead of a `UnaryLogic.on_item`, there's now a `UnaryLogic.on_batch`.

- `unary` was renamed to `stateful`. It's now a shim over `stateful_batch`. The API is basically the same.

- Moves the `ClockConfig` / `Clock` interface from Rust to Python as `Clock` / `ClockLogic`. These are basically the same as the Rust interfaces, but split out the handling of "new item" vs "EOF" vs "timeout" callbacks instead of the overloaded `Poll<Option<V>>` thing. I think it makes it a little easier to read. Adds one new callback called `on_batch` which lets you prep for a series of items.

- Moves the `WindowConfig` / `Windower` interface from Rust to Python as `Windower` / `WindowerLogic`. It is the same as the Rust interface but now with a `merged` callback; this lets us handle out-of-order items and session windows.

- Re-implements our clocks and windowers. APIs are unchanged, just slight naming changes: remove lingering `*Config`.

- The base window operator is called `window` and is configured by subclassing a `WindowLogic`. It feels very similar to `unary` but has different callbacks for the windowing events. It also returns a triple of streams in a `WindowOut`: downstream items, late items, and window metadata (whenever it changes). All are sub-keyed by window ID so you could join later.

- Re-implements `fold_window` in terms of `window`. This requires adding a `merger` callback to know how to handle window merges.

- Every window operator now returns `WindowOut`, so not backwards compatible.

- Ports all of our `SlidingWindower` boundary tests to Python. Other Rust tests ported too.

### DIFF
--- a/pysrc/bytewax/operators/window.py
+++ b/pysrc/bytewax/operators/window.py
@@ -145,6 +145,21 @@ class ClockLogic(ABC, Generic[V, S]):
         ...
 
     @abstractmethod
+    def to_system_utc(self, timestamp: datetime) -> Optional[datetime]:
+        """Convert a timestamp to a UTC system time.
+
+        This is used to determine when the runtime should wake up to
+        check for closed windows.
+
+        :arg timestamp: Previously returned by `on_item`.
+
+        :returns: A system time in UTC. If `None`, signal that there
+            is no relevant system time to wake at.
+
+        """
+        ...
+
+    @abstractmethod
     def snapshot(self) -> S:
         """Return a immutable copy of the state for recovery.
 
@@ -196,6 +211,10 @@ class _SystemClockLogic(ClockLogic[V, None]):
         return self._now
 
     @override
+    def to_system_utc(self, timestamp: datetime) -> Optional[datetime]:
+        return timestamp
+
+    @override
     def snapshot(self) -> None:
         return None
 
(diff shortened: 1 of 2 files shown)


Psykopear (approved):
Looks good to me :+1: I added some comments, but nothing major.
As far as performance goes, we still had to do this sooner or later for several reasons, so I'm ok with that. We can try to squeeze out more performance if we need in the future.

pysrc/bytewax/operators/window.py
```diff
@@ -161,67 +1357,95 @@ def collect_window(
...
     step_id: str,
     up: KeyedStream[V],
-    clock: ClockConfig,
-    windower: WindowConfig,
+    clock: Clock,
+    windower: Windower,
     builder: Callable[[], S],
     folder: Callable[[S, V], S],
-) -> KeyedStream[Tuple[WindowMetadata, S]]:
+    merger: Callable[[S, S], S],
```
whoahbot, line 1447: I do think that your suggestion of finding a way to have a default for the merger here is a good one. It would be nice if users of `fold_window` aren't required to supply a placeholder callable. I can understand how eliding it is straddling a fine line, but I couldn't think of other shapes of this function that would remove the need for this.

whoahbot (approved):
:heavy_plus_sign: Looks good overall. Left you some rambling thoughts about the signature for `fold_window`.

PR #438, opened 2024-04-04
"Reproducible" dev and CI envs

davidselassie:
After a long and tortured journey of trying things out to see what works best, here's what I settled on. I didn't like Poetry or Hatch or other full project managers because it didn't feel like there were any that supported all the things we want without having to fight the tool: lock files, multiple venvs with arbitrary dependencies, running multiple Python versions.

Check out https://bytewax--438.org.readthedocs.build/438/guide/contributing/local-development.html for the docs on how to set up the dev env locally.

* `pyproject.toml` only defines the library dependencies and project tooling.

* Use `just` as a way to write scripts to facilitate common dev actions.

* Use `uv pip {compile,sync}` as the way to define and create reproducible venvs. All venv deps are defined in `requirements/*.in` and lock files live in `requirements/*.txt`.

* Have a "preferred place" for the various venvs in `venvs/`. The `venvs/dev/` one is the main one you can point your LSP / IDE at.

Try out the step-by-step in the docs and let me know how things work.

Shuffles around what runs as a pre-commit and what runs as a lint. Pre-commit is just formatting. `just lint` will run all the other code lints. CI runs `just lint` as part of a "repo tests" or tests that don't depend on a specific version of Python and just assert things about the code: lints, Rust tests, doctests, double check pre-commit passes.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -15,7 +15,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  repo-tests:
+  repo-checks:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -17,22 +15,33 @@ concurrency:
...
+        shell: bash
+        run: |
+          uv venv venv
+          . venv/bin/activate
+          uv pip sync --strict requirements/dev.txt
+      - name: Run lints
+        shell: bash
+        run: |
+          . venv/bin/activate
+          just _test-repo
```
whoahbot, line 44: `_test-repo` also runs the rust tests we have. Should we change the name of this CI job?

whoahbot (approved):
This is fantastic! I didn't have any issues other than the fix I commented in the justfile.

Thanks for seeing this one through.

PR #439, opened 2024-04-04
Upgrade pyo3 to 0.21.1

whoahbot:


Psykopear:
> I'll spend some more time before I merge this PR looking for other places where we might see some benefit. There are a few calls to as_gil_ref that I should spend some time removing as well.

I mean, the scope of this PR is the PyO3 upgrade, and we have that without any regression, so I don't think there's any need to work on performances here, we can merge this as it is. I was just surprised the benchmarks weren't affected at all (in any direction), but this upgrade is still needed regardless of performances.

PR #442, opened 2024-04-15
Add a performance guide.

whoahbot:
Any other suggestions for topics that would be important to cover here?


docs/guide/advanced-concepts/performance.md
```diff
@@ -0,0 +1,195 @@
...
+and in this document we'll describe concrete design recommendations you should
+follow while implementing your dataflow to play to the strengths of Bytewax and
+ensure your dataflow is performant.
+
+## Bytewax's worker execution model
+
+A Bytewax **worker** is a thread that is executing your dataflow. Workers can be grouped
+into separate processes, but refer to a single thread within a single process.
+
+Bytewax's execution model uses identical workers. Workers execute all of the steps
```
Psykopear, line 32: This might be a bit misleading, as not every worker will always execute every step, I'd phrase it a bit differently:

```suggestion
Bytewax's execution model uses identical workers. Workers can execute any of the steps
```

docs/guide/advanced-concepts/performance.md
```diff
@@ -0,0 +1,195 @@
...
+
+When multiple workers are started within the same Python process using
+`-w/--workers-per-process`, they must contend over Python's GIL, or
+{external+python:std:term}`global interpreter lock`. Python's GIL ensures
+that only one Python thread can run at a time, preventing race conditions and
+ensuring thread-safety.
+
+In practice this means that when you are attempting to increase the parallelism
+of a dataflow, should try starting with multiple processes using the
+`-i/--process-id`, `-a/--addresses` instead of multiple workers with `-w`.
```
Psykopear, line 47:
```suggestion
In practice this means that when you are attempting to increase the parallelism
of a dataflow, you should try starting with multiple processes using the
`-i/--process-id`, `-a/--addresses` instead of multiple workers with `-w`.
```

docs/guide/advanced-concepts/performance.md
```diff
@@ -0,0 +1,195 @@
...
+
+If your code can make use of libraries that make progress without holding the GIL,
+adding workers to each process with `-w` may increase performance.
```
Psykopear, line 50: Since it's always hard to know what libraries release the GIL, I'd add an example of a part of the standard library that does:

```suggestion
If your code can make use of libraries that make progress without holding the GIL
(like python's urllib), adding workers to each process with `-w` may increase performance.
```

Psykopear (approved):
Looks good to me, I added some minor comments

docs/guide/advanced-concepts/performance.md
whoahbot: > What is the specific nuance that is being pointed out here? All workers have an instance of all steps but not all workers might get routed data / assigned an IO partition and so thus "don't execute the step"?

I wanted to avoid any inference that the execution model shares similarities with a task manager model. I think your phrasing helps clarify that point.

PR #443, opened 2024-04-17
More Deployment in Documentation

awmatheson:
Adds waxctl to documentation pages as well as an overview for the deployment section.


docs/guide/deployment/overview.md
```diff
@@ -0,0 +1,32 @@
...
+
+```bash
+waxctl run mydataflow.py -p 2
+```
+
+You can find more information in the [waxctl documentation](#xref-waxctl)
+
+## Running in Kubernetes
+
+Kubernetes is the most common orchestration platform for running containers. You can use the publicly available [helm chart](#xref-helm) to run Bytewax on a kubernetes cluster.
```
miccioest, line 28: If we want to encourage people to use `waxctl` instead of the helm chart, I believe we should only mention `waxctl` to deploy on Kubernetes and not even mention the helm chart.

miccioest (approved):
Thanks for doing this!

I left a small suggestion.

lfunderburk (commented):
https://github.com/bytewax/bytewax/blob/docs-updates/docs/guide/deployment/helm-chart.md seems to be making references to images that no longer exist in the repository

Screenshot 2024-04-19 at 3 32 48 PM

Screenshot 2024-04-19 at 3 33 09 PM

docs/guide/deployment/overview.md
lfunderburk: I wonder if it makes sense to remove that file and references from the directory?

https://github.com/bytewax/bytewax/blob/docs-updates/docs/guide/deployment/helm-chart.md

lfunderburk: What is the difference between the two methods?

docs/guide/deployment/overview.md
```diff
@@ -0,0 +1,31 @@
+(xref-deployment)=
+# Deployment Overview
+
+There are many ways you can run Bytewax, in this section we will highlight the main ways we recommend deploying Bytewax.
+
+## The Bytewax Platform
+
+We built the Bytewax Platform to make deploying, managing and operating dataflows. You can find out more about the platform features on the [Bytewax website](https://www.bytewax.io/platform) and the [platform documentation](https://platform.bytewax.io/).
```
whoahbot, line 8: > We built the Bytewax Platform to make deploying, managing and operating dataflows

To make it easier?

docs/guide/getting-started/execution.md
```diff
@@ -29,6 +31,14 @@ all of the available runtime options, run the following command:
 > python -m bytewax.run --help
 ```
 
+### Running with waxctl
```
whoahbot, line 34: I think we need to add some information that using `waxctl` still requires a Python installation, with at least Bytewax installed before a dataflow can be run.

docs/guide/getting-started/execution.md
```diff
@@ -65,6 +75,18 @@ To run this flow use `simple` because creating a file named
...
 ```
 
+### Running dataflows with waxctl
+
+The `waxctl run` command expects only one argument, a path or URI of a python script:
+
+```console
+$ waxctl run simple.py
+
+You can also run a remote dataflow in a public GitHub repo with `waxctl`:
```
whoahbot, line 85:
```suggestion
You can also run a dataflow from code contained in a public GitHub repo with `waxctl`:
```

docs/guide/deployment/overview.md
awmatheson: Updated this

awmatheson:
> https://github.com/bytewax/bytewax/blob/docs-updates/docs/guide/deployment/helm-chart.md seems to be making references to images that no longer exist in the repository
> Screenshot 2024-04-19 at 3 32 48 PM Screenshot 2024-04-19 at 3 33 09 PM

@lfunderburk for some reason this is only in the repo and not when the docs are deployed.

lfunderburk:
Ok in this case discard my comment

PR #448, opened 2024-04-24
Re-enables doctests

davidselassie:
Re-enables testing code in the documentation in CI. We use the Sphinx doctest runner to run the tests. There's no good way to run these tests through pytest because to truly know what content is in the final Markdown, you have to get Sphinx to compile the documentation, so let's just lean into that. This is run as part of the `repo-checks` CI step.

Goes through and replaces `python` code blocks with `{testcode}`. This will cause the Sphinx doctest runner to find those blocks and test them. They are still rendered in the final documentation as Python code. `{testoutput}` marks what output should be asserted.

A trick we can use to have nicely formatted documentation but still test full dataflows is to use a hidden `{testcode}` block which runs `run_main(flow)` and then assert the output.

Adds a `requirements/doctests.in` so you can add external deps for running doctests. We use `requests` and `orjson` in some.

Writes up how to write your own doctests in an updated `docs/guides/contributing/writing-docs.md`.

### DIFF
--- a/docs/guide/contributing/local-development.md
+++ b/docs/guide/contributing/local-development.md
@@ -6,13 +6,17 @@ environment setup.
 
 ## Environment Tools
 
-### Installing Rust
+### Install Rustup
 
 Install a Rust compiler stack. See the [Rust Project's installation
 documentation](https://www.rust-lang.org/learn/get-started) on how to
 do that using `rustup`.
 
-### Installing `just`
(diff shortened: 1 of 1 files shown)


lfunderburk (approved):
Thank you David. The new tag usage to test code looks clean and a great addition!

whoahbot (approved):

PR #450, opened 2024-04-25
Fix Mac CI

davidselassie:
Absolute epic saga. Tried a lot of stuff to get to this point, but it shook out that two main things were wrong:

1. Typo'd the maturin action's `target` so we weren't building wheels for the correct arch on each runner. This wasn't the cause of CI failing, but would mean we didn't make the correct number of wheels on next release.

2. Subprocess Python interpreters weren't being spawned in the correct virtualenv.

As part of https://github.com/bytewax/bytewax/pull/438 we were running tests in a virtualenv rather than installing everything into the system interpreter. This means that when we launch a Python interpreter subprocess, we have to ensure that it is running in the same virtualenv otherwise it won't find the packages for testing. We were launching subprocesses in two places: in the Ctrl-C tests to have a process to SIGINT, and in `python -m bytewax.testing` in order to start multiple workers. The most surefire way to guarantee you are in the same virtualenv is to invoke the Python executable using the path inside of the virtualenv[^1].

For the Ctrl-C tests, it's pretty easy to change the invocations to use `sys.executable` rather than `"python"`.

For the `bytewax.testing` module main, things got trickier and demonic. Originally this code delegated to a `test_cluster` function defined on the Rust side via PyO3. This code would look at the number of processes requested and spawn that many procs. It used a trick to spawn the same Python interpreter: it would re-use attempt to access the original invoked interpreter via `std::env::args()` and then use `args[0]`. One might assume that this is the same as on the Python side `sys.argv`, which should be the correct interpreter[^2]. It turns out that that is not the case, and the underlying native code `argv` array (which Rust uses) is left in a not correct state and points to the global path of the Python interpreter[^3]. 

That means the original Rust-side `bytewax.testing` subprocess spawning code was spawning interpreters that were not in the virtualenv.

All this was resolved by re-implementing the `bytewax.testing` module main totally on the Python side and being able to explicitly use `sys.executable`.

---

A few other smaller things that happened on the journey:

- Update the Ctrl-C tests to have timeouts and much better error reporting. We shouldn't hang CI anymore.

- Add two other Ctrl-C tests which actually test the production endpoint of `python -m bytewax.run` those are much more important than the `bytewax.testing` endpoint.

- Bumps the cache action version so you don't need the GNU tar hack anymore.

- I'm not totally sure why, but `macos-latest` is now an ARM runner, so only runs the tests on Mac when it's an ARM builder (otherwise you get "unsupported arch" errors because we only built the current arch wheel).

[^1]: Side note. Yes, `. venv/bin/activate` sets an env var `$VIRTUAL_ENV` but it is _ignored_ by any Python interpreter. The interpreter itself parses the absolute path used to invoke it to know which venv it is in. (You can test this by activating, then running `/usr/bin/python` and noting you aren't the venv anymore.) The env var appears to be just to help out other tooling. Or `$PATH` is used in a round-about way to ensure that when you use an unqualified `python` it uses the one in the venv. But just relying on path resolution is frought because you are totally not required to use `activate` and so might not have the `$PATH` shimmed.

[^2]: Well, only if the venv was activated because `$PATH` is the same so you get the same shim. `sys.argv[0]` is not always the absolute path.

[^3]: I believe this is due to later versions of the Python interpreter re-write the arguments when you do `python -m module` for unclear to me reasons. See the existence of [`sys.orig_argv`](https://docs.python.org/3/library/sys.html#sys.orig_argv). I assume checking the validity of the native-side array is not part of the Python interpreter's contract, since the Python-side is correct.

### DIFF
--- a/pysrc/bytewax/_bytewax.pyi
+++ b/pysrc/bytewax/_bytewax.pyi
@@ -233,23 +233,6 @@ def setup_tracing(tracing_config=None, log_level=None):
     """
     ...
 
-def test_cluster(
-    flow,
-    *,
-    epoch_interval=None,
-    recovery_config=None,
-    processes=1,
-    workers_per_process=1,
-):
-    """Execute a Dataflow by spawning multiple Python processes.
-
-    Blocks until execution is complete.
-
-    This function should only be used for testing purposes.
-
-    """
-    ...
-
 class AbortExecution(RuntimeError):
     """Raise this from `next_batch` to abort for testing purposes."""
 
--- a/pysrc/bytewax/testing.py
+++ b/pysrc/bytewax/testing.py
@@ -1,16 +1,17 @@
 """Helper tools for testing dataflows."""
 
+import asyncio
+import sys
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
 from itertools import islice
-from typing import Any, Iterable, Iterator, List, Optional, Union
+from typing import Any, Dict, Iterable, Iterator, List, Optional, Union
 
 from typing_extensions import override
 
 from bytewax._bytewax import (
     cluster_main,
     run_main,
-    test_cluster,
 )
 from bytewax.inputs import (
     AbortExecution,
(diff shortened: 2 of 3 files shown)


whoahbot (approved):
Wow, what an epic tale.

Nice work tracking that down!

PR #456, opened 2024-05-01
Substitute in the current Bytewax version into docs

davidselassie:
Read the current defined version from `Cargo.toml`, save it as a
Sphinx substitution variable. Enable a Sphinx / MyST extension that
lets you put subsitutions in code blocks using `|var|` syntax. Use
that in all the spots in the docs.

Adds that extension as a dep into the reproducible venv.

### DIFF
--- a/docs/guide/contributing/writing-docs.md
+++ b/docs/guide/contributing/writing-docs.md
@@ -354,7 +354,7 @@ $ pip install bytewax==|version|
 ```
 ````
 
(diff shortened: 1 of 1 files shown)


whoahbot (approved):
Looks good, do you have any thoughts on versioning strategies for dev versions? As in- should we have a versioning strategy so that the `latest` docs don't reference an incorrect version?

PR #461, opened 2024-05-09
Windowing operators process items in timestamp order

davidselassie:
Previously the low-level `window` operator would apply items in the
order they came in, regardless of the timestamp order they had. This
had some subtle probably confusing effects like:

- `collect_window` did not result in items in each `list` in order

- `fold_window` did not apply folds in order

This PR introduces a buffer `queue` inside of `_WindowLogic` to keep
incoming items in timestamp order and only flush them to the logic
when the watermark has passed so we know no out-of-order items are
coming.

This introduces no new latency on most of the windowing operators
because they already wait until the window is closed to emit a result
(so waiting until we have the complete ordering to run the logic
doesn't really slow anything down). The low-level `window` technically
now also would have to wait for the watermark to pass to emit
mid-window items, but we haven't seen anyone need that yet.

If people are extremely performance or latency sensitive but don't need ordering, they can enable the `ordered=False` option to eagerly apply incoming items to the window.

### DIFF
--- a/pytests/operators/windowing/test_fold_window.py
+++ b/pytests/operators/windowing/test_fold_window.py
@@ -1,6 +1,7 @@
 from collections import defaultdict
+from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
-from typing import Dict
+from typing import Dict, List, Tuple
 
 import bytewax.operators as op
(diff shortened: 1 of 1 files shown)


whoahbot (approved):
:heavy_plus_sign:

PR #462, opened 2024-05-10
`join` and `join_window` operators now have modes

davidselassie:
Now that we have windowing operators in Python with a more flexible interface, we can augment the features of the join operators to allow emitting join sets mid-way through a window. This means that both normal `join` and `join_window` support all the possible join modes. Adds the logics to enable that.

Remove the `join_named` and `join_window_named` operators. It is not possible to annotate them in a way where they can return any helpful type information other than `Dict[str, Any]` so you lose all type hints. Since we want to strongly encourage people to use type hints, remove them. No join functionality is lost; you can still use `join` and `join_window` and do all the same types of joins, just the API is `tuple`-based (and has type hints) rather than `dict`-based.

### DIFF
--- a/docs/guide/concepts/joins.md
+++ b/docs/guide/concepts/joins.md
@@ -125,7 +125,7 @@ behavior of the {py:obj}`~bytewax.operators.join` operator takes any
 number of upstream sides and waits until we have seen a value for a
 key on all sides of the join, and only then do we emit the gathered
 values downstream as a {py:obj}`tuple` of the values in the same order
-as the sides stream arguments. Bytewax calls this a **complete join**. 
+as the sides stream arguments. Bytewax calls this a **complete join**.
 This is similar to an _inner join_ in SQL.
 
 Let's see that in action. To recap our example:


docs/guide/concepts/joins.md
```diff
@@ -127,8 +125,8 @@ behavior of the {py:obj}`~bytewax.operators.join` operator takes any
 number of upstream sides and waits until we have seen a value for a
 key on all sides of the join, and only then do we emit the gathered
 values downstream as a {py:obj}`tuple` of the values in the same order
-as the sides stream arguments. When we wait for a value from all
-sides, Bytewax calls this a **complete** join. This is similar to an
+as the sides stream arguments. When emit once we have a value from all
+sides, Bytewax calls this a **complete join**. This is similar to an
 _inner join_ in SQL.
```
whoahbot, line 130:
```suggestion
as the sides stream arguments. Bytewax calls this a **complete join**. 
This is similar to an _inner join_ in SQL.
```

whoahbot (approved):
Left you one suggested docs fix.

PR #469, opened 2024-05-21
Fixes for re-joining `WindowMetadata`

davidselassie:
As part of thinking through
https://github.com/bytewax/bytewax/pull/466#discussion_r1607447828 and
how to properly re-join windowing operator output with metadata, I
realized there were a few little gotchas surrounding metadata and
joins that made actually performing that annoying.

Previously, the `windowing` operator (and thus all windowing operators
based on it) would emit metadata as soon as a window was opened, then
would emit updates whenever a window merged with it. This made it hard
to design the join afterwards, as you had no way of knowing for sure
that you had the final metadata. This PR now only emits the window
metadata once on window close. This means you know for sure you have
the final metadata and can focus on just thinking about how to
interpret the other side of the join.

It's now also possible to use the metadata as a "close trigger" in
some situations, since it is only emitted then.

`WindowMetadata` now includes a `merged_ids` set. This can be used to
re-construct more complicated session windowing situations. It's
possible that a windowing operator will emit data from a still open
window, so the data will be labeled with a window ID that later is
merged.

This change also adds some new modes to the `join` and `join_window`
operators to make these kinds of joins easier. There are now 2 mode
categories: "insert mode" which determines what happens when you get
multiple values on a side, and "emit mode" which determines when is a
join row emitted downstream. Having both these lets you do things like
"product complete" joins which would be useful for joining metadata to
windowing operators which emit multiple items downstream.

All of this makes doing the joins a little easier, but there is still
a lot of complexity in this area because the `windowing` operator is
so flexible.

### DIFF
--- a/pysrc/bytewax/operators/windowing.py
+++ b/pysrc/bytewax/operators/windowing.py
@@ -435,7 +435,7 @@ class WindowMetadata:
     close_time: datetime
     """The timestamp this window closed."""
     merged_ids: Set[int] = field(default_factory=set)
-    """Any window IDs that were eventually merged into this window."""
+    """Any original window IDs merged into this window before close."""
 
 
 class WindowerLogic(ABC, Generic[S]):
@@ -505,9 +505,8 @@ class WindowerLogic(ABC, Generic[S]):
         items.
 
         You only need to report a merge once. After a window ID has
-        been returned as an original window, that ID will not be
-        passed to {py:obj}`~WindowerLogic.metadata_for` nor does it
-        need to be reported as closed.
+        been returned as an original window, that ID does not need to
+        be reported as closed.
 
         :returns: An iterable of 2-tuples of `(original_window_id,
             target_window_id)`.
@@ -530,7 +529,8 @@ class WindowerLogic(ABC, Generic[S]):
             regress across calls.
 
 
-        :returns: A list of now closed window IDs.
+        :returns: A list of now closed window IDs and the their final
+            metadata.
 
         """
         ...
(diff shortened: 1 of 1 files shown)


pysrc/bytewax/operators/windowing.py
```diff
@@ -1607,107 +1590,61 @@ def shim_builder(resume_state: Optional[S]) -> _FoldWindowLogic[V, S]:
...
-@dataclass
-class _JoinWindowRunningLogic(WindowLogic[Tuple[int, V], Tuple, _JoinState]):
-    state: _JoinState
+        if self.insert_mode == "first" and not self.state.is_set(join_side):
+            self.state.set_val(join_side, join_value)
+        elif self.insert_mode == "last":
+            self.state.set_val(join_side, join_value)
+        elif self.insert_mode == "product":
+            self.state.add_val(join_side, join_value)
+        else:
```
whoahbot, line 1618: `insert_mode: first` can fall through to the else case here and raise an error when values are provided out of order:

```python
def test_join_window_first_complete() -> None:
    inp_l = [
        _Event(align_to + timedelta(seconds=1), 2),
        _Event(align_to, 1),
    ]
    inp_r = [
        _Event(align_to + timedelta(seconds=1), 2),
    ]
    out: List[Tuple[Optional[int], Optional[int]]] = []

    flow = _build_join_window_dataflow(inp_l, inp_r, out, "first", "complete")

    run_main(flow)
    assert out == [
        (1, 2),
    ]
```

whoahbot (approved):

PR #472, opened 2024-05-31
Only call `notify_at` on keys awoken that activation

davidselassie:
We were accidentally using the buffer of "all the keys that have been awoken this epoch" and interpreting it as "all the keys that were awoken on this operator activation" so `notify_at` was being called superfluously on all keys that were seen the entire epoch after every activation of the operator, even if no state could have changed, and thus the next notify time could not have changed.

The previous behavior was "technically correct" in that it resulted in `on_notify` being called at the right time, but this should help with performance.


whoahbot (approved):
Nice catch!

PR #474, opened 2024-06-03
Adds a `TestingSource.PAUSE` special item

davidselassie:
This is useful for testing stuff that interacts with the notification
system in stateful operators in "integration" style, run an actual
dataflow tests.

This should be used sparingly. First, if there is a way to phrase your problem using event time, do that instead of system time. Then have a system "now getter" mock point in
your stateful operator logic so you can unit test your `notify_at`
return value. But at some point you should do an actual test in a
dataflow that uses the awakening machinery that goes through Timely,
and there's no good way to mock that out.

### DIFF
--- a/pysrc/bytewax/testing.py
+++ b/pysrc/bytewax/testing.py
@@ -78,7 +78,14 @@ def ffwd_iter(it: Iterator[Any], n: int) -> None:
 class _IterSourcePartition(StatefulSourcePartition[X, int]):
     def __init__(
         self,
-        ib: Iterable[Union[X, "TestingSource.EOF", "TestingSource.ABORT"]],
+        ib: Iterable[
+            Union[
+                X,
+                "TestingSource.EOF",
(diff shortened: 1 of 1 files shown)


pysrc/bytewax/testing.py
```diff
@@ -180,6 +186,12 @@ class ABORT:
 
         _triggered: bool = False
 
+    @dataclass
+    class PAUSE:
+        """Signal this input to not emit items for a duration."""
+
+        for_duration: timedelta
+
     def __init__(self, ib: Iterable[Union[X, EOF, ABORT]], batch_size: int = 1):
```
whoahbot, line 195: Should the iterable type here, and the `ib` param for `_IterSourcePartition` here include `PAUSE`?

whoahbot (approved):
Looks good. I left you a comment about the type signatures where we would use `TestingSource.PAUSE`.

PR #484, opened 2024-07-06
init guide migration missing values

lfunderburk:
Migration of guide 

**https://bytewax.io/guides/handling-missing-values**


docs/tutorials/handling-missing-data/index.md
```diff
@@ -0,0 +1,175 @@
...
+
+Generate a dataflow that will impute missing values in a stream of data of random integers and nan values.
+
+We can represent our dataflow - called map_eg through this diagram, in which the data flows through three key steps:
+
+* input: includes a random integer between 0 and 10 or a numpy nan value for every 5th value
+* stateful map to impute the values: we will create a custom window to impute the missing values
+* output: output the data and the imputed value to standard output
+
+![mermaid](mermaid.png)
```
whoahbot, line 38: It should be possible to include the mermaid diagram code here. Some examples from our docs:

https://github.com/bytewax/bytewax/blob/main/docs/guide/concepts/rescaling.md?plain=1#L33

I think it is desirable to use the mermaid code itself, as it renders itself in a more mobile friendly manner.

docs/tutorials/handling-missing-data/index.md
```diff
@@ -0,0 +1,175 @@
...
+:lineno-match:
+```
+
+## Input Code
+
+For this example we will mock up some data that will yield either a random integer between 0 and 10, or a numpy nan value for every 5th value we generate.
+
+To simulate the generation of random numbers and nan values, we will create a class called `RandomNumpyData` that will return a random integer between 0 and 10 or `nan` value for every 5th value. We will build this class such that it inherits from the {py:obj}`~bytewax.inputs.StatelessSourcePartition`. This will enable us to create our input as a Bytewax input partition that is stateless.
+
+We will then create the `RandomNumpyInput` class - this acts as a wrapper for RandomNumpyData, facilitating dynamic data generation based on the distribution of work across multiple workers in a distributed processing system. When the data source needs to be built (e.g., at the start of a processing step or when distributed across workers), it simply creates and returns an instance of `RandomNumpyData`.
```
whoahbot, line 74: We note in the input sources that the `RandomNumpyData` class is based on worker distribution, but maybe we can clarify the language here to say that we create an instance of that class on every worker to explain the mechanism a bit better.

docs/tutorials/handling-missing-data/index.md
```diff
@@ -0,0 +1,175 @@
...
+:caption: dataflow.py
+:language: python
+:start-after: start-dataflow
+:end-before: end-dataflow
+:lineno-match:
+```
+
+## Custom Window Using Stateful Map
+
+Before we dive into the code, it is important to understand the stateful map operator. Stateful map is a one-to-one transformation of values in (key, value) pairs, but allows you to reference a persistent state for each key when doing the transformation. The stateful map operator has two parts to it: a builder function and a mapper function. The builder function will get evoked for each new key and the mapper will get called for every new data point. For more information on how this works, the api docs have a great explanation.
```
whoahbot, line 100:
```suggestion
Before we dive into the code, it is important to understand the stateful map operator. Stateful map is a one-to-one transformation of values in (key, value) pairs, but allows you to reference a persistent state for each key when doing the transformation. For more information see the API docs {py:obj}`~bytewax.operators.stateful_map`.
```

whoahbot (approved):
Looks good, left you some suggestions and a fix for the API of `stateful_map`.

docs/tutorials/handling-missing-data/index.md
lfunderburk: addressed here, let me know if this sounds better

https://github.com/bytewax/bytewax/pull/484/commits/bdfdb519c2ec58534fcf97c9ab98cd4e47b1e9a2

https://github.com/bytewax/bytewax/pull/484/commits/ea7b3befb385c71be58e4cada74a6d5bcf7bb718

docs/tutorials/handling-missing-data/index.md
lfunderburk: Tried it, but get not very helpful feedback of what went wrong

Screenshot 2024-07-10 at 4 35 29 PM


Do you see what I have done wrong?

```mermaid
graph TD;
    subgraph W1[map_eg_input]
        direction TD
        W1I[map_eg_input] --> W1O[map_eg.input_down portout]
    end
    W1O --> W2I[map_eg.impute_up]
    subgraph W2[map_eg_impute]
        direction TD
        W2I --> W2M[map_eg_impute]
        W2M --> W2O[map_eg.impute_down portout]
    end
    W2O --> W3I[map_eg.output_up]
    subgraph W3[map_eg_output]
        direction TD
        W3I --> W3O[map_eg_output]
    end
```

lfunderburk: omg GitHub comments help me see what went wrong

lfunderburk: will use Bytewax to generate the mermaid code itself

Issue #485, opened 2024-07-11
Sliding window intersect sometimes does not calculate overlap correctly


davidselassie:
### Bug Description

Reported in https://bytewaxcommunity.slack.com/archives/C035SBNBECW/p1720651075653099?thread_ts=1720054435.601889&cid=C035SBNBECW

I'm not totally sure why this is happening, but a timestamp that should fall in a single window, is being assigned to two windows.

### Python version (`python -V`)

Python 3.12.1

### Bytewax version (pip list | grep bytewax)

bytewax 0.21.1

### Operating System version (`uname -morp`)

6.8.0-76060800daily20240311-generic x86_64 x86_64 GNU/Linux

### Steps to Reproduce

```python
def test_intersect_bug():
    logic = _SlidingWindowerLogic(
        length=timedelta(seconds=5.5),
        offset=timedelta(seconds=5),
        align_to=datetime(2024, 7, 1, tzinfo=timezone.utc),
        state=_SlidingWindowerState(),
    )

    assert logic.intersects(
        datetime(2024, 7, 1, 0, 8, 33, 40400, tzinfo=timezone.utc)
    ) == [102]

```

This should result in only one window assignment, but results in 2: `[102, 103]`.

PR #486, opened 2024-07-11
Fix issue with `SlidingWindower` assigning to too many windows

davidselassie:
Should fix https://github.com/bytewax/bytewax/issues/485

I think what was happening was I was incorrectly using the distance from the end of the previous window as the time to the start of the next window to calculate if we needed to have the maximum overlap number of windows.

In any case, I did some more thinking and found a much simpler algorithm that gives us the same behavior. Adds tests for the missed cases.


whoahbot (approved):
Nicely fixed!

PR #515, opened 2024-09-25
Clarify event time table

davidselassie:


docs/guide/concepts/windowing.md
```diff
@@ -94,13 +94,25 @@ late. We will effectively wait for up to
...
 
-| Real-Time | Current Watermark | Incoming Timestamp | Late? | Possible Watermark | Accepted? | Resulting Watermark |
-| --------- | ----------------- | ------------------ | ------| ------------------ | --------- | ------------------- |
-| 10:32:00  | None              | 10:35:00           | No    | 10:30:00           | Yes       | 10:30:00            |
-| 10:33:00  | 10:30:00          |                    |       |                    |           | 10:31:00            |
-| 10:34:00  | 10:32:00          | 10:38:00           | No    | 10:33:00           | Yes       | 10:33:00            |
-| 10:36:00  | 10:35:00          | 10:39:00           | No    | 10:34:00           | No        | 10:34:00            |
-| 10:38:00  | 10:37:00          | 10:36:00           | Yes   |                    |           | 10:37:00            |
+In the following table, **system time** represents the real clock time
+receive an item or are updating the watermark, **system time of max**
```
whoahbot, line 98: I think this sentence got cut off.

docs/guide/concepts/windowing.md
```diff
@@ -94,13 +94,25 @@ late. We will effectively wait for up to
...
+is the system timestamp of the item with the latest timestamp seen
+thus far, **current watermark** is the most recent fast-forwarded
+watermark plus the amount of system time since the item that caused
+the fast-forwarding, **item timestamp** is the timestamp on the
+incoming item, **item late** is if the item is before the current
+watermark, **watermark candidate** is the item's timestamp minus the
+`wait_for_system_duration` if the item wasn't late, **FFWD accepted**
```
whoahbot, line 105: Maybe `Accepted Watermark` would be a better label for this column?

primalleous (changes requested):
L70: ... is used to extract ~~and assign~~ the timestamp with each value. ...

Comment on code in class on L231 of bytewax/pysrc/bytewax/operators/windowing.py 
I'm not seeing a comparison within the code for value_event_timestamp to anything. I'm also not seeing a value reference for largest timestamp ever seen. It only uses a comparison to current system time. So I'm unsure what to do about that. I suggest doing this:

L71: 2. The current system time is frequently stored internally and used to track the max system time of events seen.

L72: ~~2.~~ 3. The watermark...
L72: (System time never affects the timestamp assigned to an item; *like you'd see with a SystemClock.*)


L97: In the following table,~~**system time**~~ **System Time** represents the real clock time ~~receive an item or are updating the watermark, **system time of max**~~ initialization takes place, an item received or watermark updated. **Sys Time Max** is updated relative to System Time. ~~**current watermark**~~ **Current Watermark** ... ~~**item timestamp**~~ **Item Time** ... ~~**item timestamp**~~ **Item Timestamp** ... ~~**item late**~~ **Item late?** ... ~~**watermark candidate**~~ **Watermark Candidate** ... ~~**FFWD accepted**~~ **Accepted?**is whether the new candidate watermark would be accepted and advanced the watermark relative to ... (note: I'm actually not seeing max_ts - wait_for_system_duration + system_time_of_max explicitly in the code, I'm not sure if this is correct).

The first row represents initialization of the EventClock state where UTC_MIN is datetime.min, (1/1/1 00:00:00). In the second row, an event occurs and we do various comparisons resulting in updating the watermark. The third row represents a system time notification update (no item received) but the Sys Time Max has been updated relative to the previous row since we just had an event and it was accepted. In the fourth row, Sys Time Max is not updated since there was no new item/event for comparison, accepted is N/A. In the fifth row, since the Watermark Candidate is less than the Watermark it is not accepted/not stored and the fifth row showcases a late item since the Item Time is before the Watermark.

Watermark Calculation (calculated when system time notification occurs or item received).
Stored Watermark + (System Time - Sys Time Max)

Stored Watermark = Item Time - Wait Time (calculated and possibly updated when item received).

| System Time | Sys Time Max | Watermark         | Item Time | Item Late? | Watermark Candidate | Accepted? | Stored Watermark |
| ----------- | ------------ | ----------------- | --------- | ---------- | ------------------- | --------- | ---------------  |
| 10:31:00    | 10:31:00     | UTC_MIN           |           |            |                     |           | UTC_MIN          |
| 10:32:00    | 10:31:00     | ~UTC_MIN          | 10:35:00  | No         | 10:30:00            | Yes       | 10:30:00         |
| 10:33:00    | 10:32:00     | 10:31:00          |           |            |                     |           |                  |
| 10:34:00    | 10:32:00     | 10:32:00          | 10:38:00  | No         | 10:33:00            | Yes       | 10:33:00         |
| 10:36:00    | 10:34:00     | 10:35:00          | 10:39:00  | No         | 10:34:00            | No        | 10:33:00         |
| 10:38:00    | 10:34:00     | 10:37:00          | 10:36:00  | Yes        |                     |           |                  |

whoahbot (approved):

PR #533, opened 2025-01-21
Bump PyO3 to 0.23.4

davidselassie:
- **Bump PyO3**
- **Don't use deprecated `*_bound` calls**
- **Fixes some other rename deprecations**
- **Fixes deprecation on implicit trailing `None` default args**
- **Removes a use of `GilOnceCell` that doesn't need GIL**
- **`TdPyAny` is now `Clone` via `Arc`**
- **Migrates to new PyO3 conversion traits**
- **Uses easier conversion method**
- **Migrates some other old gilref uses to `Bound`**
- **`TdPyAny` is no longer `PartialEq`; not used**
- **Make `wrap_key` get GIL in batches**
- **Clippy fixes**

The most notable of the changes here is that `TdPyAny` is now `Arc<PyObject>` and removing `impl Clone` on all our other Python ABC shim types (which were `PyObject`). This is because of https://pyo3.rs/v0.23.4/migration.html#pyclone-is-now-gated-behind-the-py-clone-feature which basically says the only way to soundly increment the Python obj refcount is to be actually holding the GIL. This would introduce a ton of overhead in that you'd acq-rel the GIL on every `clone` and `drop` call. The best (and now default) course of action is to not `impl Clone` so you have to manually use `clone_ref` which takes a GIL token and you can performantly do this. The other option they provided during migration is a `py-clone` feature flag which `impl Clone`, but will panic if you are not holding the GIL; because most folks were using `clone` in a GIL block, this keeps the performance and behavior identical, but will panic instead of a soundness issue. Since we need `Clone` because Timely uses it to duplicate data if a `Stream` has two consumers, we have no control over how or when `clone` is called. Thus we can't guarantee the GIL will be held and we can't use the `py-clone` feature.

The other "hack" they recommend is to instead wrap your `Py` in `Arc`. This means there's two layers of reference counting: if a Python object is cloned on the Rust / Timely side, it increments the ARC refcount; on the Python side or when the GIL is held, it increments the Python object refcount. We provide a "destructor" which will "shim" changes to the ARC refcount into the Python refcount via a new function we wrote `TdPyAny::into_py` which we can call once we actually have the GIL.

This will have some not-obvious performance implications, but I think they'll be minimal and dependent on if there is a lot of branching in the dataflow.

FWIW, the discussion in https://pyo3.rs/v0.23.4/migration.html#pyclone-is-now-gated-behind-the-py-clone-feature implies that memory leaking is still a possibility: Any time `drop` is called on `Py` when the GIL isn't held, the implementation is to buffer the refcount decrement, but that could always hypothetically result in mem leaks. I don't see any way around that with the current architecture, although I have no idea how often this would happen. To me, this implies that `Py` should be a "must move" type which you always "bind" in order to ensure you have the GIL when dropping.

Funnily enough, all this goes away in a fully-nogil future because the Python refcount can be modified by multiple threads.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -90,7 +90,7 @@ checksum = "c980ee35e870bd1a4d2c8294d4c04d0499e67bca1e4b5cefcc693c2fa00caea9"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.52",
+ "syn 2.0.96",
 ]
 
 [[package]]
@@ -440,7 +440,7 @@ checksum = "87750cf4b7a4c0625b1529e4c543c2182106e4dedc60a2a6455e00d212c489ac"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.52",
+ "syn 2.0.96",
 ]
 
 [[package]]
@@ -544,9 +544,9 @@ dependencies = [
 
 [[package]]
 name = "heck"
-version = "0.4.1"
+version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "95505c38b4572b2d910cecb0281560f54b440a19336cbbcb27bf6ce6adc6f5a8"
+checksum = "2304e00983f87ffb38b55b444b5e3b60a884b5d30c0fca7d82fe33449bbe55ea"
 
 [[package]]
 name = "hermit-abi"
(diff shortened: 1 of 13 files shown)


Psykopear:
I read the original thread, and I too came to the conclusion that wrapping the objects with Arc was probably the way to go.
But if I understand this correctly, does that mean that anytime we exchange data between workers, the python object will leak because it's dropped by timely without the GIL, and the refcount on the python side never gets to 0? That could be a serious problem for any long running dataflow

PR #547, opened 2025-04-22
Fix Kafka source hanging when oauth_cb is configured

lpsinger:
There is a known issue in confluent-kafka-python that the admin client does not automatically start calling any configured oauth_cb callbacks. The workaround is to call `poll()` once before calling `list_topics()`.

See https://github.com/nasa-gcn/gcn-kafka-python/blob/v0.3.3/gcn_kafka/core.py#L149-L151.


lpsinger:
Thank you!

